PR #1, opened 2016-02-08
Fixed default mutable arg

chadlung:
See: https://pythonconquerstheuniverse.wordpress.com/2012/02/15/mutable-default-arguments/


Miserlou:
Excellent catch! Thanks!

PR #5, opened 2016-02-09
allow for different regions

emmjaykay:
I noticed when I ran Zappa in us-west-2 instead of us-east-1 I ran into errors about no integrations found.

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -685,10 +685,14 @@ class Zappa(object):
         """
 
         user_home = expanduser("~")
+        credentials = ConfigParser.ConfigParser()
         config = ConfigParser.ConfigParser()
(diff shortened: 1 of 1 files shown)


Miserlou:
Ah, this is a good thought but what will happen if there is no config file? There needs to be a default.

In the advanced settings of django_zappa, you can pass an 's3_region'. But that should override this value.

The PR is nearly there, but it is a good start! Thanks for sending this!

Miserlou:
Actually, I'm going to merge this is and then add a few things around it.

emmjaykay:
Thanks for doing so. I love Zappa! I hope to add more to it in the future!

Miserlou:
Sounds great to me! There is plenty to be done. I've opened some new Issues for work that still needs to be done!

PR #11, opened 2016-02-17
Adding VPC Lambda support

mathom:
AWS recently added [VPC support](https://aws.amazon.com/blogs/aws/new-access-resources-in-a-vpc-from-your-lambda-functions/) to Lambda. This PR adds a keword argument to support this and bumps botocore to the version it was introduced in. Since the VPC support needs ec2:CreateNetworkInterface to function, I had to add that in the IAM profile as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 boto==2.39.0
 boto3==1.2.3
-botocore==1.3.22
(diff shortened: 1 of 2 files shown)


Miserlou:
This is excellent!

PR #15, opened 2016-02-17
Changed URLs to be without trailing slash

Doerge:
I ran into this while creating a minimal flask example.

The WSGI application receives urls with a trailing slash from `zappa/wsgi.py:create_wsgi_request`, regardless of the url visited actually having one or not.

I poked around API Gateway, and I could not find a way to allow both addresses with and without trailing slashes. By experimenting, it appears the `$input.path` is stripped from trailing slashes. I could not find this behavior documented anywhere. [Here is the docs](http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html) for the mapping template in case anyone wants to take a look.

However I think the most sensible thing is to let the default be without the trailing slash.

What do you guys think?

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -94,5 +94,22 @@ class TestZappa(unittest.TestCase):
         }
         request = create_wsgi_request(event)
 
+    def test_wsgi_path_info(self):
+
+        event = {
+            "body": {},
+            "headers": {},
+            "params": {
+                "parameter_1": "asdf1",
+                "parameter_2": "asdf2",
+            },
+            "method": "GET",
+            "query": {}
+        }
+        request = create_wsgi_request(event)
+
+        self.assertEqual("/asdf1/asdf2", request['PATH_INFO'])
+
+
 if __name__ == '__main__':
     unittest.main()
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -95,7 +95,22 @@ class TestZappa(unittest.TestCase):
         request = create_wsgi_request(event)
 
     def test_wsgi_path_info(self):
+        # Test no parameters (site.com/)
+        event = {
+            "body": {},
+            "headers": {},
+            "params": {},
+            "method": "GET",
+            "query": {}
+        }
 
+        request = create_wsgi_request(event, trailing_slash=True)
+        self.assertEqual("/", request['PATH_INFO'])
+
+        request = create_wsgi_request(event, trailing_slash=False)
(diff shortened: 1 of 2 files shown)


Miserlou:
Good, and you included a test! I think the URL received by WSGI should match the one supplied by the browser.

Just as a style note, however, I don't like lines like

"key=lambda (x,_): x)"

Because it is not clear what x and _ are in this case. Even for anonymous functions I prefer descriptive variable names. Those two lines are now more difficult to read. 

For the health of the project, easy-to-read code is a top priority. This means that code is easier to maintain, and it is easier to contribute to.

Also, I think this code won't work in Python 3 because of PEP 3113. Can you try to clean those to lines up a little bit? Although we currently only target 2.7, I'd greatly prefer if we could make all new contributes future-proof so that the eventual migration is easier.

Doerge:
All comments noted. Will attempt to fix.

> I think the URL received by WSGI should match the one supplied by the browser.
This is what I wanted as well, but I could not see a way to distinguish between `.../myresource` and `.../myresource/` in the mapping template. Can you?

Miserlou:
Ohhh, I see. So you're saying that we're forced to chose between 

site.com/resource/ : site.com/resource
site.com/resource : site.com/resource

and

site.com/resource/ : site.com/resource/
site.com/resource : site.com/resource/

?

The issue with defaulting to no-slash is that Django's APPEND_SLASH, which is default: True, uses a 301 redirect, so going to site.com/resource redirects to site.com/resource/, but for us that would be mapped to site.com/resource, and would thus cause and infinite redirect loop.

So my suggestion is to err on the side of appending slashes automatically. I am open to other opinions here, but I would strongly prefer to support the primary use case as the sane default. It would be okay to have configurable alternatives. This is another unfortunate API GW limitation.

Doerge:
Exactly.

Ah, did not know that Django did that. Usually when I'm creating flask applications, I go without the slash. Perhaps we should make it a configurable option in Zappa core (default=trailing slash), which could then be altered in child projects?

Miserlou:
Yes, I think that is probably the best solution. Not elegant, but I think the best we can do for now.

BTW, if you are working on the Flask example, can you consider https://github.com/Miserlou/django-zappa/issues/9 as well? This was an idea that was presented last night that I think makes sense.

Doerge:
I have reverted/updated the code and added an optional parameter to `create_wsgi_request` for now.

Miserlou:
Fantastic!

PR #16, opened 2016-02-18
DISCUSS - use boto session for creating clients

collingreen:
Use boto session (with explicit credentials and region) when creating
lambda and api gateway clients. This can help boto deal with different
regions more effectively.

Deploying failed every time with `botocore.exceptions.NoRegionError: You must specify a region.` without these changes. Perhaps there is a way to set default region for boto but without it this is necessary?

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -376,7 +376,8 @@ class Zappa(object):
 
         """
 
-        client = boto3.client('lambda')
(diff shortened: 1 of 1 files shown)


Miserlou:
Region should be loaded from 

cat ~/.aws/config 
[default]
output = json
region = us-east-1

But I like this change

PR #17, opened 2016-02-19
Add back rollback functionality

### LINKED ISSUE #2: Add 'Rollback' Capability
It'd be nice if we could do something like

    python manage.py rollback dev 

to return the previously deployed function, or

    python manage.py rollback dev

to rollback 5 versions, etc.

graemeglass:
Resolves #2

I would like to be able to get the FunctionName via the apigateway. Then the user will only have to pass in the stage_name. Any ideas?

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -7,6 +7,7 @@ import math
 import os
 import time
 import zipfile
+import requests
 
 from boto.s3.connection import S3Connection
 from boto.s3.key import Key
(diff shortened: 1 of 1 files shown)


Miserlou:
Is it actually necessary to rollback the API gateway? I thought you could just rollback the Lambda function and have that work? Or are you saying that those are separate functionalities?

Miserlou:
But this is looking good!

graemeglass:
I have never used lambda before, so I am not sure. My assumption was (and could very well be wrong), that you need to update the function, and then redeploy the stage with the updated function. I will investigate this further over the weekend.

Miserlou:
I'm pretty sure that the code and the APIGW are totally separate, and that rolling back the function can be done totally separately from the APIGW. You can check this manually in the console.

graemeglass:
@Miserlou Ah, ok. So in the original issue:

> python manage.py rollback dev 

that dev was the function name, not the stage name? 

If that is the case, then `rollback_lambda_function_version` method is all that is needed to resolve this issue.

Miserlou:
Right, so you set your different production environments in your ZAPPA_SETTINGS, (ex - dev / staging / production), and then rollback would just rollback for the env. Does that make sense?

If you just add one more commit to remove the API GW rollback, I'll merge this in? 

Thanks very much for your contribution!!

graemeglass:
Done. Thanks for the project :-)

Miserlou:
Sweeeet!

Do you want to make a Django management command for this feature as well?

graemeglass:
Will do.

PR #21, opened 2016-02-21
Remove boto dependency + introduce moto tests

### LINKED ISSUE #14: Remove old boto dependancy
Zappa currently uses two versions of boto. (This is the result of me reusing code from an old project).

It's just the code related S3, so it should be easy enough to replace with boto3 S3 stuff.

vascop:
This should be able to close #14.

I used the boto3 transfer API to handle the chunking + retries for us instead of doing it explicitly. If we want more control we can also use the TransferConfig object but it didn't feel necessary at this point. http://boto3.readthedocs.org/en/latest/reference/customizations/s3.html#module-boto3.s3.transfer. I kept the tqdm progress bar you were using, but changed it to bytes instead of chunks because boto will decide for us how to chunk up the file.

In the issue thread you also talked about moto, which I never worked with but it seems like it's working fine to check if the bucket and object were indeed created/uploaded.

There's a problem with the upload, where I can't seem to stop it with a keyboard interrupt mid upload, any clues on what I might have done wrong?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: python
 python:
   - "2.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
-sudo: false
 language: python
 python:
   - "2.7"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
+sudo: true
 language: python
 python:
   - "2.7"
 # command to install dependencies
 install:
-  - "pip install setuptools --upgrade; pip install -U pip; pip install -r requirements.txt"
+  - "pip install setuptools --upgrade; python setup.py install"
 # command to run tests
 script: nosetests
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ python:
   - "2.7"
 # command to install dependencies
 install:
-  - "pip install setuptools --upgrade; python setup.py install"
+  - "pip install setuptools --upgrade; pip install -U pip; python setup.py install"
 # command to run tests
 script: nosetests
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ python:
   - "2.7"
 # command to install dependencies
 install:
-  - "pip install setuptools --upgrade; pip install -U pip; python setup.py install"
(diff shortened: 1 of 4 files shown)


vascop:
Well, besides the error which I'm looking into, I just realised that boto is a dependency for moto. So I guess we won't get rid of it after all?

vascop:
The failure on travis (passes on my machine) looks like this https://github.com/spulec/moto/issues/483

Miserlou:
Well, the point is to remove the old boto from the code, not necessarily from the requirements. Having it as a sub dep of moto is okay!

Annoying about the local/Travis error - can you try running with sudo: false in the .travis.yml?

vascop:
From what I gathered "sudo: false" only forces the build to run in their container-based infrastructure (docker), so that was already activated. Thoughts?

vascop:
After me being stupid messing with the reqs file to try and test if a PR would fix the issue, I guess that using if we really want to use moto this is stuck behind this httpretty bug. We could do this the old fashioned way and just mock S3.

mathom:
Here's another testing library we can try out since we're going pure boto3: https://github.com/garnaat/placebo

Miserlou:
Sorry, I meant with sudo: True - I want to see if the problem only occurs when Travis uses Docker.

vascop:
Well, something else is up.

Miserlou:
So, I don't want to keep the lack of passing tests prevent this from being merged in since the work is still good - it's Travis's fault, not yours.

Can you send me a PR with the just commits up to a050985, so that I can merge that and we can close this?

@mathom - placebo looks awesome, and I love @garnaat's work, but it looks like it will be quite hard to set up, since somebody will have to manually capture the placebo output and create the tests around those.

vascop:
Let me have a go at placebo instead of mock though, if it's not too much work at least we keep a green light. I'll probably have a bit of time today

Miserlou:
Even better!

garnaat:
If there is anything that could be added to placebo to make things easier, feel free to ask.

spulec:
We've released a fixed moto that should address the issues that were being seen here. Completely understand if you've moved onto another solution, but wanted to follow up.

mathom:
Thanks for the update @spulec !

PR #25, opened 2016-02-23
Remove boto dependency + introduce placebo tests

vascop:
Up untill commit https://github.com/Miserlou/Zappa/commit/a0509855496c97aa16c3138e20f1ff996a87f038 this is the same as PR #21. After that I used placebo instead of moto to test the calls. Also ended up testing the remove_from_s3 method because why not.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ filechunkio==1.6
 futures==3.0.4
 jmespath==0.9.0
 nose==1.3.7
+placebo==0.7.2
 python-dateutil==2.4.2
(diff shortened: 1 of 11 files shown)


vascop:
Boto3 transfer API decides automatically if it should use multipart or not, and locally it wasn't doing it. In Travis it was. I forced it to use multipart locally while recording using the TransferConfig object and setting a lower threshold for multipart. Now it should test correctly on both cases.

mathom:
Wow, this looks great!

PR #27, opened 2016-02-24
Update the IAM role if it exists

mathom:
:construction: Waiting on https://github.com/Miserlou/Zappa/pull/25 so I can add some unit tests :construction: 

This will let you use an existing role name or update an older policy from Zappa. I could also check for the assume role policy contents but the odds of us updating that are pretty slim.


mathom:
Another question is how do we want to keep the API Gateway resources up to date? It seems like it would be both a pain and slow to verify all the resources and methods. Maybe we should add a django command to recreate these? I've noticed if you re-deploy it makes another API gateway with the same name and just uses that instead.

Miserlou:
Since #25 is merged in now, do you want to add a test for this PR as well? :) 

(We don't have any excuses anymore!)

Miserlou:
Come to think of it, a little bit of documentation on how to write tests with placebo would also be greatly appreciated so I can try to crank coverage up.

mathom:
:+1: I'll try to get something built up today

mathom:
I'll be AFK for a while but I wanted to share how this all works before I leave. Maybe we should set up a GitHub wiki on this project for info like this?

First, you'll want to decorate your test method with `placebo_session`:
```python
    @placebo_session
    def test_create_iam_roles(self, session):
        z = Zappa()
        arn = z.create_iam_roles(session)
        self.assertEqual(arn, "arn:aws:iam::123:role/{}".format(z.role_name))
```

Now, you'll be able to record the AWS interactions with an environment variable:
```bash
$ PLACEBO_MODE=record nosetests tests.tests:TestZappa.test_create_iam_roles
```

You can optionally pass an AWS profile to use:
```bash
$ PLACEBO_PROFILE=zappa PLACEBO_MODE=record nosetests tests.tests:TestZappa.test_create_iam_roles
```

In this example, it has created the following JSON blobs:
```bash
tests/placebo/TestZappa.test_create_iam_roles
tests/placebo/TestZappa.test_create_iam_roles/iam.CreateRole_1.json
tests/placebo/TestZappa.test_create_iam_roles/iam.GetRole_1.json
tests/placebo/TestZappa.test_create_iam_roles/iam.GetRolePolicy_1.json
tests/placebo/TestZappa.test_create_iam_roles/iam.PutRolePolicy_1.json
```

NOTE: Make sure you edit these for any sensitive information you don't want to share (such as account numbers)!

After the JSON has been created, simply drop the environment variables and re-run your test:
```bash
$ nosetests tests.tests:TestZappa.test_create_iam_roles
.
----------------------------------------------------------------------
Ran 1 test in 2.588s

OK
```

PR #29, opened 2016-02-24
add placebo test for create_iam_roles

vascop:
Like the ones for s3, introduces placebo json replies that can be replayed and tests the creation of a role by checking the returned arn


mathom:
I've pulled this into https://github.com/Miserlou/Zappa/pull/27

Thanks!

PR #34, opened 2016-02-29
use the class session in fetch_logs

vascop:


vascop:
Coverage test failed without a decrease in tests? (-0.0%)

PR #35, opened 2016-02-29
Adds an exclude option to the create_lambda_zip method.

Doerge:
It's nice to be able to exclude stuff that you don't want uploaded, but which are still in the project root. This enables the `create_lambda_zip` method to receive an extra parameter which should be a list of patterns to exclude.

The patterns support relative, and absolute, and unix like syntax through the fnmatch module.
E.g. `exclude = ['./foo/*', '*.git*', '*.DS_Store', 'tmp/*']` will exclude all files in the `foo` and 'tmp' directories, and all git-related and OS X related stupid files.

Still needs to be hooked up to the respective cli's.

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -187,23 +187,23 @@ class Zappa(object):
         Returns path to that file.
 
         """
-        print("Packaging project as zip..")
+        print("Packaging project as zip....")
 
         venv = os.environ['VIRTUAL_ENV']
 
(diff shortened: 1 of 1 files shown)


Miserlou:
Can you clean this up a bit? There are a few random print statements, and a hard-coded path that includes your own system's path.

It'd be good if the the '.DS_Store', '.git', etc could call use the same exclude system.

PR #37, opened 2016-03-02
Allow external redirects

mathom:
This supports external redirects in the exception mapping. I also removed boto from the blacklist because we don't use it anymore and it's handy to be able to ship your own boto at a specific version.


Miserlou:
May I ask for a test for that section? I'd like to be sure that that regex works for both scenarios.

PR #51, opened 2016-04-06
update .gitignore and slightly increase test coverage

dtnewman:
I'm just delving into the project code here. I'm hoping to submit more stuff in the future. For now, I have a small PR doing two things:

* Update .gitignore file with new comments and took out a duplicate reference to "env/" that was already there on line 11
* made a small tweak to an existing test in order to get 100% code coverage for the middleware.py file. Basically, on line 148 of middleware.py there is a statement `if key != 'Location': continue` which never evaluates to True, because in the tests, the Location key is always first in the response headers list. I moved the code in the test around without otherwise changing functionality so that this conditional evaluates to True and therefore the next line of code gets hit during the test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,5 +58,9 @@ docs/_build/
 # PyBuilder
 target/
 
+# PyCharm
+#  Project settings from the PyCharm IDE are stored in the .idea folder
+.idea/
+
 env/
 .DS_Store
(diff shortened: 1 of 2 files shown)


Miserlou:
Awesome, thank you very much!

If you'd like to discuss more, feel free to join us on Slack as well!

PR #54, opened 2016-04-08
Incorrect reques.body content

mialinx:
event_info['body'] contains parsed structure for request with 'application/json' content type.
In such case we should serialize it back to JSON, not just convert into a string.


Miserlou:
Ah, good. I think this is correct.

Miserlou:
Pushed into 0.17.2 - good timing!

PR #55, opened 2016-04-09
Fixes content-type issues with non rfc compliant http clients

maljac:
Some http clients send the content-type as "application/json; charset=utf-8". Thus we need to consider these requests when preparing the http body for the wsgi request.

Also fixes some PEP8 issues in wsgi.py


Miserlou:
Great!

PR #56, opened 2016-04-09
.dist-info dir is needed for cryptography package

mialinx:
cryptography uses pkg_resources.iter_entry_points routine
to locate available cryptographic backends.


Miserlou:
Ah, okay.

PR #57, opened 2016-04-12
Update zipping code

dtnewman:
The main goal of this branch was originally to add test coverage to zappa.py. While doing that, I noticed that some of the code for zipping the pre-compiled packages was inefficient since it was traversing a list inside of a for loop. Instead, I put the list into a set upfront and then do the lookups from there.

Additionally, I added a mocked test so that the changed code is covered (the portion worked on here was not hit by the tests prior to this change either since none of the packages in requirements.py are in the set of `lambda_packages`).


Miserlou:
Great!

PR #60, opened 2016-04-15
set CONTENT_LENGTH for PUT requests

abnerjacobsen:
Zappa is not creating environ['CONTENT_LENGTH'] when the HTTP method called is PUT. This path try to solve this issue.


Miserlou:
Great!

Thanks for your contribution! This should land in the next release, which should come tomorrow or Sunday.

PR #66, opened 2016-04-24
Some bugfixes

mialinx:
Bugfixes:
- canonical headers names
- remove temporary directory after upload
- bug with checking GatewayAPI existance by name instead of id

Features:
Full DynamoDb access added to default zappa policy. Does it make sense?


Miserlou:
Excellent!

PR #67, opened 2016-04-24
test_pr

mialinx:
just to test travis


Miserlou:
Are you getting hung up on the mocked tests that use placebo? https://github.com/garnaat/placebo#usage-as-a-decorator

Happy to teach how to use this if you want to stop by on Slack.

PR #77, opened 2016-05-11
add -v flag to cli

collingreen:
Print the version number and exit if -v or --version is given when
running from the command line.


Miserlou:
Noice

PR #93, opened 2016-05-17
Don't ignore some settings

mathieu1:
Fixes the `use_precompiled_packages` and `exclude` settings being ignored.


Miserlou:
Ooof, this really should have been caught by the tests. Thanks.

PR #96, opened 2016-05-18
adding new env key for aws creds

kienpham2000:
@Miserlou 

I'm using Zappa with Travis CI, it doesn't understand `aws configure` very well, so using AWS session key would be best: http://boto3.readthedocs.io/en/latest/reference/core/session.html


Miserlou:
Great!

PR #106, opened 2016-05-24
Fix iam session

### LINKED ISSUE #105: Unable to locate credentials
When I run at latest master:
`zappa update staging`

I get this error:
```
Updating assume role policy on ZappaLambdaExecution IAM Role.
Unable to locate credentials
```

I do have custom `"profile_name": "myprofile"` at the zappa_settings.json file

Looks like the bug is here that it doesn't use the existing session:
https://github.com/Miserlou/Zappa/blob/master/zappa/zappa.py#L851

Do you think changing to this will work?
`client = self.boto_session.client('iam')`

kienpham2000:
Fixes #105


kienpham2000:
@collingreen @Miserlou CR please.

PR #113, opened 2016-05-25
Enable CORS should also enable per parameter resource

kienpham2000:
I ran into this issue where you have to enable cors for all parameter as well. 

```
Ex: 
/users             <---- enable cors here does not enable cors for all child resources
/users/1/ban
```


Miserlou:
Takk!

PR #126, opened 2016-05-31
use dict comprehension for zappa_returndict

Xuefeng-Zhu:


zappa/handler.py
```diff
@@ -129,14 +129,14 @@ def handler(self, event, context):
...
 
                 # This is the object we're going to return.
-                zappa_returndict = dict()
+                # Pack the WSGI response into our special dictionary.
+                zappa_returndict = {header_name: header_value
+                                    for (header_name, header_value)
+                                    in response.headers}
 
                 if response.data:
                     zappa_returndict['Content'] = response.data
```
collingreen: If 'Content' is in response.headers, it now behaves differently than before (it will be set by response.data instead of by the header). This is probably correct, but should be double checked.

zappa/handler.py
```diff
@@ -129,14 +129,14 @@ def handler(self, event, context):
                 response = Response.from_app(app, environ)
 
                 # This is the object we're going to return.
-                zappa_returndict = dict()
+                # Pack the WSGI response into our special dictionary.
+                zappa_returndict = {header_name: header_value
```
collingreen: As long as we are making tiny refactor PRs, might as well just do `dict(response.headers)`

Xuefeng-Zhu:
@collingreen I have addressed your comment

collingreen:
Cool, thanks!

Issue #131, opened 2016-06-02
zappa schedule doesn't add CloudWatch event to the lambda Event source


kienpham2000:
It created the CloudWatch Rules but I have to update the CloudWatch Rules manually so that the event shows up in Event source of the lambda function. 

After I updated the Rules, the Event source is showing and the Scheduled event is triggering the function.

Miserlou:
With latest from pip?

Are you talking about the keep warm or a custom event? A screenshot of the problem would be helpful here.

kienpham2000:
latest from master. Custom event. 

Screenshot. You can see that the Custom event is using version 44. But the Lambda function doesn't have the Event source linked to it. My current workaround is to update the CloudWatch event manually and it will be linked at the Lambda function. 

![001](https://cloud.githubusercontent.com/assets/22681/15833108/0b8ba030-2bda-11e6-88e1-19249c625d75.jpg)

![002](https://cloud.githubusercontent.com/assets/22681/15833112/101f809e-2bda-11e6-9fba-9c4bc67b1c48.jpg)

PR #142, opened 2016-06-09
Fix CloudWatch scheduled event

kienpham2000:
Trying to fix: https://github.com/Miserlou/Zappa/issues/131

- [x] I'm adding a new `example/` dir with simple flask app to test schedule, and giving new users benefit of a working example right away
- [x] fixing some CLI UX stuff
- [x] zappa schedule without creating a new zip if function already existed
- [x] add -y to `zappa undeploy dev -y`
- [x] added new setting: `lambda_handler` to setting file to specify a specific function to trigger. Default is `handler.lambda_handler`
- [x] added `use_apigateway` boolean to setting file to set whether to use api gateway or not. Events doesn't need api gateway. Default is `True`
- [x] refactor all clients into init https://github.com/Miserlou/Zappa/pull/142/commits/095c8c35e9c4493b679a30b87a411db9e770e3a0
- [x] fix Event source to tie to Lambda function


Miserlou:
I LOVE LOVE LOVE the idea of having a samples directory, this is all great

kienpham2000:
@Miserlou did some more work today: https://github.com/Miserlou/Zappa/pull/142/commits/1f4faa07f52c88dfb57a28f4c246786bad799425

I added a few options to setting file such as: 
- [x] add -y to `zappa undeploy dev -y`
- [x] added new setting: `lambda_handler` to setting file to specify a specific function to trigger. Default is handler.lambda_handler
- [x] added `use_apigateway` boolean to setting file to set whether to use api gateway or not. Events doesn't need api gateway. Default is `True`

Wanted to hear your feedback if this is cool

kienpham2000:
@Miserlou ready for CR

Miserlou:
Awesome! Lots to review here! @collingreen - do you want to take a look too? You're good at code review

Miserlou:
Sorry I haven't had time to respond to this properly yet - I am on holiday! I hope to find some time to get to it soon..

zappa/cli.py
```diff
@@ -474,36 +492,35 @@ def create_package(self):
...
+                settings_s = settings_s + "API_STAGE='{0!s}'\n".format((self.api_stage))
+
+                # Lambda requires a specific chmod
+                temp_settings = tempfile.NamedTemporaryFile(delete=False)
+                os.chmod(temp_settings.name, 0644)
+                temp_settings.write(settings_s)
+                temp_settings.close()
+                lambda_zip.write(temp_settings.name, 'zappa_settings.py')
+                os.remove(temp_settings.name)
+                # lambda_zip.close()
```
collingreen: probably should just delete this commented line now that this is wrapped in a with zipfile

zappa/zappa.py
```diff
@@ -222,8 +225,17 @@ class Zappa(object):
...
+        self.aws_region = aws_region
         self.load_credentials(boto_session, profile_name)
+        self.s3_client = self.boto_session.client('s3')
+        self.lambda_client = self.boto_session.client('lambda')
+        self.events_client = self.boto_session.client('events')
+        self.apigateway_client = self.boto_session.client('apigateway')
+        self.logs_client = self.boto_session.client('logs')
+        self.iam_client = self.boto_session.client('iam')
+        self.iam = self.boto_session.resource('iam')
+        self.s3 = self.boto_session.resource('s3')
```
collingreen: How does boto handle holding these resources open and reusing them instead of grabbing a new one as needed?

zappa/zappa.py
```diff
@@ -869,107 +848,99 @@ def schedule_events(self, lambda_arn, lambda_name, events):
...
+                logger.debug('Rule created. ARN {}'.format(rule_response['RuleArn']))
+
+            logger.debug('Adding new permission to invoke Lambda function: {}'.format(lambda_name))
+            permission_response = self.lambda_client.add_permission(
                 FunctionName=lambda_name,
                 StatementId=''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(8)),
                 Action='lambda:InvokeFunction',
                 Principal='events.amazonaws.com',
-                SourceArn=response['RuleArn'],
+                SourceArn=rule_response['RuleArn'],
```
collingreen: if you are checking `if 'RuleArn' in rule_response above, might as well set a default so this never fails either

zappa/zappa.py
```diff
@@ -869,107 +848,99 @@ def schedule_events(self, lambda_arn, lambda_name, events):
...
-            target_response = client.put_targets(
+            target_response = self.events_client.put_targets(
                 Rule=name,
                 Targets=[
                     {
-                        'Id': function,
-                        'Arn': target_arn,
-                        'Input': inp,
-                    },
+                        'Id': 'Id' + ''.join(random.choice(string.digits) for _ in range(12)),
```
collingreen: can we not rely on event['function'] to be unique any longer?

example/app.py
```diff
@@ -0,0 +1,16 @@
+import logging
+from flask import Flask
+
+app = Flask(__name__)
+logging.basicConfig()
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
+
+@app.route('/', methods=['GET', 'POST'])
+def lambda_handler(event=None, context=None):
```
collingreen: See discussion about lambda_handler.

example/zappa_settings.json
```diff
@@ -0,0 +1,49 @@
...
+       "keep_warm": false,
+       "debug": true,
+       "log_level": "DEBUG",
+       "events": [{
+         "function": "dev_event_zappa_test_flask_app",
+         "expression": "rate(1 minute)"
+         }],
+       "aws_region": "us-west-2",
+       "s3_bucket": "zappa-test-bucket",
+       "lambda_handler": "app.lambda_handler",
```
collingreen: See discussion about lambda_handler.

zappa/cli.py
```diff
@@ -64,6 +77,8 @@ class ZappaCLI(object):
     zip_path = None
     vpc_config = None
     memory_size = None
+    use_apigateway = None
+    lambda_handler = None
```
collingreen: See discussion about lambda_handler.

zappa/cli.py
```diff
@@ -175,7 +194,7 @@ def deploy(self):
         self.lambda_arn = self.zappa.create_lambda_function(bucket=self.s3_bucket_name,
                                                        s3_key=self.zip_path,
                                                        function_name=self.lambda_name,
-                                                       handler='handler.lambda_handler',
+                                                       handler=self.lambda_handler,
```
collingreen: See discussion about lambda_handler.

zappa/cli.py
```diff
@@ -425,21 +447,17 @@ def load_settings(self, settings_file="zappa_settings.json", session=None):
             self.api_stage].get('domain', None)
         self.timeout_seconds = self.zappa_settings[
             self.api_stage].get('timeout_seconds', 30)
+        self.use_apigateway = self.zappa_settings[
+            self.api_stage].get('use_apigateway', True)
+        self.lambda_handler = self.zappa_settings[
```
collingreen: See discussion about lambda_handler.

README.md
```diff
@@ -160,7 +160,8 @@ to change Zappa's behavior. Use these at your own risk!
         "http_methods": ["GET", "POST"], // HTTP Methods to route,
         "integration_response_codes": [200, 301, 404, 500], // Integration response status codes to route
         "keep_warm": true, // Create CloudWatch events to keep the server warm.
-        "log_level": "DEBUG", // Set the Zappa log level. Default DEBUG, can be one of CRITICAL, ERROR, WARNING, INFO and DEBUG. 
+        "lambda_handler": "your_custom_handler", // The name of Lambda handler. Default: handler.lambda_handler
```
collingreen: see discussion about lambda_handler

collingreen:
This has a ton of great improvements. I left some questions I was unsure about, but in general this has lots of good code.

I wanted to start a discussion about adding lambda_handler as a configurable option. Unless I'm mistaken on how this works, I don't think this is something we want to configure -- I think this refers to https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L51. Adding lambda_handler doesn't really help anyone out since they almost certainly need the zappa handler code if they are using zappa in the first place and I think it is misleading (it sounds like it's the function I want to put on lambda, but that is app_function, not lambda_handler). Similarly, lambda_handler probably shouldn't be the api function name in an example app for the same reason.

After clearing some things up and some direction from @Miserlou, I say +1

zappa/zappa.py
kienpham2000: @collingreen looking at the botocore (https://github.com/boto/botocore/blob/master/botocore/client.py#L78), looks like it creates a new instance of those class for the `.client` call. So what I'm doing here is saving the instance of that class. I think it's cleaner to initialize all clients, resources that we needed to use for a specific zappa CLI session call.

zappa/zappa.py
kienpham2000: yea I think this is one source of the conflict. When I query the CloudWatch events that are made by AWS UI, it has this Id in this format and unique. So I'm just trying to simulate that and it seems to work.

zappa/cli.py
kienpham2000: cool will remove

zappa/zappa.py
kienpham2000: sounds good, will handle it properly

kienpham2000:
@collingreen thanks for CR! About `lambda_handler` options:

The way I see how `app_function` works, it won't work for CloudWatch event. Zappa wraps it behind its own handler.py file that have api_gateway logic. If we deploy with Zappa now and test Lambda function manually from AWS Lambda UI, we get no response either because of how we implement the handler.py

One of the feature I wanted to introduce to Zappa is to be able to trigger Lambda function from CloudWatch event. So I introduced the `lambda_handler` options is a way to skip the `app_function` wrapper / handler logic. With this option, we can define our own function as a new entry point for Lambda to call. So we either have to update `app_function` on how it should work to support Event, or adding a new entry point function. I decided to add the new entry point since it's easier to implement.

zappa/zappa.py
collingreen: I agree it is cleaner - it just needs to not cause problems if they get re-used later on instead of immediately after being opened. I expect it works; just something to think about.

zappa/zappa.py
collingreen: I get that 12 random digits is pretty safe from conflict, but I'd feel better if it had even less potential problems - is there a limit to the id length or can we do the current unix timestamp + a random number/number+letter string?

collingreen:
@kienpham2000 that makes a ton of sense -- can we name it something that makes this really clear that it is NOT the standard app handler (with all the zappa magic in it), but a replacement handler instead? I like it as a feature, I just don't want to cause confusion.

kienpham2000:
@collingreen @Miserlou 

cool, let me try to throw some name out here, I'm not good at naming things :-p

`custom_event`
`entry_function`
`event_function`
`event_handler`

collingreen:
I think [we are all bad at naming things](http://martinfowler.com/bliki/TwoHardThings.html) :)

I'll also randomly throw out names and we can make @Miserlou decide. Alternatively, we can just add a very clear comment in the readme and call it done.

`custom_event_handler`
`event_handler_override`
`custom_lambda_handler`
`lambda_handler_override`

Miserlou:
Okay, just read through all the code, there are bunch of really great improvements in here!

Going to fiddle with a few things here and there I think, but on the whole all of this is great. Sorry for the long delay, I was on vacation but I'm back now!

Merging now, will write new tests and publish this stuff tomorrow.

PR #146, opened 2016-06-12
Revert "expand cli help test to include full command list"

Miserlou:
Reverts Miserlou/Zappa#145


Miserlou:
Build failed upon merge. Can you try making this into [] list object rather than a comma delineated string?

PR #155, opened 2016-06-23
[DISCUSS] add support for loading environment vars from s3

collingreen:
A few things should be considered/discussed before merging this in:

1. Is this worth baking into zappa or is it better to just let this get handled application by application? A fairly simple snippet or django app / flask extension can do the same thing (in fact, this came from one I have been using) and be more flexible for whatever the user/application needs. On the other hand, including it gives everyone clear instructions on how to answer a fairly common use case that requires at least a bit of lambda/s3 knowledge to tackle correctly, effectively reducing the barrier to entry to using zappa.

1. Should the file format be a json encoded object? This is setting environment vars, which is simple, flat, and typeless - it might make more sense to use a bash- or ini-like format (eg: KEY_NAME=val). On the other hand, we are already zappa_settings.json, so adding another json format file kind of 'fits'.

From the commit message:

On first initialization, attempt to load a json settings file from s3
and use it to populate environment variables. This will be read every
time lambda creates a new instance or restarts but will be cached for
'warm' runs.

Currently assumes the remote env file is a json formatted object
without any nesting -- each top level key-value pair is added to the
environment. Items are added to the environment instead of passed
around as a configuration object for both simplicity within zappa and
maximium compatibility with third party libraries (eg, both flask and
django often take configuration from environment variables).

Access this feature by specifying `remote_env_bucket` and
`remote_env_file` in your zappa_settings.json file.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -451,6 +451,10 @@ class ZappaCLI(object):
             self.api_stage].get('use_apigateway', True)
         self.lambda_handler = self.zappa_settings[
             self.api_stage].get('lambda_handler', 'handler.lambda_handler')
+        self.remote_env_bucket = self.zappa_settings[
+            self.api_stage].get('remote_env_bucket', None)
+        self.remote_env_file = self.zappa_settings[
+            self.api_stage].get('remote_env_file', None)
 
         self.zappa = Zappa(boto_session=session, profile_name=self.profile_name, aws_region=self.aws_region)
 
@@ -508,6 +512,14 @@ class ZappaCLI(object):
                     settings_s = settings_s + "DOMAIN='{0!s}'\n".format((self.domain))
                 else:
                     settings_s = settings_s + "DOMAIN=None\n"
+
+                # Pass through remote config bucket and path
+                settings_s = settings_s + "REMOTE_ENV_BUCKET='{0!s}'\n".format(
+                    self.remote_env_bucket
+                )
+                settings_s = settings_s + "REMOTE_ENV_FILE".format(
+                    self.remote_env_file
+                )
 
                 # We can be environment-aware
                 settings_s = settings_s + "API_STAGE='{0!s}'\n".format((self.api_stage))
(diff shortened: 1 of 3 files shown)


zappa/handler.py
```diff
@@ -47,6 +50,50 @@ def __init__(self, settings_name="zappa_settings"):
         self.settings = importlib.import_module(settings_name)
         self.settings_name = settings_name
 
+        remote_bucket = getattr(self.settings, 'REMOTE_ENV_BUCKET', None)
+        remote_file = getattr(self.settings, 'REMOTE_ENV_FILE', None)
+
+        if remote_bucket and remote_file:
+            self.loadRemoteSettings(remote_bucket, remote_file)
+
+    def loadRemoteSettings(self, remote_bucket, remote_file):
```
collingreen: should be load_remote_settings

zappa/handler.py
```diff
@@ -47,6 +50,50 @@ def __init__(self, settings_name="zappa_settings"):
         self.settings = importlib.import_module(settings_name)
         self.settings_name = settings_name
 
+        remote_bucket = getattr(self.settings, 'REMOTE_ENV_BUCKET', None)
+        remote_file = getattr(self.settings, 'REMOTE_ENV_FILE', None)
+
+        if remote_bucket and remote_file:
+            self.loadRemoteSettings(remote_bucket, remote_file)
```
collingreen: load_remote_settings

zappa/cli.py
```diff
@@ -425,6 +425,10 @@ def load_settings(self, settings_file="zappa_settings.json", session=None):
             self.api_stage].get('domain', None)
         self.timeout_seconds = self.zappa_settings[
             self.api_stage].get('timeout_seconds', 30)
+        self.remote_env_bucket = self.zappa_settings[
+            self.api_stage].get('remote_env_bucket', None)
```
collingreen: .get(x, None) is 'wrong' but is here for consistency

collingreen:
Rebased on master (@miserlou - would you prefer dev?).

Miserlou:
LOVE THIS!

Sorry for not merging it earlier! Very happy with this one. Tests and version bump soon. Merging now.

I love well documented PRs..

PR #165, opened 2016-07-04
Fix the bug for undeploy command on CLI

LyuGGang:


collingreen:
Haven't tested it but it looks reasonable.

Miserlou:
Not sure how I missed this until now! Needs test coverage.

PR #170, opened 2016-07-07
remove input in target creation when scheduling tasks

angstwad:
AFAICT, when injecting JSON on the target instead of the "matched event" prevents the event data this function is looking for above, specifically:
https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L120-L122

Since it never gets  the event with the data it's looking for here, like the ARN and the `detail-type`, we're never entering this if block, and the scheduled code isn't running.

Removing the input fixes this.

I can probably help with tests and such, just wanted to let CI run.

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -910,7 +910,6 @@ class Zappa(object):
                     {
                         'Id': 'Id' + ''.join(random.choice(string.digits) for _ in range(12)),
                         'Arn': lambda_arn,
-                        'Input': json.dumps({'detail': function})
                     }
                 ]
             )


Miserlou:
That's interesting, but I'm confused about what's happening.

What function are you scheduling? My keep-warms all work, but I don't yet have a reliable example for a non-scheduled (ex S3) because I don't have a good syntax example.

angstwad:
Yeah, i think that's the issue.  It looks like keep-warms and scheduled functions are created differently. 
https://github.com/Miserlou/Zappa/blob/master/zappa/zappa.py#L873

I'm scheduling some interval/scheduled tasks to do some administrative reports and stuff.

I struggled for a while with this, and removing the input was the thing I found that injected the proper matched event data for scheduled functions.

Miserlou:
If that line isn't there, how does the schedule know which function to call?

angstwad:
The function is determined from the ARN from the details injected by the
matched event. Additionally, if the matched event is not injected it cannot
be determined if anything should run at all.

On Thursday, July 7, 2016, Rich Jones <notifications@github.com> wrote:

> If that line isn't there, how does the schedule know which function to
> call?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/Miserlou/Zappa/pull/170#issuecomment-231219508>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABLupI3NuSZHE1yAjMVUKdY3WsQFcudlks5qTXYIgaJpZM4JGsz_>
> .
>


-- 
-Paul
Sent from my iPhone

Miserlou:
Ah, okay, it does look like even in the test event, the schedule_me function is defined in the ARN and there is no detail JSON. I wonder how this got in there then. Can you confirm that this works for both the default keep warm and other functions you've scheduled?

PR #171, opened 2016-07-07
Raise exceptions when app_module fails to load

gricey432:
Currently I'm getting an import error deep inside my app but I can't see that because it's caught and then I get an exception about app_module being referenced before assignment.
It'd be better if it showed me the real exception


gricey432:
I've today realised this is wrong and isn't actually doing what I wanted. I've raised #174 to correct it

Issue #195, opened 2016-07-25
pyenv-compatible deployment


puhitaku:
Your Zappa with normal virtualenv works flawlessly and I want it to work also with pyenv.

Pyenv generates a config file named `.python-version` in working directory calling `$ pyenv local venv` and I've implemented auto-detection of it. I'm ready to issue PR if you like it.

Thanks!

Miserlou:
This is an interesting idea, I don't know very much about pyenv though. Zappa currently only supports Python 2.7 though - isn't PyEnv for managing different python versions?

Still, PRs a more than welcome, as long as the test coverage doesn't decrease and there's documentation!

puhitaku:
My words were too few. pyenv can manage also venvs like different python versions while auto-activation functionality is still active.

Anyway, I've created new PR!

Miserlou:
This is merged and will be published in the next version of Zappa! Yay!

PR #197, opened 2016-07-26
Pyenv-compatible deployment

puhitaku:
## Issue
Related issue: #195 

## What I did
 - Added deployment method for venv which is under pyenv's management

`.python-version` is a file which `pyenv` makes when setting venv for the working directory and pyenv writes the name of venv into it. Pyenv makes auto-activating possible with this file. Unlike normal `virtualenv` command, pyenv makes / stores new venv into `~/.pyenv/versions/[name_of_venv]` and Zappa has to bring the venv directory from there when making a ZIP of venv.

The command `pyenv which python` shows the venv directory where the real binary of python sits on and Zappa will get it from the STDOUT.

`subprocess.check_output` is not a best method but I believe this is a simplest way to do this because we don't have to take care of pyenv's installed directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ _Before you begin, make sure you have a valid AWS account and your [AWS credenti
 
     $ pip install zappa
 
-Please note that Zappa *must* be installed into your project's [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/).
+Please note that Zappa *must* be installed into your project's [virtual environment](http://docs.python-guide.org/en/latest/dev/virtualenvs/). If you use pyenv and love to manage virtualenvs with **pyenv-virtualenv**, you just have to call `pyenv local [your_venv_name]` and it's ready.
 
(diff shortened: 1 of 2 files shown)


puhitaku:
Forgot to add some docs. Please wait a sec.

Miserlou:
Hey Takumi!

This all looks great, I look forward to merging this in!

My one concern is the subprocess opening - is there any way that could be used as an attack vector? I don't know very much about pyenv - surely there's a way to do that programmatically?

puhitaku:
I did some research about `which`-ing mechanism of pyenv, and found that re-implementing pyenv's `which` is too complicated.

Yes, as you say, `subprocess` *can* be unsafe method but the command that Zappa executes is not user-defined, and additionally I added a commit 355b0aa that avoids executing `pyenv` command via shell for avoiding vulnerability of shellshock.
I think this should be enough.

Miserlou:
Merged! Thanks very much for you contributions!

Issue #200, opened 2016-07-27
Customizable IAM policies


puhitaku:
Zappa's hard-coded default IAM policies (assume / attach policy) are enough for doing most things but I'd like to customize them for advanced usage.

My idea is that defining the name of external JSON files which includes IAM's policy on `zappa_settings.json` then Zappa reads the external JSONs on deployment.

I've made new PR for this idea. Please consider merging this!

Miserlou:
Merged! Thank you very much for your contribution!

PR #201, opened 2016-07-27
Customizable IAM policies

puhitaku:
## Issue
 #200 

## What I did
Added custom value for specifying policy JSON files.

Without specifying policy files Zappa will use its default hard-coded policy.


puhitaku:
Added an event in `test_settings.json` for increasing coverage.

PR #209, opened 2016-07-31
Cloudformation

mathom:
:fire: WIP :fire: This is going to be broken for a while.

This is a rewrite of the boto3 management backend with [troposphere](https://github.com/cloudtools/troposphere) and [CloudFormation](https://aws.amazon.com/cloudformation/).

Right now it deploys and seems to work for simple test projects. :+1: 

All kinds of stuff is broken, including:
 * lambda versioning
 * a lot of the tests
 * ~~there's a hard limit from AWS on template size (with current Zappa this is 6 deep, max)~~

I'll probably look into some chunking ideas and see if the size limitation is simple to work around.

Here's the resulting monstrosity (6 deep):
fireshot capture 6 - aws cloudformati_ - https___console aws amazon com_cloudformation_designer_home

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,5 +13,6 @@ requests>=2.10.0
 six==1.10.0
 tqdm==4.8.3
 Werkzeug==0.11.10
+troposphere==1.7.0
 wheel==0.29.0
 wsgi-request-logger==0.4.5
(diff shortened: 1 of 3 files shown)


mathom:
Closing for now - goes out of date quick!

PR #212, opened 2016-08-01
Change default log level to INFO, getting way to much crap from boto …

mjgorman:
…getting settings files, can be changed to debug with settings, loaded later in the hanlder.py Hiding the setting stuff unless you're log level is set to debug as well. Want to be able to give devs access to logs without them accessing env info that could have keys and such. 

Sorry about formatting changes. VIM + pylint stuff does that sometimes.


Miserlou:
Great to me!

PR #219, opened 2016-08-04
Expose context lambda payload, context.

aectann:
Useful e.g. when you need to get access to current lambda function name.

```current_lambda_name = request.environ['lambda.context'].function_name```

And having payload passed to 'command' invocations allows for parametrised calls.

### DIFF
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -1,0 +1,41 @@
+import unittest
+from zappa.handler import LambdaHandler
+
+
+def no_args():
+    return
+
+
+def one_arg(first):
+    return first
+
+
+def two_args(first, second):
+    return first, second
+
+
+def var_args(*args):
+    return args
+
+
+def var_args_with_one(first, *args):
+    return first, args[0]
+
+def unsupported(first, second, third):
+    return first, second, third
+
+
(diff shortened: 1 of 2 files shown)


Miserlou:
I like the context in the environ, is there a way we could check the signature of the function? Apparently inspect.getargspec would solve this - then we could could conditionally support simple functions and functions that take (e, c). This pattern would be good for direct invocations, timed functions and AWS reactive events.

aectann:
I thought of using introspection to do that, but then decided to keep it simple. After all requiring a function to accept `*args` shouldn't be that big of a deal. 

If you think this flexibility is required I can look into doing the introspection check.

Miserlou:
Yeah, I think I'd like to have the flexibility so that we can use the same pattern for all event types. For AWS events, it makes sense to require (e, c), but I don't think it makes sense to require that for recurring events, invocations, etc, and I'd like to be able to reuse the same pattern.

It's short, just something like

`
import inspect
if len(inspect.getargspec(func).args) == 2: 
    func(event, context)
else:
   func()
`

aectann:
@Miserlou updated

Miserlou:
Excellent! Thank you very much for your contribution!

PR #222, opened 2016-08-04
Fix attach policy to access SQS

parroyo:
Fix attach policy to use SQS resources from Zappa


Miserlou:
Ahhhh, I _just_ pushed a fix to this! Thanks for reporting and the PR though!

parroyo:
Ops, I didn't see it. sorry ;)

PR #226, opened 2016-08-05
More scheduled event docs

mathom:
This fleshes out the example a bit more and provides some details for running things locally.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -6,7 +6,7 @@ This will deploy a simple Flask app
     # configure your AWS keys
     $ aws configure
 
-## Deploy
+## Deploy WSGI App
     # deploy API to dev environment
     $ zappa deploy dev_api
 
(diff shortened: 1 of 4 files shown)


example/README.md
```diff
@@ -22,3 +25,25 @@ This will deploy a simple Flask app
...
+
+You can also call a scheduled function packaged along with your normal WSGI app:
+
+    # deploy API to prod environment
+    $ zappa deploy prod
+
+    # schedule CloudWatch to run mymodule.myfunc every 5 minutes
+    $ zappa schedule prod
+
+This function can optionally take the usual `event` and `context` lambda arguments.
```
collingreen: an example here will probably save people time

collingreen:
+1

example/README.md
mathom: good idea, added an example

PR #231, opened 2016-08-10
Content type aliases

aectann:
Allow setting up aliases for integration template content types. Useful for e.g. picking up  `application/whatever+json`  as `application/json`.


Miserlou:
This looks like a good change, but needs an updated entry in the advanced settings of the README.md before I can merge it in! Even better would be a little example about why somebody might want to use this feature, etc.

PR #235, opened 2016-08-10
Allow overriding of method_header_types

### LINKED ISSUE #234: Cannot set custom headers
Zappa does not let me specify custom headers which should be returned in the response.

jakul:
Fixes #234 

@Miserlou


Miserlou:
Good one!

PR #248, opened 2016-08-15
Cloud watch event management improvements

aectann:
- Use lambda name as prefix for all CW events. 
- Make scheduling/un-scheduling part of deploy/undeploy. 
- Schedule warm-up event together with the rest of CW events.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -259,10 +259,8 @@ class ZappaCLI(object):
                                                        timeout=self.timeout_seconds,
                                                        memory_size=self.memory_size)
 
-        # Create a Keep Warm for this deployment
-        if self.stage_config.get('keep_warm', True):
-            keep_warm_rate = self.stage_config.get('keep_warm_expression', "rate(5 minutes)")
-            self.zappa.create_keep_warm(self.lambda_arn, self.lambda_name, schedule_expression=keep_warm_rate)
+        # Schedule events for this deployment
+        self.schedule()
         endpoint_url = ''
         if self.use_apigateway:
             # Create and configure the API Gateway
@@ -415,17 +413,12 @@ class ZappaCLI(object):
             self.zappa.remove_api_gateway_logs(self.lambda_name)
 
         gateway_id = self.zappa.undeploy_api_gateway(self.lambda_name, self.api_key_required)
-
-        if self.stage_config.get('keep_warm', True):
-            self.zappa.remove_keep_warm(self.lambda_name)
-
+        self.unschedule()  # removes event triggers, including warm up event.
         self.zappa.delete_lambda_function(self.lambda_name)
         if remove_logs:
             self.zappa.remove_lambda_function_logs(self.lambda_name)
 
         print("Done!")
-
-        return
 
     def schedule(self):
         """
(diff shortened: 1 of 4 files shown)


Miserlou:
Build failed :(

aectann:
@Miserlou I'm on it. Will update after lunch )

Miserlou:
Warm up recreation needs to be in update, there's currently no way to set the keep-warm to $LATEST, so each successive revision needs its own keep-warm. It should be investigated whether or not all re-scheduling should happen automatically (with setting) into the update command until this is availabl.e

aectann:
@Miserlou so are you saying that other CW events are being lost after an `update`?

Miserlou:
Not that they're "lost", just that they will execute the version of the code that was $LATEST when the function was scheduled. Another alternative is to add aliasing to our lambda functions, in fact this probably makes the most sense, since we can't use $LATEST, we just maintain our own pseudo-$LATEST.

Miserlou:
Calling unschedule/schedule as part of 'update' would be an interim soln.

Miserlou:
Although that may depend on the type of event. Some event sources may be able to use $LATEST, that's the tricky bit.

aectann:
@Miserlou included `schedule()` to be invoked as part of `update`; until we figure out better way to do it.

Miserlou:
Does that cause multiple event executions? Does unschedule need to happen before the second schedule?

aectann:
@Miserlou unschedule is called as part of schedule, so calling schedule multiple times won't result in multiple CW events being created.

Miserlou:
Ah, gotcha

Miserlou:
Are you sure about that? Deleting a rule isn't the same as unscheduling an event, it's a just one subtype.

aectann:
@Miserlou I'm not sure if I understand the question. Schedule does remove rules and their target before anything else is added. Did I miss something?

Miserlou:
Yeah, there are non-CWE event sources (S3/SQS/Kinesis, etc.). It's a different code path. I don't know what happens if a function gets updated, if their event triggers keep pointing at $LATEST or if they also get stuck on an old version.

aectann:
@Miserlou well, those are still being removed by unschedule (I didn't change that), as long as their names are in the events array.

zappa/cli.py
```diff
@@ -422,7 +422,7 @@ def schedule(self):
                 events = []
             keep_warm_rate = self.zappa_settings[self.api_stage].get('keep_warm_expression', "rate(5 minutes)")
             events.append({'name': 'zappa-keep-warm',
-                           'function': self.lambda_handler,
+                           'function': 'handler.heater',
```
Miserlou: Why are you doing this?

The goal of this is to call the web function handler so that the function is kept warm.

aectann: @Miserlou Since it's just a CW event, `lambda_handler` will be called no matter what. Then it will call `handler.heater` via Scheduled event branch. If I'm reading it correctly, it never was able to get to the web request handling code (i.e. to warm up application module import). Having said that, we now can import any modules in the `heater` method and probably even to call back into a user specified warmup function. 

I might be reading it wrong, of course.

Miserlou: Ahhh, hm, okay, I see your point. I don't think it needs its own empty function though. The point is that it should actually call the default WSGI handler so that any imports etc that need to happen inside the application are warm as well. I don't think imports are kept in cache unless they are explicitly made.

"`heater`" is cute but I think ultimately confusing: keep_warm_callback maybe? Is there actually a use case for this though, versus just scheduling a custom function? I don't want to add advanced_settings unless they're actually useful.

aectann:
@Miserlou I've updated the method name and made it to call web app branch of the handler code.

Miserlou:
Sorry I didn't get to this today, was too deep with the new R53 stuff. I'll hopefully get to this tomorrow, but I need to test a few things re: non-CWE events before I can merge.

aectann:
@Miserlou no worries. I've been using this branch for two days now. So I might confidently say that at least CW events work just fine :)

aectann:
Rebased over master.

zappa/cli.py
```diff
@@ -462,26 +461,32 @@ def unschedule(self):
 
         """
 
-        if self.stage_config.get('events', None):
-            events = self.stage_config['events']
+        # Run even events are not defined to remove previously existing ones (thus default to []).
```
collingreen: missing an 'if'?

zappa/cli.py
```diff
@@ -462,26 +461,32 @@ def unschedule(self):
...
+            return
 
-            try:
-                function_response = self.zappa.lambda_client.get_function(FunctionName=self.lambda_name)
-            except botocore.exceptions.ClientError as e: # pragma: no cover
-                print("Function does not exist, please deploy first. Ex: zappa deploy {}".format(self.api_stage))
-                return
+        if not events:
+            print("No events defined in zappa config. Proceeding to unschedule to remove leftovers (in case "
+                  "configuration was changed).")
```
collingreen: Is there a case where the user needs to see this in the logs / can we just silently continue here?

zappa/cli.py
aectann: Yup, fixed.

zappa/cli.py
aectann: Indeed, that was there for myself while debugging. Removed.

Miserlou:
Dang, I think this needs another rebase because I merged the other PR first. 

I'm also waiting for a version bump from Kappa that relates to this PR before I merge. Once it's merged, we can do a version release.

aectann:
Do you mind if I just merge master in? I tried rebasing and its conflicts
all-over.

Miserlou:
Please do. `-dev` was wishful thinking, it seems.

aectann:
@Miserlou Done.

Miserlou:
Merged!

It's really noisy in the console but I think it makes sense on the whole. Making the CLI prettier is another issue.

Issue #251, opened 2016-08-16
Nosetests are failing on Windows


mynameismevin:
Ran the nosetests on Windows, they are currently failing in a virtualenv:

```python
.....E...E
  0%|          | 0/3 [00:00<?, ? endpoint/s]
4 endpoint [00:00, 1000.01 endpoint/s]      
....
  0%|          | 0/3 [00:00<?, ? endpoint/s]
4 endpoint [00:00, 800.02 endpoint/s]       
................E............
======================================================================
ERROR: test_cli_aws (tests.tests.TestZappa)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "F:\Programming\Repositories\Zappa\tests\utils.py", line 49, in wrapper
    return function(*args, **kwargs)
  File "F:\Programming\Repositories\Zappa\tests\tests.py", line 460, in test_cli_aws
    zappa_cli.update()
  File "F:\Programming\Repositories\Zappa\zappa\cli.py", line 319, in update
    self.create_package()
  File "F:\Programming\Repositories\Zappa\zappa\cli.py", line 933, in create_package
    django_py = ''.join(os.path.join([base, os.sep, 'ext', os.sep, 'django.py']))
  File "C:\Users\Mike\zappa-dev\lib\ntpath.py", line 65, in join
    result_drive, result_path = splitdrive(path)
  File "C:\Users\Mike\zappa-dev\lib\ntpath.py", line 116, in splitdrive
    normp = p.replace(altsep, sep)
AttributeError: 'list' object has no attribute 'replace'
-------------------- >> begin captured stdout << ---------------------
This is a prebuild script!
This application is already deployed - did you mean to call update?

This is a prebuild script!
Packaging project as zip...

--------------------- >> end captured stdout << ----------------------
-------------------- >> begin captured logging << --------------------
placebo.pill: DEBUG: attaching to session: Session(region_name='us-east-1')
placebo.pill: DEBUG: datapath: F:\Programming\Repositories\Zappa\tests\placebo\TestZappa.test_cli_aws
botocore.credentials: DEBUG: Skipping environment variable credential check because profile name was explicitly set.
botocore.credentials: DEBUG: Looking for credentials via: env
botocore.credentials: DEBUG: Looking for credentials via: assume-role
botocore.credentials: DEBUG: Looking for credentials via: shared-credentials-file
botocore.credentials: INFO: Found credentials in shared credentials file: ~/.aws/credentials
[...]

PR #252, opened 2016-08-16
Implemented Windows-specific fix

mynameismevin:
This fixes a Windows pathing issue regarding Miserlou/Zappa#251.

All of the tests pass, with the exception of one: `ERROR: test_upload_remove_s3 (tests.tests.TestZappa)`

This is due to random filesystem permissions on my Windows 10 box, so if someone else has a Windows box and can run the tests, it would be nice to double tap.

Here is the test case I was using to find the fix:

```python
import os
base = __file__.rsplit(os.sep, 1)[0]

try:
    print("test: using a redundant join method.")
    test = ''.join(os.path.join([base, os.sep, 'foo', os.sep, 'bar.py']))
    print("success: {0}".format(test))
except Exception as e:
    print("failure: known error: {0}!".format(e))

try:
    print("test: using regular join method.")
    test = "".join([base, os.sep, 'foo', os.sep, 'bar.py'])
    print("success: {0}".format(test))
except Exception as e:
    print("failure: join() error: {0}!".format(e))

try:
    print("test: using concatenation.")
    test = os.path.join([base, os.sep, 'foo', os.sep, 'bar.py'])
    print("success: {0}".format(test))
except Exception as e:
    print("failure: os.path.join() error: {0}".format(e))

try:
    print("test: using os.path.join() without casting list.")
    test = os.path.join(base, [os.sep, 'foo', os.sep, 'bar.py'])
    print("success: {0}".format(test))
except Exception as e:
    print("failure: os.path.join() error: {0}".format(e))

print("done.")
```

If you want to hear about the in-depth reason, I'm more than willing to go into as much detail as you'd like.

:dromedary_camel:

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -930,7 +930,9 @@ class ZappaCLI(object):
                 # Copy our Django app into root of our package.
                 # It doesn't work otherwise.
                 base = __file__.rsplit(os.sep, 1)[0]
-                django_py = ''.join(os.path.join([base, os.sep, 'ext', os.sep, 'django.py']))
+
+                # Zappa Issue 251: this is a regular join because of how it interacts with Windows.
+                django_py = ''.join([base, os.sep, 'ext', os.sep, 'django.py'])
                 lambda_zip.write(django_py, 'django_zappa_app.py')
 
                 # Lambda requires a specific chmod


Miserlou:
Excellent, thank you! Certainly a bit of a strange error, thanks for fixing!

mynameismevin:
Welcome! :)

sitrucp:
would this fix have made it into "pip install" ?

Miserlou:
I think that line just got fixed in master recently, maybe try that? New version will be published soon

sitrucp:
Ok great, master worked fine with  `pip install --upgrade git+git://github.com/Miserlou/Zappa@master`

PR #255, opened 2016-08-17
Ability to remove logs during an execution of undeploy.

jotes:
Hi @Miserlou, could you look at this? My company would love if that feature would land in the upstream :D


Miserlou:
Excellent!

PR #265, opened 2016-08-19
Update README.md

pywebdesign:
Add Bundling Additionnal Packages like OpenCV section


Miserlou:
Actually, many of the most popular packages, including OpenCV, _do_ work out of the box, because of our sister project: https://github.com/Miserlou/lambda-packages

You can mention this instead if you want, but your example is wrong..

PR #268, opened 2016-08-22
Cookie expiry parsing

MajorTal:
Cookie year format is 4 digits, hence a lowercase "y"

### DIFF
--- a/zappa/middleware.py
+++ b/zappa/middleware.py
@@ -175,7 +175,7 @@ class ZappaWSGIMiddleware(object):
             for kvp in kvps:
                 kvp = kvp.strip()
                 if 'expires' in kvp.lower():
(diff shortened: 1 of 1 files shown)


Miserlou:
%Y is four digits:
```
>>> import datetime
>>> now = datetime.datetime.now()
>>> now.strftime('%Y')
'2016'
>>> now.strftime('%y')
'16'
>>> now.strftime('%a, %d-%b-%y %H:%M:%S GMT')
'Mon, 22-Aug-16 11:55:29 GMT'
>>> now.strftime('%a, %d-%b-%Y %H:%M:%S GMT')
'Mon, 22-Aug-2016 11:55:29 GMT'
```

Error is happening somewhere else

Miserlou:
Okay, the issue is that we actually need to support %y and %Y. Damn RFC.

PR #270, opened 2016-08-22
Support both 2 and 4 digit year formats in cookies

MajorTal:
Cookie year format is sometimes 2 and sometimes 4 digits, hence a lowercase "y" vs "Y"


Miserlou:
boom baby

PR #280, opened 2016-08-24
Uses more click goodness

xuru:
Added the following click addons:
* click-didyoumean
* click-extensions
* click-plugins
Added example plugin with documentation.
Added prefix to all non-cli commands "cli_"
Now uses the decorator "catch_exceptions" to catch exceptions and remove the S3 zip file if neccessary.

Moved adding conditional command "manage" to __init__.py in order to avoid a circular dep

Added management command (also supports all commandline options of django)

Small fix to invoke when using extra_args

A bit of cleanup, and a look at how difficult tests will be.  Thinking we are going to need to postpone writing tests until we can decouple CLI from the loader code.  This will make testing much easier.

Added click to the requirements explicitly

Changed position of click in the requirements lol

Trying importing to the requirements.txt in order to get travis to pass

Had trouble installing the click-extensions library, so pulled out what we need instead.

Removing more cli tests that passed on my machine (with creds) but not on travis.  Will add them back when we can seperate CLI and logic.

New way of doing settings
Broke down a lot of the large methods to make it more testible.
Lot of unit tests.

Cleaned up some code quality stuff


xuru:
Still need to work in all the changes that happened on master.

zappa/commands/scheduler.py
```diff
@@ -0,0 +1,108 @@
...
+
+    """
+    _unschedule(ctx.obj, env)
+
+
+def _unschedule(config, env):
+    loader = config.loader(env)
+    settings = loader.settings
+
+    function_response = validate_events(loader, settings)
```
aectann: That will fail if event is not defined, while unschedule might still be required (e.g. events was defined, but was intentionally removed to get rid of all CW jobs or when warmup was disabled).

Miserlou:
Unfortunately, there are a lot of things that are preventing me from merging this. :( Primarily - coverage has dropped 18%. Secondly, I very much don't like the pattern of "ripping out what we need" from upstream libraries, as this will eventually become a source of conflict. Far better to file any tickets upstream and have the integration be fixed.

xuru:
What part of it is "upstream libraries"?  As far as I know I only broke down `zappa/ZappaCLI.py` into separate commands in the commands directory and a `zappa/ZappaLauncher.py`.  I was careful not to touch `zappa/zappa.py` as much as possible.

I know this is a huge PR, and I don't blame you for not accepting it.  I will leave the branch available on my repo in case you want to look at what I did later on when things slow down.

BTW, It's a shame that placebo adds so many files.  Really hard to see what is going on.  Thanks for having a look at it!

Issue #284, opened 2016-08-25
Use an existing API Key


parroyo:
It would be nice if Zappa API Gateway could reuse an existing API Key instead of creating a new one on every deployment.

I have a commit including the option "api_key" in zappa_settings.json to cover that. If you agree, I could create a pull request to add this feature.

Miserlou:
Sounds good to me

From my phone..

parroyo:
Closing this one. Already merged.

PR #285, opened 2016-08-25
Add "api_key" option to use an existing API key instead of creating a new one

parroyo:
Related to #284

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -862,58 +862,60 @@ class Zappa(object):
 
         return "https://{}.execute-api.{}.amazonaws.com/{}".format(api_id, self.boto_session.region_name, stage_name)
 
+    def get_api_keys(self, api_id, stage_name):
+        """
+        Generator that allows to iterate per API keys associated to an api_id and a stage_name.
+        """
+        response = self.apigateway_client.get_api_keys(limit=500)
+        stage_key = '{}/{}'.format(api_id, stage_name)
+        for api_key in response.get('items'):
+            if stage_key in api_key.get('stageKeys'):
+                yield api_key.get('id')
+
     def create_api_key(self, api_id, stage_name):
         """
-        Create new API key
+        Create new API key and link it with an api_id and a stage_name
         """
-        try:
-            print("Creating API Key..")
-            response = self.apigateway_client.create_api_key(
-                name='{}_{}'.format(stage_name, api_id),
-                description='Api Key for {}'.format(api_id),
-                enabled=True,
-                stageKeys=[
-                        {
-                            'restApiId': '{}'.format(api_id),
-                            'stageName': '{}'.format(stage_name)
-                        },
-                    ]
-                )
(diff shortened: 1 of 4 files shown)


zappa/zappa.py
```diff
@@ -923,13 +961,13 @@ def undeploy_api_gateway(self, project_name, api_key_required=False, domain_name
...
-            if api_key_required:
-                print("Removing API Key..")
-                api_key_id = [key for key in self.apigateway_client.get_api_keys()['items']
-                              if api['id'] in key['name']][0]['id']
-                api_key_response = self.apigateway_client.delete_api_key(
-                        apiKey="{}".format(api_key_id)
-                )
+            if remove_api_key:
+                try:
+                    api_key_id = [key for key in self.apigateway_client.get_api_keys()['items']
```
Miserlou: This line is really, really ugly - can we refactor this into something more readable?

zappa/zappa.py
```diff
@@ -877,6 +862,59 @@ def deploy_api_gateway(self, api_id, stage_name, stage_description="", descripti
...
+                description='Api Key for {}'.format(api_id),
+                enabled=True,
+                stageKeys=[
+                        {
+                            'restApiId': '{}'.format(api_id),
+                            'stageName': '{}'.format(stage_name)
+                        },
+                    ]
+                )
+            print('x-api-key: {}'.format(response['id']))
```
Miserlou: No need to print this

zappa/zappa.py
```diff
@@ -877,6 +862,59 @@ def deploy_api_gateway(self, api_id, stage_name, stage_description="", descripti
...
+        except botocore.client.ClientError as e:
```
Miserlou: I'd say that unless this function is returning an expected value that will be checked, that we actually _don't_ try/catch this error. Same goes for remove.

zappa/zappa.py
parroyo: Well, I'm not sure of this. 
Like the API gateway URL, the API Key id is created on this stage, and I think is useful to print it. Otherwise, the user have to find it in the AWS console in order to be able to do the server requests.

zappa/zappa.py
parroyo: Makes sense. I'll remove them.

zappa/zappa.py
parroyo: Yes, I've just move the line, but I'll change it to something more readable :)

zappa/zappa.py
Miserlou: Ah, right, okay. Then can we gussy it up a little bit? "Created a new x-api-key: "

Miserlou: Also, I suggest that this be added to the `zappa status` command so that the x-api-key can be retrieved without using the console, as you say.

parroyo: Perfect. I'll take a look at it.

Miserlou:
Excellent!

PR #288, opened 2016-08-26
Fixes #286

### LINKED ISSUE #286: Targets of event rules are duplicated if delete permission is missing
I am experimenting with minimal permissions for the IAM user used by zappa. If the permission for deleting a target/rule is missing, the targets in a schedule are duplicated without that an error is raised. This means, the lambda function gets triggered multiple times simultaneously. 

The issue can be reproduced by restricting the event permissions of the Zappa AWS user to the policy below and run `zappa update`.

```
{
  "Action": [
    "events:PutRule",
    "events:PutTargets",
    "events:ListRules"
  ],
  "Resource": "*",
  "Effect": "Allow"
}
```

claudeatsafe:
I would have preferred to write a test for it but I was a bit overwhelmed by the effort it would have taken me (tested it in my own app though).

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -1275,8 +1275,12 @@ class Zappa(object):
         try:
             targets = self.events_client.list_targets_by_rule(Rule=rule_name)
         except botocore.exceptions.ClientError as e:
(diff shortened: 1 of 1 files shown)


Miserlou:
Looks good!

Issue #225, opened 2016-08-05
Cannot create triggers for SNS events


parroyo:
There is an error running zappa schedule when SNS event sources are configured:

```
$ zappa schedule <stage>
Scheduling..
ERROR:Unable to add SNS event source
Traceback (most recent call last):
  File "/av_home/venv/local/lib/python2.7/site-packages/kappa/event_source/sns.py", line 50, in add
    Endpoint=function.arn)
  File "/av_home/venv/local/lib/python2.7/site-packages/kappa/awsclient.py", line 86, in call
    data = op(**kwargs)
  File "/av_home/venv/local/lib/python2.7/site-packages/botocore/client.py", line 278, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/av_home/venv/local/lib/python2.7/site-packages/botocore/client.py", line 572, in _make_api_call
    raise ClientError(parsed_response, operation_name)
ClientError: An error occurred (InvalidParameter) when calling the Subscribe operation: Invalid parameter: Lambda endpoint ARN
Created sns:Publish event schedule for <module>.<function>!
```

In zappa_settings.json:
```
"events": [
            {
                "function": "<module>.<function>",
                "event_source": {
                    "arn":  "<sns arn>",
                    "events": [
                        "sns:Publish"
                    ]
                }
            }
        ]
```

Looking at the code it seems that the problem is how the arn is split and used in Kappa event source:

In S3EventSource.add:
`'LambdaFunctionArn': '%s:%s' % (function.arn, function._context.environment),`

But in SNSEventSource.add:
`Endpoint=function.arn)`

As the value of 'arn' is split in order to work with the S3EventSource it doesn't work fine with SNS cause it expects the full arn value.

```
    split_arn = lambda_arn.split(':')
    arn_front = ':'.join(split_arn[:-1])
    arn_back = split_arn[-1]
    ctx.environment = arn_back

    ....
    funk.arn = arn_front
    funk._context = ctx
```

I don't know if it should be fixed in Kappa or in Zappa. For me is strange the way Kappa is building the 'LamdbaFuncionArn' in S3EventSource using the context.environment...

Miserlou:
Damn, I was afraid of this.

Yeah, this is directly related to: https://github.com/garnaat/kappa/issues/81

I'd rather have this fixed in Kappa, what we're doing is pretty dirty already, and I'd rather not have to check the the event type. Thanks for digging into this one and surfacing this. 

[...]

PR #290, opened 2016-08-29
Add handler support for SNS, dynamodb and kinesis events

parroyo:
Related to #225

I added a fix to process SNS, kinesis and DynamoDB AWS events. Sadly, I couldn't find a generic way to handle all the events like the s3 events.

The function to call in the handler is not send in other events but s3 ones. So I've added a AWS_EVENT_MAPPING property to zappa_settings.py in order to resolve the function to invoke based on the event source arn.

### DIFF
--- a/test_settings.json
+++ b/test_settings.json
@@ -11,10 +11,21 @@
        "debug": true,
        "parameter_depth": 2, 
        "prebuild_script": "test_settings.prebuild_me",
-       "events": [{
-          "function": "tests.test_app.schedule_me",
-          "expression": "rate(1 minute)"
-       }]
+       "events": [
+           {
+               "function": "tests.test_app.schedule_me",
+               "expression": "rate(1 minute)"
+           },
+           {
+               "function": "test.test_app.method",
+               "event_source": {
+                   "arn": "arn:aws:sns:::1",
+                   "events": [
+                     "sns:Publish"
+                   ]
+               }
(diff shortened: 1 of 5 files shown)


Miserlou:
@parroyo - sorry for not commenting on this earlier!

Mapping events to functions is what I was afraid of, but you've solved it with with `AWS_EVENT_MAPPING`, which looks to be an inspired solution. Thank you for including the tests as well!

Miserlou:
Ah. I think there could be a problem with this, however.

This seems like it only coverts event _sources_, and not event _types_. For example, if I want to have a different function execute in response to SQS pushes and SQS pops on the same queue, this setting won't work. AWS_EVENT_MAPPING must need to be a dict of dicts?

parroyo:
@Miserlou As far as I know, the only lambda subscription that supports different event types is the S3 one. And for S3 events, we are getting the function name to invoke based on the **configurationId** property, so you can have different functions to handle different events from the same S3 source 
```
        if 's3' in record:
             return record['s3']['configurationId']
```
About the SQS events, they cannot invoke lambda functions. You can only subscribe to a SNS topic instead.

Miserlou:
Ah, I see. Okay, gotcha. I'm going to think about this some more, but I think that we're all set then.

I really appreciate you taking the initiative on this issue, truly excellent work!

PR #291, opened 2016-08-30
Pick best match for a domain name when filtering on hosted zones.

aectann:
Fixes an issue when zappa picks an incorrect zone if two zones have same substring. E.g. if both `example.com.au.` and `wider.example.com.au` are present, zappa could pick the former one given `subdomain.wider.example.com.au` as domain name.

The change makes zappa pick best match (longest substring).


Miserlou:
Nice one!

PR #292, opened 2016-08-30
fix for flask deployment

sdementen:
make the django packaging conditional on being a django app


sdementen:
Yes indeed but I would not say it is a chance...

Issue #294, opened 2016-08-31
NameError: global name 's3_client' is not defined (when doing 'zappa certify' using account.key from S3)


smoll:
Following the second example from https://github.com/Miserlou/Zappa/blob/master/docs/domain_with_free_ssl_dns.md#step-3-certification, when I run `zappa certify dev` I get

```
Traceback (most recent call last):
  File "/Users/s/workspace/myapp/api/env/lib/python2.7/site-packages/zappa/cli.py", line 1175, in handle
    sys.exit(cli.handle())
  File "/Users/s/workspace/myapp/api/env/lib/python2.7/site-packages/zappa/cli.py", line 218, in handle
    self.certify()
  File "/Users/s/workspace/myapp/api/env/lib/python2.7/site-packages/zappa/cli.py", line 848, in certify
    s3_client.download_file(bucket, key_name, '/tmp/account.key')
NameError: global name 's3_client' is not defined
```

Looking at the code, it looks like `s3_client` is never actually defined/initialized.

smoll:
Looks like a really simple fix, just trying to wrap my head around how placebo works to see if there's a way to write a failing test case for this before submitting my PR

Miserlou:
Yeah, I'm sure the fix is probably just a `self.` or something like that, but the real issue is that this never had a test. :(

Did you get a chance to look at this?: https://github.com/garnaat/placebo#usage-as-a-decorator

Basically, you write the test, then call, placebo once in 'record' mode, and then that saves all of the communication with AWS as local JSON files for use during offline tests.

Miserlou:
That being said, don't spend _too_ long on it if it'll prevent you from sending a PR with the simple fix! There's one test function with a ton of stuff in it that should probably be broken up, I was just trying to crank the coverage when I wrote that. If it's giving you too much trouble, just send the PR with the fix you've functionally tested manually.

Miserlou:
Fix merged, will be published today.

PR #295, opened 2016-08-31
Fix certification via key hosted in S3

smoll:
As discussed in https://github.com/Miserlou/Zappa/issues/294

Sorry I haven't been able to spend any time digging into placebo today, but I tested that this works functionally on the real Flask API I am working on.

Hopefully I can play around with placebo more and help with that in a subsequent PR. Also, sorry about the stripped trailing spaces in the diff (my editor automatically does that on save).


Miserlou:
Excellent, just waiting for build before merge.

PR #313, opened 2016-09-09
Call django.setup() from zappa only for django < 1.7.0

koriaf:
* because since django 1.7 it leads to double initialization, which is problematic on some installations

To be honest I didn't test it on any django except 1.10, but I am sure that with django 1.9 and 1.10 on my installation it has such problems without this fix. Shall we add tox here?

The error message was about "Populate isn't reentrant".

### DIFF
--- a/zappa/ext/django_zappa.py
+++ b/zappa/ext/django_zappa.py
@@ -1,16 +1,20 @@
+import os
 import sys
 
 # add the Lambda root path into the sys.path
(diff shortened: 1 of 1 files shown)


Miserlou:
Hell yeah, this is great.

This, combined with another change I just made, means also Zappa deployments are now _way the heck faster_! So stoked.

PR #316, opened 2016-09-10
Add support for AWS_IAM authentication

### LINKED ISSUE #314: AWS_IAM authentication
I have an API I'd like to put behind AWS_IAM authentication. I see that `apigateway.put_method` is always called with `authorizationType='none'`. If there's no in-flight change to add support for IAM authentication, I'll submit a PR.

mwedgwood-rmn:
Closes #314


slank:
I'll look at the tests to see about bringing back some coverage.

PR #328, opened 2016-09-18
Use Cloudformation for base API Gateway

mathom:
This is a partial port of https://github.com/Miserlou/Zappa/pull/209

I'm aiming to keep the changes small and just focus on supporting API Gateway updates.


mathom:
I'll need to update the placebo stuff

Miserlou:
It works!!!!

It seems like it takes a long time to create the stack, without any feedback - is there anything we can do there? A progress bar, or something like that?

This is a really promising change, and it's absolutely great that changes to the routes can now happen without a redeploy.

mathom:
You could probably post some dots for progress as it's polling CF but as far as I know there's no way to get an ETA or anything out of the template creation process.

mathom:
You might be able to pull status and see what resources were created vs what you have in the template and calculate a % off that... I'll have to experiment and see what is available when it's in the creation stage.

Miserlou:
Outputting little dots would certainly be better than nothing, it just took so long without any feedback that I thought the process might have hung. I'm tempted to hit the new Code Review button..

mathom:
It now has a shiny progress bar @Miserlou

Miserlou:
I'm really happy with these changes, the only thing left before merge and publish would be to fix the test coverage. Do you want to create the new placebo recordings?

mathom:
I think this PR is basically done unless you guys think it needs anything else.

Miserlou:
YESSSS

going to bed now, merge tomorrow. EXCELLENT WORK

gottk:
Hy guys, I have been using zappa for a few weeks now for our project in the company and built a whole application with it. I used cloudformation already to create dynamodb tables and sqs queues which I need for my framework. But now, as u included a cloudformation as well, I am having troubles to get it running again. Is there a way to disable the cloudformation which gets created by default via Zappa or probably to customise it? Did you had any thoughts about this for the future? As I use Zappa a lot, I would be as well happy to contribute in the future.

Miserlou:
Can you give a bit more information about what's not working? Tracebacks, etc.

gottk:
I cannot provide u currently with a Traceback, as I do not get my pieces to communicate with each other anymore, I can give u an insight of what i built.


I have currently a script which sets up:
- a Cloudformation which creates SQS queues and Dynamodb tables and a policy which provides my role with the access rights to the SQS and Dynamodb
- a Zappa file which obviously includes the function to access the Dynamodb tables and which automatically transfers messages from the DB to the queue (scheduler)
- a Swaggerfile which is used to enable CORS programmatically additionally

but as Zappa creates now as well a Cloudformation I am running into troubles with policies/ access rights ... It would be helpful if users could apply changes to the Cloudformation script via a settings file maybe?

I keep u updated, I am still testing, as finding troubles without errors is very hard.

mathom:
The CF should be pretty independent of any other CF stacks you have built (I'm using it with a couple other stacks simultaneously myself). It's encapsulating the API gateway routes and the rest is handled via manual boto calls from zappa. If you can open an issue with more info about your particular problem that will help @gottk

gottk:
Thanks @mathom! how are u managing different policies with it? The default policy generated with Zappa is just too general in my case.

mathom:
Do you mean IAM policies ? I use in-app authentication and not API gateway auth for all my stuff, so I can't say. It would be easy to add the auth stuff back in - originally I had about 75% of zappa managed by CF. See: https://github.com/Miserlou/Zappa/pull/209

PR #329, opened 2016-09-18
Fixed minor typo in zappa/cli.py

cyhsutw:
Great thanks to all the authors and contributors for this awesome project!

I found a typo in an error message while using zappa with my messy environment, 
and here's the fix 😉


Miserlou:
👍

PR #339, opened 2016-09-22
Simplify API Gateway configuration

koriaf:
Checked it on couple of my installations.
It probably requires redeployment, but `¯\_(ツ)_/¯`
If cloudformation replaces fine then it shall work fine; if cloudformation script have no changes and still sends old parameters then it's also shall be fine.
Also, I use `"http_methods": ["ANY"]`, and I can't see the reason not to make it default behaviour. Just because it N times simplier for API Gateway and CF creation (N equal to current length of this list).

Also I've tried to make better slash handling (true slash passing, not monkeyappending it), but unsuccessfully yet. I'll return to it when I have more time.

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -695,18 +695,15 @@ class Zappa(object):
         self.create_and_setup_methods(restapi, root_id, lambda_arn, api_key_required,
                                       integration_content_type_aliases, authorization_type, 0)
 
-        parent_id = root_id
-        for i in range(1, self.parameter_depth):
-            resource = troposphere.apigateway.Resource('Resource{0}'.format(i))
-            self.cf_api_resources.append(resource.title)
-            resource.RestApiId = troposphere.Ref(restapi)
-            resource.ParentId = parent_id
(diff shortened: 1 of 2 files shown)


zappa/zappa.py
```diff
@@ -752,8 +749,8 @@ def create_and_setup_methods(self, restapi, resource, lambda_arn, api_key_requir
             integration.Credentials = credentials
             integration.IntegrationHttpMethod = 'POST'
             integration.IntegrationResponses = []
-            integration.PassthroughBehavior = 'NEVER'
-            integration.RequestParameters = {}
+            # integration.PassthroughBehavior = 'NEVER'
```
aectann: Why did you drop passthrough behavior? I think that was intentionally added there to make API Gateway return `415 Unsupported Media ` for unexpected mime types..

koriaf: Can't say right now; some side effect of my experiments.
I was able to force ApiGateway to forward real url to Zappa in some conditions, but didn't have the time to finish it. I think it may be returned back.

Miserlou:
Merged and tested, this seems to work awesomely.

Issue #340, opened 2016-09-22
CLI Exits Cleanly When a Non-Zero Should Be Used


mcrowson:
Jenkins is succeeding builds that should fail. This is due to the CLI returning or using a sys.exit(0).

PR #341, opened 2016-09-22
Sysexit nonzero

mcrowson:
Using sys.exit(-1) instead of the mixed return/pass/sys.exit()
#340


mcrowson:
Looks like there was stuff in master that will also get pulled back into dev

PR #347, opened 2016-09-24
Allow usage of an up-to-date version of boto3

### LINKED ISSUE #346: It is not possible to use an up-to-date boto3 version
Zappa filters out the boto3 and botocore package. Unfortunately, the available boto3 version in Lambda is quite old and is lacking some features.

claudeatsafe:
Resolve #346.

boto3 and boto can still be excluded by using the `exclude` setting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ to change Zappa's behavior. Use these at your own risk!
             }
         ],
         "exception_handler": "your_module.report_exception", // function that will be invoked in case Zappa sees an unhandled exception raised from your code
-        "exclude": ["*.gz", "*.rar"], // A list of regex patterns to exclude from the archive
+        "exclude": ["*.gz", "*.rar"], // A list of regex patterns to exclude from the archive. To exclude boto3 and botocore (available in an older version on Lambda), add "boto3*" and "botocore*".
         "http_methods": ["GET", "POST"], // HTTP Methods to route,
         "iam_authorization": true, // optional, use IAM to require request signing. Default false.
         "integration_response_codes": [200, 301, 404, 500], // Integration response status codes to route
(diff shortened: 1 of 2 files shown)


aectann:
Hmm.. I think default behaviour should still be not to include boto, since it's provided by Lambda environment already. And if you need to include a newer version of boto, **that** should be an option.

claudeatsafe:
The problem I see with the current situation is that you cannot write unit tests for your Zappa code and count on that it will run in Lambda. The local environment does not match the remote one. If we decide on not shipping boto3, Zappa should fix its version to match the Lambda version.
Another point, it is an unexpected failure and kind of tricky to figure out what is going on for new users. You might not be aware of the different versions between Lambda and Zappa (there is currently no documentation about that).

What speaks against shipping the version? Does it have a negative impact on the Lambda performance (honestly curious)?

Miserlou:
The filesize. Boto is a pretty large package because it includes some big blobs of JSON and some certificate files. Some Django deploys are already getting pretty close to the filesize limit. I think there are good arguments on both sides here, I'm going to think about it. I'm leaning towards removing boto from the excludes.

I wonder how often Lambda update their boto version..

claudeatsafe:
It added a bit more than 2MB to my ZIP, I think that would be acceptable if it is possible opt-out.

I was wondering that as well, the current version is 1.3.0 which was released in March. As usual with AWS, no information about update cycles (http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html). Wayback machine was also not very helpful to get an idea (https://web.archive.org/web/*/http://docs.aws.amazon.com/lambda/latest/dg/current-supported-versions.html).

PR #348, opened 2016-09-24
Fixes #333

### LINKED ISSUE #333: DynamoDB streams as event source is broken
I could identify two issues, first it seems like in https://github.com/Miserlou/Zappa/blob/master/zappa/util.py#L151 the name should be the Lambda ARN and not the function name (`funk.name = lambda_arn`). Otherwise I get the following error:
```
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/zappa/cli.py", line 1251, in handle
    sys.exit(cli.handle())
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/zappa/cli.py", line 219, in handle
    self.schedule()
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/zappa/cli.py", line 563, in schedule
    events=events
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/zappa/zappa.py", line 1212, in schedule_events
    self.unschedule_events(lambda_name=lambda_name, lambda_arn=lambda_arn, events=events)
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/zappa/zappa.py", line 1358, in unschedule_events
    self.boto_session
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/zappa/util.py", line 192, in remove_event_source
    rule_response = event_source_obj.remove(funk)
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/kappa/event_source/kinesis.py", line 96, in remove
    uuid = self._get_uuid(function)
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/kappa/event_source/kinesis.py", line 36, in _get_uuid
    EventSourceArn=self.arn)
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/kappa/awsclient.py", line 83, in call
    data = results.build_full_result()
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/botocore/paginate.py", line 290, in build_full_result
    for response in self:
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/botocore/paginate.py", line 102, in __iter__
    response = self._make_request(current_kwargs)
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/botocore/paginate.py", line 174, in _make_request
    return self._method(**current_kwargs)
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/botocore/client.py", line 159, in _api_call
  File "/Users/cvessaz/.virtualenvs/tokenserviceenv/lib/python2.7/site-packages/botocore/client.py", line 494, in _make_api_call
    raise ClientError(parsed_response, operation_name)
ClientError: An error occurred (ValidationException) when calling the ListEventSourceMappings operation: 1 validation error detected: Value 'replication.replicate_records' at 'functionName' failed to satisfy constraint: Member must satisfy regular expression pattern: (arn:aws:lambda:)?([a-z]{2}-[a-z]+-\d{1}:)?(\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\$LATEST|[a-zA-Z0-9-_]+))?
```

The second issue is related to the `create_event_permission` call in the `schedule_events` function (https://github.com/Miserlou/Zappa/blob/master/zappa/zappa.py#L1263). It seems like this is not necessary for DynamoDB streams events, it started working once I commented that section out.

Anyways, I do not fully understand yet how this exactly works, so before I open a pull request and break a whole bunch of other things, any idea what is wrong and what is the best way to fix it?

claudeatsafe:
Depends on #347 because I needed an up-to-date boto3 version in my testing environment.

I fixed the issue with DynamoDB streams and also introduced a check if it is already existing to prevent rescheduling stream events (dynamodb, kinesis) on every update. I made the experience that it takes up to 10 minute until a newly added stream processes the first event and it causes that events are duplicated multiple times if the setting `TRIM_HORIZON` is used. All other events still are removed and re-added on every update.
While testing I discovered as well that there was an unnecessary loop which looped over all records and is submitting ALL records for processing in each iteration. This means right now, if you receive a batch of 10 records, the batch is processed 10 times.

As you predicted, this was a lot of trial and error, I hope I did not break anything existing. I tested manually against S3, DynamoDB and CloudWatch events. 
Also, I am fairly new to Python, so let me know if I did something weird.

### DIFF
--- a/README.md
+++ b/README.md
@@ -265,23 +265,6 @@ Similarly, for a [Simple Notification Service](https://aws.amazon.com/sns/) even
        ]
 ```
 
-Please note that your Lambda execution role needs the following permissions on your stream:
-
-```javascript
-{
-  "Effect": "Allow",
-  "Action": [
-    "dynamodb:GetRecords",
-    "dynamodb:GetShardIterator",
-    "dynamodb:DescribeStream",
-    "dynamodb:ListStreams"
-  ],
-  "Resource": {
-    "Fn::Sub": "arn:aws:dynamodb:us-east-1:1234554:table/YourTable/stream/*"
-  }
-}
-```
-
 #### Undeploy
 
 If you need to remove the API Gateway and Lambda function that you have previously published, you can simply:
--- a/README.md
+++ b/README.md
@@ -363,7 +363,7 @@ to change Zappa's behavior. Use these at your own risk!
             }
         ],
         "exception_handler": "your_module.report_exception", // function that will be invoked in case Zappa sees an unhandled exception raised from your code
-        "exclude": ["*.gz", "*.rar"], // A list of regex patterns to exclude from the archive
+        "exclude": ["*.gz", "*.rar"], // A list of regex patterns to exclude from the archive. To exclude boto3 and botocore (available in an older version on Lambda), add "boto3*" and "botocore*".
         "http_methods": ["GET", "POST"], // HTTP Methods to route,
         "iam_authorization": true, // optional, use IAM to require request signing. Default false.
         "integration_response_codes": [200, 301, 404, 500], // Integration response status codes to route
--- a/README.md
+++ b/README.md
@@ -258,7 +258,7 @@ Similarly, for a [Simple Notification Service](https://aws.amazon.com/sns/) even
                "event_source": {
                     "arn":  "arn:aws:dynamodb:us-east-1:1234554:table/YourTable/stream/2016-05-11T00:00:00.000",
                     "starting_position": "TRIM_HORIZON", // Supported values: TRIM_HORIZON, LATEST
-                    "batch  size": 50, // Max: 1000
+                    "batch_size": 50, // Max: 1000
                     "enabled": true // Default is false
                }
            }
--- a/README.md
+++ b/README.md
@@ -218,7 +218,7 @@ In your *zappa_settings.json* file, define your [event sources](http://docs.aws.
             "event_source": {
                   "arn":  "arn:aws:s3:::my-bucket",
                   "events": [
-                    "s3:ObjectCreated:*"
+                    "s3:ObjectCreated:*" // Supported event types: http://docs.aws.amazon.com/AmazonS3/latest/dev/NotificationHowTo.html#supported-notification-event-types
                   ]
                }
             }],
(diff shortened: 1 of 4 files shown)


Miserlou:
This looks great, I have one question about not-rescheduling the events.

If I remember correctly, the purpose of that was to ensure that the function executed was the latest (`$LATEST`) version of the function. 

Since we're not re-scheduling the event source here - which version of the function is executed? The latest, or the version of the code that happened when the event was scheduled?

Miserlou:
I'd also ask that you add a DynamoDB or Kinsesis example in the Zappa documentation, please!

claudeatsafe:
I can confirm that the DynamoDB trigger points to `$LATEST`.
The test was to add the line `logger.info("I am using version {}".format(context.function_version))` to my function and I got the following output:
`[INFO]	2016-09-26T18:11:49.16Z	317f02bd-76de-4301-8244-fd1e6fbd535e	I am using version $LATEST`

I'll push the doc update shortly.

Miserlou:
Okay, great! To confirm: it's _always_ $LATEST, even after the code has been updated since then?

claudeatsafe:
Yes, that's what it seems to be. I got the log line after I used `zappa update` with the new behaviour (so the stream did not get recreated) and the output points to $LATEST. I think if no version is specified in the ARN it defaults to $LATEST.
From the [API doc](http://docs.aws.amazon.com/lambda/latest/dg/API_CreateEventSourceMapping.html#API_CreateEventSourceMapping_RequestSyntax):
> If you are using versioning, you can also provide a qualified function ARN (ARN that is qualified with function version or alias name as suffix). For more information about versioning, see AWS Lambda Function Versioning and Aliases

Miserlou:
In the docs, you use say "batch  size", with two spaces - should that be "batch_size"?

Similarly, should we add the new permissions to the default execution role rather than having it be a documentation requirement? I'd greatly prefer that Zappa JustWork^tm.

claudeatsafe:
Good catch, not sure how the spaces ended up there 😳 

I just realized, the default Zappa role comes already with full permissions on all DynamoDB:
```
        {
            "Effect": "Allow",
            "Action": [
                "dynamodb:*"
            ],
            "Resource": "arn:aws:dynamodb:*:*:*"
        }
```
So it should just work. Should I remove the section again from the docs?

Miserlou:
Yeah, if no changes to the role are need then I think you can take that out.

Issue #349, opened 2016-09-24
Add support for API Gateway Authorizers


claudeatsafe:
I thought it would be neat if I could configure more complex token validations (e.g. look it up in a DynamoDB table) and have the logic directly in Zappa. The validation itself is fairly well documented from AWS (http://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html, https://github.com/awslabs/aws-apigateway-lambda-authorizer-blueprints/blob/master/blueprints/python/api-gateway-authorizer-python.py), so it should not be a problem to detect a request from an Authorizer against Zappa and point it to the right function (similar to events).

I implemented a proof of concept to Zappa which supports authorizer events but does not create the authorizers yet. Before I keep going I will open a pull request with what I have so far to initiate a discussion regarding the implementation.

Miserlou:
This looks like a very interesting idea and a good effort so far!

I have to ask: why should this be handled this way, and not by the application itself? Is the primary benefit being able to use authorization as an event source?

claudeatsafe:
The main advantage I see is integrated caching, you can define how long a token should be valid and AWS takes care of that. So let's say I am using DynamoDB as a token store for an API endpoint powered by Zappa I do not have to query DynamoDB for every single API request. 
Another thing is, you can define quite fine grained access permissions. I honestly do not quite understand yet how they work exactly, but I am sure they have some advantages as well 😄 

In the long run, you could go even further and implement the blueprint for the authorizers and Zappa users would only have to define a method to give access details and user ID and Zappa would take care of the policy generation. This would be part of another sprint though.

claudeatsafe:
#350 got merged, closing.

smoll:
Quick question for @claudeatsafe or anyone else who has used this. I've been able to set up a custom authorizer and it works great, however I'm struggling to figure out how I can use the `context` (https://github.com/awslabs/aws-apigateway-lambda-authorizer-blueprints/blob/master/blueprints/python/api-gateway-authorizer-python.py#L71) that we set in the `authResponse`.

According to the docs (http://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html#api-gateway-custom-authorizer-output and http://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference) we should be able to access it via `$context.authorizer.property`, but I have no clue how to map that to a variable we can access in Python/the Flask app.

Any ideas?

[...]

PR #350, opened 2016-09-24
Add support for API Gateway Authorizers

claudeatsafe:
Belongs to #349.

This is a proof of concept for a possible implementation of API Gateway Authorizers support in Zappa. It currently only allows the processing of Authorizer requests but does not create resources yet. I added all potentially necessary settings to the README.

To finish this implementation, the following extra features need to be added:

- [x] Add authorizers to CloudFormation template
- [x] The default Zappa role permissions need to be updated
- [x] Tests
- [x] Documentation with use cases

@Miserlou what do you think, is this something you can see in Zappa?


Miserlou:
Looks like this is coming along nicely! Let me know when you're ready for a review!

claudeatsafe:
@Miserlou I think this is good for a review.

If you would like to test it, just use the blueprint function from AWS and replace ` policy.denyAllMethods()` with `policy.allowAllMethods()`.
I was planning on adding this function with allowAll to the examples but I was not sure about copy/pasting it because of the Apache 2.0 license.

Miserlou:
I don't think there's a license problem there, just do it. Looks like one of the README sections is empty?

claudeatsafe:
I think that was a caching issue of Github.

Added the example.

claudeatsafe:
@Miserlou any idea when you're going to look at this?

Miserlou:
Hey, sorry about sleeping on this. Reviewing again now.

README.md
```diff
@@ -381,7 +383,13 @@ to change Zappa's behavior. Use these at your own risk!
...
         "http_methods": ["GET", "POST"], // HTTP Methods to route,
-        "iam_authorization": true, // optional, use IAM to require request signing. Default false.
+        "iam_authorization": true, // optional, use IAM to require request signing. Default false. Note that enabling this will override the authorizer configuration.
+        "authorizer": {
+            "function": "your_module.your_auth_function", // Required. Function to run for token validation. For more information about the function see below.
+            "result_ttl": 300, // Optional. Default 300. The time-to-live (TTL) period, in seconds, that specifies how long API Gateway caches authorizer results. Currently, the maximum TTL value is 3600 seconds.
+            "token_source": "Authorization", // Optional. Default 'Authorization'. The name of a custom authorization header containing the token that clients submit as part of their requests.
+            "validation_expression": "xxx", // Optional. A validation expression for the incoming token, specify a regular expression.
```
Miserlou: 'xxx' looks like a placeholder - is there a better example we can give here?

README.md
```diff
@@ -588,6 +607,10 @@ To manually define the permissions policy of your Zappa execution role, you must
 
 Ongoing discussion about the minimum policy requirements necessary for a Zappa deployment [can be found here](https://github.com/Miserlou/Zappa/issues/244).
 
+#### Using API Gateway Authorizer
+
```
Miserlou: This section is still incomplete?

zappa/wsgi.py
```diff
@@ -100,6 +105,9 @@ def create_wsgi_request(event_info, server_name='zappa', script_name=None,
             if script_name in path_info:
                 environ['PATH_INFO'].replace(script_name, '')
 
+        if remote_user:
+            environ['REMOTE_USER'] = remote_user
```
Miserlou: I like this

zappa/zappa.py
```diff
@@ -697,13 +705,36 @@ def create_api_gateway_routes(self, lambda_arn, api_name=None, api_key_required=
...
+    def create_authorizer(self, restapi, uri, identity_source, authorizer_result_ttl, identity_validation_expression):
+        """
+        Create Authorizer for API gateway
+        """
+        if not self.credentials_arn:
+            self.get_credentials_arn()
+
+        authorizer_resource = troposphere.apigateway.Authorizer("Authorizer")
+        authorizer_resource.RestApiId = troposphere.Ref(restapi)
+        authorizer_resource.Name = "ZappaAuthorizer"
```
Miserlou: Will this create a collision if you have multiple applications with authorizers on the same account?

Maybe should be project_name + "ZappaAuthorizer" ?

README.md
claudeatsafe: No, it is my proposal for the doc 😊  My thoughts were that it is an advanced feature and a user will have to read the doc from AWS to implement it. So instead of copying I tried to give a short introduction and give links to docs and examples.
Is there something specific you are missing?

zappa/zappa.py
claudeatsafe: I just tested it and there is no conflict between APIs because Authorizers belong to one specific API. This would become a problem if Zappa would support more than one Authorizer per project, but I can't see how this would be implemented, so the name should be fine. I am always careful with adding strings with unlimited length to AWS resources as you suddenly might run into a length limitation from AWS, if you are fine I'd like to keep the Authorizer name as it is now.

README.md
Miserlou: Is a commit missing? I'm confused. If I look at your doc, there is just a random section header that says `Using API Gateway Authorizer` but there is nothing below it, just the next section?

claudeatsafe: Gosh, I should read more carefully, this is a leftover, sorry about that. The doc is in the **Enabling Secure Endpoints on API Gateway** section above. Will remove it.

zappa/zappa.py
```diff
@@ -1088,7 +1088,7 @@ def update_stack(self, name, working_bucket, wait=False, update_only=False):
                 if not result['Stacks']:
                     continue  # might need to wait a bit
 
-                if result['Stacks'][0]['StackStatus'] == 'CREATE_COMPLETE':
+                if result['Stacks'][0]['StackStatus'] in ['CREATE_COMPLETE', 'UPDATE_COMPLETE']:
```
Miserlou: huge 👍

README.md
```diff
@@ -388,7 +388,7 @@ to change Zappa's behavior. Use these at your own risk!
             "function": "your_module.your_auth_function", // Required. Function to run for token validation. For more information about the function see below.
             "result_ttl": 300, // Optional. Default 300. The time-to-live (TTL) period, in seconds, that specifies how long API Gateway caches authorizer results. Currently, the maximum TTL value is 3600 seconds.
             "token_source": "Authorization", // Optional. Default 'Authorization'. The name of a custom authorization header containing the token that clients submit as part of their requests.
-            "validation_expression": "xxx", // Optional. A validation expression for the incoming token, specify a regular expression.
+            "validation_expression": "^Baerer \\w+$", // Optional. A validation expression for the incoming token, specify a regular expression.
```
Miserlou: * Bearer :)

Miserlou:
Let's go!

Issue #359, opened 2016-09-26
Allow Scheduled Events to take Arguments


mcrowson:
in the zappa_settings.json, allow scheduled events to take arguments. This would possible include another key to the event dict possibly.

```
{
     "expression": "rate(10 minutes)",
     "function": "path.to.funk",
     "function_kwargs": {"key": "val"}
}
```

Miserlou:
This seems like a handy feature, but is this even possible with scheduled event sources?

mcrowson:
It looks like it would be passed as an input with the target:

https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L291 
http://boto3.readthedocs.io/en/latest/reference/services/events.html#CloudWatchEvents.Client.put_targets

But that then overwrites the event info created by cloudwatch and gets rid of things like 'detail-type' which makes the handler not able to deal with the event.

https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L291

I'll keep digging, but so far this isn't promising.

mcrowson:
This might be possible now that CW Put-Targets supports InputTransformation: 

https://aws.amazon.com/about-aws/whats-new/2017/03/amazon-cloudwatch-events-now-supports-input-transformation-of-events/

I'll take another look into this if someone doesn't get to it first.

PR #360, opened 2016-09-26
Feature/event kwargs

mcrowson:
#359

This allows for a dict to be passed to an event. Doing so causes us to lose some of the event arguments passed by CW. Not sure if this loss is acceptable. The following keys are no longer in the event. 

['account', 'region', 'detail', 'version', 'time', 'id']

These are not used by Zappa, and won't break existing events because the function_kwargs argument wouldn't be in the settings file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Zappa Changelog
 
+## 0.25.0
+
+* Add ability to invoke raw python strings, like so:
+
+    `zappa invoke dev "print 1+2+3" --raw`
+
+* Fixes multi-argument `manage` commands.
+* Updated related documentation.
+
 ## 0.24.2
 
 * Fix a problem from trying to `update` old-style API routes using new Tropo code. Ensures `touch` works as intended again. Fix by @mathom.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 * Fixes multi-argument `manage` commands.
 * Updated related documentation.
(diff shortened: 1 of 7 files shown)


Miserlou:
Time doesn't work like that, that'll just pass the time that the event is scheduled rather than the time of the event itself?

mcrowson:
Yup, good call.

Miserlou:
I'm not sure what to do about this one, to be honest.

Can you describe the use case to me?

The problem I'm seeing is that the benefit of having kwargs at the cost of losing region, version and time information (especially for global deployments, scientific/medical deployments which need log data, etc.), doesn't seem worth the cost. 

Especially because the kwargs would be static anyway, it seems like the better solution would simply be to write a function with those static kwargs already in it, and then use that as the event source. Does that make sense?

mcrowson:
That function solution is what we're implementing now, but it just feels strange. Pre-zappa we had a function that would go and get data from a source and parse/store it. The source was an input to the function. Rewriting this into many functions to accomodate all of the sources didn't feel right.

There may be other calls that we could make to get the region and version. I'm not sure how to handle time though.

mcrowson:
I think we now have everything except for the cloudwatch event id that triggers the rule. One other thing is that the time variable is created by Zappa, not by cloudwatch. As these are only for Scheduled Event CW types I don't see this as being a problem.

mcrowson:
@Miserlou Any more thoughts on this?

mcrowson:
Is there something that could get this through, or is it just too hackish with the way lambda takes arguments currently?

mcrowson:
Not needed with #1091

PR #365, opened 2016-09-29
(doc) change to os.environ.get

scoates:
super simple doc change to fix accidental use of `os.environ` as a method


Miserlou:
Please change this in the README file, not just the docs folder. README is the One True Doc.

scoates:
Done. Didn't even notice it there. Good catch.

PR #375, opened 2016-10-02
Correct parsing of S3 key URI.

### LINKED ISSUE #374: certify can fail for lets_encrypt_key in S3
The `zappa certify` command fails
```
ClientError: An error occurred (404) when calling the HeadObject operation: Not Found
```
given the setting
```
        "lets_encrypt_key": "s3://my-bucket/my-folder/server.key",
```
because the key in the S3 URI is extracted incorrectly if it has slashes.

jlevy:
Fixes #374.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -947,8 +947,7 @@ class ZappaCLI(object):
             return
 
         if 's3://' in account_key_location:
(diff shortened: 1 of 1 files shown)


Miserlou:
Ugh, good catch. This is still the least-tested part of the project. :(

PR #396, opened 2016-10-14
make sure that environment variables are strings

scoates:
I accidentally set an integer environment variable, and Zappa deployed properly, but failed on request. This showed up in the `tail`:

```
Zappa Event: {}
[1476408522194] putenv() argument 2 must be string, not int: TypeError
  File "/var/task/handler.py", line 476, in lambda_handler
  File "/var/task/handler.py", line 187, in lambda_handler
    handler = cls()
  File "/var/task/handler.py", line 109, in __init__
    os.environ[key] = self.settings.ENVIRONMENT_VARIABLES[key]
  File "/usr/lib64/python2.7/os.py", line 473, in __setitem__
    putenv(key, item)
TypeError: putenv() argument 2 must be string, not int
```

This PR checks the `environment` for strings and errors out if they're not.

### DIFF
--- a/tests/test_bad_environment_vars.json
+++ b/tests/test_bad_environment_vars.json
@@ -1,0 +1,19 @@
+{
+    "ttt888": {
+       "touch": false,
+       "s3_bucket": "lmbda",
+       "app_function": "tests.test_app.hello_world",
+       "callbacks": {
+           "settings": "test_settings.callback",
+           "post": "test_settings.callback",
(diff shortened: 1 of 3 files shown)


Miserlou:
Good catch! This looks like it would be easy to write a test for! :)

scoates:
As you wish. (-:

Miserlou:
👍

PR #408, opened 2016-10-18
Clarify contribution workflow

aehlke:
## Description

As discussed in #407, Zappa's current contribution workflow is to target PRs against `master`, until automation can get set up to ensure `dev` doesn't fall behind `master`.

Keywords for these two workflows are "GitHub Flow" (master + PRs) vs. "GitFlow" (master to represent the latest release, dev for trunk + PRs against dev, or against master for "hotfixes").

Also added a PR template to cover the linked issue contribution requirement.

## GitHub Issues

#407

Issue #407, opened 2016-10-18
Continuous Deployment with Zappa


aehlke:
We're continuously-deploying the services we're using Zappa for: on push to GitHub, our Jenkins 2 server gets pinged and once the test suite passes, the service gets automatically deployed to production.

This is slightly aspirational but we're close to having this set up. I'd like to document the pain points for potential followup work in Zappa.

The first one we've run into is that the Zappa command to deploy depends on whether the CloudFormation stack already exists - on initial deploy, it's `zappa deploy`, and then `zappa update` after that. Since `zappa deploy` gives exit code 0 whether or not it succeeded, and since its output is not machine-readable, we had to resort to something like this in our automation:

```
if zappa deploy | tee /dev/stderr | grep 'already deployed'; then
    zappa update
fi
```

... which is fragile and unintuitive. `zappa status` also gives exit code 0 whether or not the stack exists. The quick fix could be to return a non-zero exit code. Since Zappa configuration is descriptive, it could also be nicer to consolidate `deploy` and `update` into one command that "applies" the described service, much like you'd do with configuration management tools like Ansible.

I'd also be interested to hear about anyone else's experience with continuously-deploying Zappa services.

Miserlou:
Nice, this is an excellent issue!

Changing the status codes would be trivial. Another option would be putting the `status` output in a machine readable format (ex with `--json`), so that you could simply parse it with `jq`.

Combining `deploy` and `update` is a different discussion, but much more feasible now that we are updating the API Gateway with CF stacks rather than boto API calls. [That being said, I'm actually not convinced that CF stacks are the way to go now that we have APIG configured to use Proxy+].

aehlke:
Sounds like the status code change makes sense first, and then we can add e.g. `--json` on top of that without having to throw away any work.

Surprised to hear you're thinking to move away from CloudFormation - I don't know what Proxy+ is so I guess I'll dig into that. Since we've begun adopting CF via Troposphere, we're now moving everything in that direction since it makes so many operational needs simpler and safer. Happy to discuss this one more once I catch up on the context.

Miserlou:
Previously we moved from boto to CF because the API Gateway templates were HUGE (~500Kb), but with a new API Gateway feature, they're down to <1.5kKb. More info here: https://blog.zappa.io/posts/unhacking-zappa-with-new-apigateway-features

[...]

Issue #243, opened 2016-08-12
Zappa needs Sphinx docs on ReadTheDocs.org 


bjinwright:
I aim to make that happen.

bjinwright:
#

jotes:
@bjinwright I think we can close this one :+1:

Miserlou:
Leave this open until we have documentation how to contribute/compile/publish the docs please!

bxm156:
Was thinking, should we just delete the /docs so that user's don't get confused? Sounds like the README is place for documentation.

aehlke:
@bxm156 I attempted to move in the other direction here https://github.com/Miserlou/Zappa/pull/410 so going the other way and deleting `/docs` sounds good. Do we have anything in `/docs` that needs to get backported though?

bxm156:
@aehlke  The last notable change to /docs was in Mar 23, 2017. Some of them were last touched a year ago, and they aren't being published on readthedocs.org either. 

I don't think there is anything in them that needs backported, I'd say the goto place for documentation has always been the readme. In fact, some of the documentation in /docs is so old that it simply just incorrect for today's version of zappa. @bjinwright  did some work on the /docs  so he might have some thoughts on this.

bxm156:
I think there are a few examples for DNS/domain setup in /docs that are linked to from the README. I think we can just integrate them in to the README

PR #410, opened 2016-10-18
Consolidate contrib docs and fix link

aehlke:
## Description
- Fixes link to testing doc
- Consolidates the two places we document contributions

## GitHub Issues
Some discussion on updating contrib docs in #407 and the initial PR was in #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -661,17 +661,7 @@ Zappa goes quite far beyond what Lambda and API Gateway were ever intended to ha
 
 ## Contributing
 
-This project is still young, so there is still plenty to be done. Contributions are more than welcome!
-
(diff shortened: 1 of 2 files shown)


Miserlou:
I prefer having it in a single document, sorry about that. I always try to minimize the amount of cruft and legacy in a project..

aehlke:
> I prefer having it in a single document, sorry about that. I always try to minimize the amount of cruft and legacy in a project..

Would you prefer removing the 2nd contrib.rst doc then? I can consolidate in the other direction :)

Miserlou:
Oh man, I didn't even know about that file. 

This is in fact directly related to #243 - basically somebody needs me to teach me how and why to publish to ReadTheDocs..

PR #415, opened 2016-10-19
Update current list of supported regions

aehlke:
## Description
Adds missing regions

## GitHub Issues
*I decided against opening an issue first since this seems obvious enough. LMK if you'd like to discuss.*

See https://docs.aws.amazon.com/general/latest/gr/rande.html#apigateway_region and https://docs.aws.amazon.com/general/latest/gr/rande.html#lambda_region


Miserlou:
w00t, hello Ohio!

Issue #413, opened 2016-10-19
Default region should be from profile, not us-east-1


Miserlou:
Somebody mentioned this on Slack, I think it's a good idea.

atif1996:
Very true. I was stuck for over a day on that.

Miserlou:
👍

PR #416, opened 2016-10-19
Feature/region from profile

mcrowson:
## Description
Get region from profile 

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/413


Miserlou:
Great, thank you for this!

PR #417, opened 2016-10-19
Include kinesis permissions to allow suscribing to a stream

aectann:
## Description
Adds kinesis permission to lambda execution policy managed by zappa. This is required to allow zappa to subscribe for a kinesis stream. 

Without it following message is printed to console on schedule:

> ClientError: An error occurred (InvalidParameterValueException) when calling the CreateEventSourceMapping operation: Cannot access stream arn:aws:kinesis:ap-southeast-2:032345039960:stream/d-facebook-bot-notifications. Please ensure the role can perform the GetRecords, GetShardIterator, DescribeStream, and ListStreams Actions on your stream in IAM.

Sample event definition:

```
"events": [
         {
            "function": "app.events.run",
            "event_source": {
              "arn": "arn:aws:kinesis:ap-southeast-2:032345039960:stream/d-incoming-notifications",
              "starting_position": "TRIM_HORIZON",
              "batch  size": 1,
              "enabled": true
            }
         }
       ]
```


Miserlou:
Good and necessary PR, please re-record with Placebo if you can.

Miserlou:
What is your use case for Kinsesis and Zappa? I've been wanting to use that combination for something but I don't know enough about using Kinesis in the field.

aectann:
@Miserlou updated tests. 

We use kinesis as event bus at the moment to process notification from various sources. While an SNS topic subscription would've done the same, kinesis offers replay functionality and data retention, so we decided to try that out first.

Miserlou:
Excellent! And an interesting use case, would love to know more once it's been battle tested.

Issue #423, opened 2016-10-23
Make Warnings Fatal


Miserlou:
People aren't reading the warnings. We should make them `--ignore-warnings`.

mcrowson:
This might cause breakages in deployment pipelines where certain warnings are being ignored already (such as the zip size as I saw in one of the unit tests).

PR #425, opened 2016-10-24
warnings cause system exit unless --ignore-warnings are passed.

mcrowson:
## Description
Warnings now cause a system exit unless --ignore-warnings is passed to the CLI.

## GitHub Issues

#423


jneves:
@mcrowson Can you check what the situation is with the tests, please?

Issue #400, opened 2016-10-17
The 'zappa' distribution was not found and is required by the application


Miserlou:
This is an annoying warning that you get if you're using a local version of Zappa for devving/testing. It's in the version updated checker code.

mcrowson:
I'm not getting this error. How are you using a local version of zappa? I have been using setuptools.

Miserlou:
This is just if you use a local Zappa with a symlink.

PR #426, opened 2016-10-24
setting the version differently so we can import it differently.

mcrowson:
## Description

https://packaging.python.org/single_source_version/
Using approach 3
## GitHub Issues

#400


mcrowson:
Closing because its ancient and #871 seems to have picked up the slack for me

Issue #404, opened 2016-10-18
Command-line subcommands don't support --help


tuukkamustonen:
Neither `zappa deploy --help` nor `zappa --help deploy` print out help for that command in specific.

skarale25:
@tuukkamustonen When I run same command, I get following output
$ zappa deploy --help
usage: zappa [-h] [-n NUM_ROLLBACK] [-s SETTINGS_FILE] [-a APP_FUNCTION] [-v]
             [-y] [--remove-logs] [--raw] [--no-cleanup] [--all]
             [U [U ...]]

Zappa - Deploy Python applications to AWS Lambda and API Gateway.

positional arguments:
  U                     Please supply a command to execute. Can be one of:
                        certify, deploy, init, invoke, manage, rollback,
                        schedule, status, tail, undeploy, unschedule, update

optional arguments:
  -h, --help            show this help message and exit
  -n NUM_ROLLBACK, --num-rollback NUM_ROLLBACK
                        The number of versions to rollback.
  -s SETTINGS_FILE, --settings_file SETTINGS_FILE
                        The path to a zappa settings file.
  -a APP_FUNCTION, --app_function APP_FUNCTION
                        The WSGI application function.
  -v, --version         Print the zappa version
  -y, --yes             Auto confirm yes
  --remove-logs         Removes log groups of api gateway and lambda task
                        during the undeployment.
  --raw                 When invoking remotely, invoke this python as a
                        string, not as a modular path.
  --no-cleanup          Don't remove certificate files from /tmp during
                        certify. Dangerous.
  --all                 Execute this command for all of our defined Zappa
                        environments.
Do you get same output? Correct me if I am getting wrong.

tuukkamustonen:
@skarale25 Yeah, that's the output (btw. you should [format your paste](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting)).

Miserlou:
@skarale25 - he wants specific help for each sub command.

rozuur:
Is this open? Can I work on this?

PR #432, opened 2016-10-26
Feature/cli rework

mcrowson:
## Description

Redid the CLI so we could have help on the commands. A few questions:
1. The test_cli_args currently just returns and prints "The command is not recognized" The rework causes this to fail because derp was never a command. If the intention was to exit, i have modified the test to grab that exit. If the intent was to succeed, let me know what success looks like and I can update the test so it doesn't exit.
2. Do both -s and -a arguments come before the command argument?
3. Do both -y and --all arguments come after the command argument?
## GitHub Issues

#404

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -54,21 +54,6 @@ CUSTOM_SETTINGS = [
     'touch',
 ]
 
-CLI_COMMANDS = [
-    'certify',
-    'deploy',
-    'init',
-    'invoke',
-    'manage',
-    'rollback',
-    'schedule',
-    'status',
-    'tail',
-    'undeploy',
-    'unschedule',
-    'update'
-]
-
 ##
 # Main Input Processing
 ##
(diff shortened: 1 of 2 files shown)


mcrowson:
Looks like it has issues with how we're getting the versions. This might be easier to merge after https://github.com/Miserlou/Zappa/pull/426 or implement a different version method.

mcrowson:
It looks like there is interest in progressing the CLI, If this PR seems more appealing than #519 I can bring it up to date with master.

zappa/cli.py
```diff
@@ -132,49 +117,84 @@ def handle(self, argv=None):
         Parses command, load settings and dispatches accordingly.
 
         """
-        help_message = "Please supply a command to execute. Can be one of: {}".format(', '.join(x for x in sorted(CLI_COMMANDS)))
-
-        parser = argparse.ArgumentParser(description='Zappa - Deploy Python applications to AWS Lambda and API Gateway.\n')
-        parser.add_argument('command_env', metavar='U', type=str, nargs='*', help=help_message)
-        parser.add_argument('-n', '--num-rollback', type=int, default=0,
-                            help='The number of versions to rollback.')
+        parser = argparse.ArgumentParser(description='Zappa - Deploy Python applications to AWS Lambda and API Gateway.\n', version=pkg_resources.require("zappa")[0].version, prog="zappa")
```
rmoe: Neat, I've never seen version= used here to provide a version argument. TIL.

mcrowson: yea, argparse has version built in. https://docs.python.org/2.7/library/argparse.html#action

mcrowson:
@rmoe is taking over this stuff with #519

PR #442, opened 2016-11-01
Make sure path info is unquoted before evnironment down to the wsgi app

aectann:
## Description
Uses same logic as werkzeug when preparing environment for wsgi request. Url decode path parameters so that underlying wsgi app doesn't have to cater for that.


Miserlou:
👍

Issue #383, opened 2016-10-06
Make project_name A Required Setting


mcrowson:
The project_name currently defaults to pwd. When have multiple working environments for developers or CI servers this can lead to deployment issues that are not straight forward to troubleshoot. If it stays as an optional setting, maybe update the warning from:

Warning! Couldn't get function - have you deployed yet?

to

Warning! Couldn't get function - It is not deployed or was deployed under a different project_name.

Miserlou:
Did you set the project up using `init`? It could be put in there but I don't think it needs to be a required setting.

mcrowson:
I did not use init, just made the JSON. I think putting it in init, updating that warning, and possibly adding a warning on deploy as well if project_name isn't specified in the settings. I would think most prod-level deployments will come across this scenario at some point so it would be good to communicate the project_name importance.

PR #447, opened 2016-11-02
added region and lambda name to "not deployed" error.

mcrowson:
## Description
Added the lambda name and region name to the "not deployed" error. This error sometimes comes up when different environments are deploying and updating Zappa installations and the project_name or aws_region are determined automatically through environment settings and folder names. 

Hopefully this will help debug these errors.

## GitHub Issues

#413 after 0.28 and region was not being assumed us-east-1 by Zappa. Discussed in Slack
#383


aectann:
Looks great, thanks!

PR #448, opened 2016-11-02
Delete the right domain name

kwellman:
Api gateway domain doesn't get deleted as intended. "blog.zappa.io" seems to be hardcoded by accident.


Miserlou:
THAT EXPLAINS A LOT

PR #451, opened 2016-11-03
create A record for apex domains

kwellman:
## Description

Create an "Alias" record to handle apex domains. 

Parameters for record was set by following http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-aliastarget.html

- EvaluateTargetHealth must be False when alias target is a cloudfront distribution (says so in the console).
- HostedZoneId should be set to Z2FDTNDATAQYW2 for cloudfront targets (see AWS link above)

## GitHub Issues

#448

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -1238,6 +1238,29 @@ class Zappa(object):
             stage=stage
         )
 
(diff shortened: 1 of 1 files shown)


Miserlou:
Wow, Z2FDTNDATAQYW2 looks like some of the darkest AWS voodoo I've ever seen. Amazing work, will test this as soon as possible.

PR #460, opened 2016-11-07
Filter private hosted zones to avoid conflicts while certifying

### LINKED ISSUE #459: Having two hosted zones (one public and one private) with same name makes certify with let's encrypt (DNS auth) fail
When you have two hosted zones (one public and one private) with the same name under the same registered domain, deploying to this domain with let's encrypt certificate using DNS auth doesn't work.

What happens is that the `_acme-challenge.XX.XX` link always end up in the private hosted zone. Therefore let's encrypt can't access it so the validation process fails with error:
`DNS problem: NXDOMAIN looking up TXT for _acme-challenge.XX.XX`

I guess it would be good if Zappa was able to handle this particular case.

Solution seems simple:
In `get_best_match_zone()` in `zappa.py` we just need to check if the matched zone is private or not, then continue until finding a public one and return None (or fail ?) if there isn't any public hosted zone for the domain. 
If the issue is agreed I can work on it.

acharp:
## Description
Tested functionally it works. Even with two hosted zones (private and public) having the same name behind a same domain, now Zappa manages to select the public hosted zone, certify and expose the custom domain name.
Though I didn't manage to run the tests of the project :(

## GitHub Issues
Resolve #459

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -862,7 +862,10 @@ class TestZappa(unittest.TestCase):
         zone = Zappa.get_best_match_zone(all_zones={ 'HostedZones': [
             {
                 'Name': 'example.com.au.',
-                'Id': 'zone-correct'
+                'Id': 'zone-correct',
+                'Config': {
+                    'PrivateZone': False
+                }
             }
         ]},
             domain='www.example.com.au')
@@ -872,7 +875,10 @@ class TestZappa(unittest.TestCase):
         zone = Zappa.get_best_match_zone(all_zones={'HostedZones': [
             {
                 'Name': 'example.com.au.',
-                'Id': 'zone-incorrect'
+                'Id': 'zone-incorrect',
+                'Config': {
+                    'PrivateZone': False
+                }
             }
         ]},
             domain='something-else.com.au')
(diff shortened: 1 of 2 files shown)


Miserlou:
Looks good, can you confirm that _all_ zones will have the PrivateZone field? Else `zone['Config']['PrivateZone']` will crash

acharp:
Yep, I can confirm that. It's a mandatory field from boto3 response therefore always there: `True` for private zones and `False` for public ones.

Miserlou:
Ah, okay.

I think the only thing left to do here is to add test Config fields to the test in tests/tests.py#860.

Why can't you get the tests to run? You should just be able to install `pip install -r test_requirements.txt; ./test.sh` - if that doesn't work, then that's a separate bug..

acharp:
Yes that is what I was doing but it fails because my region profile is eu-west-1:
`self.assertEqual(z.boto_session.region_name, 'eu-west-1')
AssertionError: 'us-east-1' != 'eu-west-1'`
Anything to do to avoid that and being able to run the tests locally ?

Thought I didn't notice that in Travis it was failing because of missing key Config. I'll look at it closer and try to add some tests.

acharp:
Looks better with testing :)

Nevertheless I didn't find why I'm not able to run the tests locally. 
It still fails with `self.assertEqual(z.boto_session.region_name, 'eu-west-1') AssertionError: 'us-east-1' != 'eu-west-1'`
I will investigate that further when I have a room.

Miserlou:
Gorgeous! Tusen takk!

PR #471, opened 2016-11-14
Only a single event was being output from status.

mcrowson:
## Description
After moving to a key driven status, having multiple events caused this to just overwrite and only show a single event at a time. This creates a list of dicts to store the data in.


Miserlou:
:taco:

Issue #477, opened 2016-11-17
Unable to use subfolders in S3 for storing access key for Let's Encrypt


khaliullov:
Hello, I've tried to store access.key of Let's Enctypt in S3 bucket's subfolder, but got 404 error, because S3 URL was incorrectly parsed. Instead of using of /full/path used just first node because they splitted by slash ('/')

Miserlou:
Good catch, want to submit a PR? 👍

Miserlou:
Woah, you submitted one before I even finished typing my comment! 💯

Miserlou:
Since we're actually going to be re-using this code for #456, would you mind refactoring this into a utility function and adding a test for different key paths? Thanks!

khaliullov:
@Miserlou OK, I'll try.

khaliullov:
@Miserlou added `parse_s3_url` function

PR #478, opened 2016-11-17
allow to use s3 subfolders for storing Let's Encrypt key

khaliullov:
## Description
split only bucket and path in lets_encrypt_key S3 URL.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/477


Miserlou:
This looks excellent, thank you very much!

Issue #473, opened 2016-11-15
Can't call lambda function for AWS S3 events


AbsolutOD:
I have been trying to create a lambda function that will parse cloudwatch logs for certain events and react on them.  I have just playing around to see what is in the event object coming from S3.  When I tail the logs, I can see the debug messages, but then I keep getting the following stack trace.

```python
[1479156641917] need more than 1 value to unpack: ValueError
Traceback (most recent call last):
  File "/var/task/handler.py", line 477, in lambda_handler
    return LambdaHandler.lambda_handler(event, context)
  File "/var/task/handler.py", line 189, in lambda_handler
    return handler.handler(event, context)
  File "/var/task/handler.py", line 349, in handler
    app_function = self.import_module_and_get_function(whole_function)
  File "/var/task/handler.py", line 179, in import_module_and_get_function
    module, function = whole_function.rsplit('.', 1)
ValueError: need more than 1 value to unpack
```

It has been pretty puzzling.  I can invoke the function from the command line, but all S3 events fail with the above stack trace.

https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L343

I dug into the code and found why I am getting this error.  In the handler.py on line 343 it checks to see if the event has a 'Records' key and then calls `get_function_for_aws_event`.  The problem is that function returns the value in `event['s3']['configurationId']` which is returning the lambda's ARN (arn:aws:lambda:us-east-1:XXXXXXX:function:helloevent-dev).  The ARN causes the exception in `import_module_and_get_function`.

https://github.com/Miserlou/Zappa/blob/master/zappa/util.py#L99

When the S3 event is registered it names it after the lambda_arn.  The function `utility.get_event_source` takes the target_function as a param, but doesn't actually use it.  It sets the `funk.name = lambda_arn`.

https://github.com/Miserlou/Zappa/blob/master/zappa/util.py#L150

I think the the funk.name should be set to the target_function.  I am going to play around with it today.  I just wanted to post the issue in case someone had a better fix.

bjinwright:
I can confirm this is happening.

Miserlou:
Is this a regression?

bjinwright:
Yep it happens in 0.27 as well

AbsolutOD:
Yeah, I got a slightly different error.

```python
[1479240657613] need more than 1 value to unpack: ValueError
Traceback (most recent call last):
  File "/var/task/handler.py", line 476, in lambda_handler
[...]

PR #482, opened 2016-11-18
fixed s3 event trigger function name

AbsolutOD:
## Description
I changed the funk name to be the <project_name>:<event function>.  And I changed the get_function for AWS_event function to parse the record['s3']['configurationId'] properly.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/473

* changed the get_function_for_aws_event to returen the function part of the
  record['s3']['configurationId']

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -258,7 +258,7 @@ class LambdaHandler(object):
         Support S3, SNS, DynamoDB and kinesis events
         """
         if 's3' in record:
-            return record['s3']['configurationId']
+            return record['s3']['configurationId'].split(':')[-1]
 
         arn = None
         if 'Sns' in record:
--- a/zappa/util.py
+++ b/zappa/util.py
@@ -157,6 +157,8 @@ def get_event_source(event_source, lambda_arn, target_function, boto_session, dr
         arn_back = split_arn[-1]
         ctx.environment = arn_back
         funk.arn = arn_front
+        funk.name = target_function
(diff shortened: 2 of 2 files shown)


Miserlou:
I want to poke @parroyo for additional review of this one since he wrote this code. The primary issue is ensuring that it works for both S3 events and stream events.

parroyo:
Hey @Miserlou I've just checked the Pull Request, and It's fine.
S3 events are sending the 'configurationId' field with the following format:
`{lambda-arn}:{funcion-name}` 
I don't know if I introduced the bug adding the aws function mapping or it was another change. Sorry about that.

I guess we should also change the S3 event in test.py:
` u'Records': [{'s3': {'configurationId': 'test_settings.aws_s3_event'}}],`
to something like:
`u'Records': [{'s3': {'configurationId': 'arn:aws:lambda:::test_settings.aws_s3_event'}}],`

AbsolutOD:
I just wanted to mention that the reason that I didn't do the full lambda ARN is because I was concerned about field max size.  I was talking with a peer and he warned me about the size of the Event's Name field on a S3 bucket.

I couldn't find any docs on the actual limit of that field, but I think we could hit the limit if we did the whole <lambda_arn>:<module>.<function>.  So, my thought was to just keep it to the actually lambda function name and the event call function.

Miserlou:
Just tested, works great.

Issue #319, opened 2016-09-12
Properly Package Source and Non-Wheel Dependencies


dtenenba:
Please see #185 for details. The gist of it is in the subject line of this issue.

There seem to be several issues:
- When installing a dependency from GitHub, where the actual python module is in
  a subdirectory of the git repos, 
  it gets installed in a non-standard place in the virtualenv and when
  deploying, Zappa does not see it. 
- In an attempt to mitigate the previous point, if you clone the dependency and
  install it with `python setup.py install`, Zappa still does not see it when packaging
  up your virtualenv.
- What works is to create a wheel from the dependency and then install the 
  wheel into the virtualenv. The process for doing this is described 
  [here](https://github.com/Miserlou/Zappa/issues/185#issuecomment-246439291).

Steps to reproduce this are detailed [here](https://github.com/Miserlou/Zappa/issues/185#issue-165831188).

PR #495, opened 2016-11-28
include packages installed with pip install -e ... which end up in src/ virtualenv path

schuyler1d:
## Description
include packages installed with pip install -e ... which end up in src/ virtualenv path

## GitHub Issues
I think this solves at least some issues for https://github.com/Miserlou/Zappa/issues/319

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ python:
   - "2.7"
 # command to install dependencies
 install:
-  - "pip install setuptools --upgrade; pip install -r test_requirements.txt; python setup.py install" 
+  - "pip install setuptools --upgrade; pip install -r test_requirements.txt; pip install -e git+https://github.com/django/django-contrib-comments.git#egg=django-contrib-comments; python setup.py install" 
 # command to run tests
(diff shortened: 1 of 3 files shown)


Miserlou:
This looks like a great solution! Unfortunately the code so far doesn't actually work if there _isn't_ a src directory, so the build is failing. Can you please update the code so that an appropriate check is in place, and that it's tested and covered so that Travis passes without any grumbles?

schuyler1d:
so that fixes the issue.  Can you point me to an example in the code for how to test a virtualenv if you want me to add a test?  I'm not seeing anything obvious in the tests/ directory.

Miserlou:
It'd actually be related to the Travis environment, perhaps you could add a package with a -e to the `test_requirements.txt`? Although that'd only check the code for the positive path, not the negative path. Otherwise, it'll take some more tricky code.

Miserlou:
And I suppose if you want to take that route, adding a lib64 package would increase coverage as well.

Miserlou:
Looks great!

Issue #494, opened 2016-11-27
Zappa Option for Enabling Cross-Origin Resource Sharing


AusIV:
I'm building an application that I want to have IAM authenticated and accessed through a browser. I have my browser application set up to properly sign requests to IAM authenticated APIs, but there's a problem. The browser does pre-flight OPTIONS requests to the same URL as the XHR request, but there's no way to get the browser to sign the pre-flight request. So if the entire API is IAM authenticated, the pre-flight request will fail authentication and be rejected, at which point the browser will decide it shouldn't allow cross-origin requests for this URL.

This can be resolved manually by enabling CORS with some manual steps after doing the Zappa deployment (outlined [here](https://github.com/Miserlou/Zappa/issues/41)).

Those manual steps can pretty easily be taken care of in the CloudFormation template by adding a section along the lines of:

```
"OPTIONS0": {
  "Type": "AWS::ApiGateway::Method",
  "Properties": {
    "ResourceId": {
      "Fn::GetAtt": [
        "Api",
        "RootResourceId"
      ]
    },
    "RestApiId": {
      "Ref": "Api"
    },
    "AuthorizationType": "NONE",
    "HttpMethod": "OPTIONS",
    "Integration": {
      "Type": "MOCK",
      "IntegrationResponses": [
        {
          "ResponseParameters": {
            "method.response.header.Access-Control-Allow-Headers": "'Content-Type,X-Amz-Date,Authorization,X-Api-Key,X-Amz-Security-Token'",
            "method.response.header.Access-Control-Allow-Methods": "'GET,POST,OPTIONS'",
            "method.response.header.Access-Control-Allow-Origin": "'*'"
          },
          "ResponseTemplates": {
            "application/json": ""
          },
          "StatusCode": "200"
        }
      ],
      "PassthroughBehavior": "NEVER",
      "RequestTemplates": {
        "application/json": "{\"statusCode\": 200}"
      }
    },
    "MethodResponses": [
      {
        "ResponseModels": {
          "application/json": "Empty"
        },
        "ResponseParameters": {
          "method.response.header.Access-Control-Allow-Headers": true,
          "method.response.header.Access-Control-Allow-Methods": true,
          "method.response.header.Access-Control-Allow-Origin": true
        },
        "StatusCode": "200"
      }
    ]
  }
},
"OPTIONS1": {
  "Type": "AWS::ApiGateway::Method",
  "Properties": {
    "ResourceId": {
[...]

PR #496, opened 2016-11-28
Enable cors

AusIV:
## Description

As outlined here: #494 setting up a Zappa application to use CORS along with an api-gateway managed authentication mechanism requires enabling cors at the api-gateway layer to deal with the pre-flight options requests. There is a fairly simple button to manage this in the API-gateway management console, but I'd rather not have to take those steps after doing a deployment.

The "cors" option in the settings can either be "true" (defaulting to the AWS
API Gateway "Enable Cors" values) or a dictionary specifying "allowed_headers"
"allowed_methods" and "allowed_origin". Any keys not specified will default
to the API Gateway defaults.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/494

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,6 +436,7 @@ to change Zappa's behavior. Use these at your own risk!
         "timeout_seconds": 30, // Maximum lifespan for the Lambda function (default 30, max 300.)
         "touch": false, // GET the production URL upon initial deployment (default True)
         "use_precompiled_packages": false, // If possible, use C-extension packages which have been pre-compiled for AWS Lambda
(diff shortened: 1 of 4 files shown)


Miserlou:
Huge!

Requesting input from @aectann @bjinwright and @collingreen since I think they use these services.

aectann:
I haven't tested it yet. But I do like how it seemingly makes CORS configuration a breeze. Nice work! I'll try to give it a go later today.

Miserlou:
I think we're good!

PR #497, opened 2016-11-29
This will find all .egg-link files and add the correct python module to the zip package.

### LINKED ISSUE #319: Properly Package Source and Non-Wheel Dependencies

xuru:
## Description
The previous pull request: https://github.com/Miserlou/Zappa/pull/495 adds the cloned github repo to the zip package instead of the actual python module.  For example, if I have:
```
-e git+https://github.com/Miserlou/Zappa.git#egg=zappa
```
in my requirements.txt file, it will create a `~/.virtualenvs/zappa/lib/python2.7/site-packages/zappa.egg-link` with the following contents:
```
/Users/xxxxxxx/.virtualenvs/zappa/src/zappa
.
```
The directory `/Users/xxxxxxx/.virtualenvs/zappa/src/zappa` contains the full contents of the cloned zappa repo:
```
zappa
├── CHANGELOG.md
├── LICENSE
├── MANIFEST.in
├── setup.py
├── zappa
│   ├── __init__.py
```
So, when running the lambda, importing zappa will fail because it's the 'inner' zappa directory is what we actually want.

This change, finds all the 'egg-link' files, and uses setuptools.find_packages to gather the actual python packages to include and copies them to the zip file staging area.

## GitHub Issues
Should fix #319 for realz

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -2,6 +2,8 @@
 import base64
 import collections
 import json
+from contextlib import nested
+
 import mock
 import os
 import random
@@ -13,7 +15,7 @@ import unittest
 from click.exceptions import ClickException
 from lambda_packages import lambda_packages
 
-from .utils import placebo_session
+from .utils import placebo_session, patch_open
 
 from zappa.cli import ZappaCLI, shamelessly_promote
 from zappa.ext.django_zappa import get_django_wsgi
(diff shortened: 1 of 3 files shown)


Miserlou:
Crazy test. Love it. Thanks!

Issue #6, opened 2016-02-10
Only works with Python 2.7?


tiagofassoni:
First of all, awesome work, folks! Really liked it and intend to use.

This is more of a reminder for the README, because, when reading the code, I saw "python2.7" hardcoded and I figured it would be best to write that in the README

Miserlou:
The problem is that Lambda itself only supports Python 2.7 for now.

I don't think there's any reason why it _wouldn't_ work on P3, but the
target is what AWS supports.

tiagofassoni:
Oops. Totally forgot about that. Mind if I update the README to write about that “limitation"?

Miserlou:
So, I was thinking more about this, and only the stuff that actually runs AWS side needs to be Py2.7.

All the rest can be Python3. However, this means that the tests need to be aware that some functions should only be tested on the target platform. Do you know how to do that with Travis?

tiagofassoni:
Well, I have absolutely no idea, sorry.

mathom:
I believe you could run Python 3 on lambda if you really wanted but you would have to ship a statically compiled version of the python binary along with your code. Sounds pretty gross, though.

Miserlou:
That's a nightmarish idea.. not supporting that! :-P

Honestly I expect that Lambda will support multiple Python versions in the next 12 months. Screenshot this and we'll come back to it in a year to see if I was right..

Miserlou:
That being said, I think we should target 3.5 for all of the local Zappa tools (although I don't know what that will do to the venv packaging stuff. Could relate to our idea to pass custom venvs through the ZAPPA_SETTINGS.)

maljac:
Sidenote: I was not able to use zappa with Python 2.7.6 (SSL / Insecure.. warnings from urllib3). With 2.7.11 I had no problems. Maybe you should put this in the Readme as requirement.

Miserlou:
That's interesting. I think that comes because you need to install requests[security] on pythons.

hayd:
FWIW Here's some work towards py3 https://github.com/hayd/Zappa/tree/py3. IIUC it's all for nought while lambda in python 2.7 only, but may be worth testing out early anyway.

*Note: Some unicode handling became a little messier in a few places, but a lot of this could be cleaned up (for one thing by defining b64encode on str).*

xuru:
Hey guys... we can always add a `compat.py` file to allow python2.7/python3.5 compatibility functions.  That way we'll be ready when it happens.

[...]

PR #499, opened 2016-11-30
Feature/support python35

SBillion:
## Description
I did some update to work with python3. It's not finish and I won't have the time to do more before a while, so I would like to share it with you

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/6

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 python:
   - "2.7"
+  - "3.5"
 # command to install dependencies
 install:
   - "pip install setuptools --upgrade; pip install -r test_requirements.txt; pip install -e git+https://github.com/django/django-contrib-comments.git#egg=django-contrib-comments; python setup.py install" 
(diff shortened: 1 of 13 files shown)


Miserlou:
Damn, this is huge, and I have a feeling that it'll be very relevant in the coming 48h - want to take the final step and make that test pass? Also poking @xuru since I think it's your test :)

xuru:
Hey, sorry, just saw this.  Yeah, I think the problem is in python 3.5, `str` doesn't have a `decode` method.  Looking through the documentation for both 3.5 and 2.7, it looks like we can just add a `b` to the open function.  For example:
```
    with open(egg_link, 'rb') as df:
```
The "binary" mode will have no effect in python 2.7, but will return the string from the `.read` method as a `byte` array which does have a decode method.

The other option, is to do some kind of `if` statement around every file read which would be really ugly.

SBillion:
Yes, problems are around bytes and str, mostly.
I didn't have the time to find a good way to use base64 with python35
An other point of fail is patch_open:

```python
with (patch_open(), mock.patch('glob.glob'), mock.patch( 'zappa.util.copytree')
        ) as ((mock_open, mock_file), mock_glob, mock_copytree):
```

Miserlou:
Closing because we support P36 now! :D

Issue #468, opened 2016-11-13
Allow importing local files for callbacks


TronPaul:
If your project is small it doesn't really make sense to install a module to your virtualenv or system modules just for callbacks. For example:

`zappa_settings.json`
```
{
    "dev": {
        "app_function": "myapp.app",
        "callbacks": {
                "post": "myapp.post_callback"
            }
    }
}
```

The above will fail because `importlib` will not search the working directory for modules to import. Given that zappa is only compatible with Python 2.7 I think that `imp` should be used over `importlib` and that on an `ImportError` the current working directory is also searched for modules to load.

Miserlou:
That sounds like a fair idea to me, but we need to find a Python3 compatible solution because Py3 Lambda is on the horizon..

TronPaul:
If Python3 compatibility is a requirement there would need to be 2 codepaths. 1 if `imp` can be imported (2.7), 1 if not (3 with `importlib`). I can take a stab at making a PR.

Miserlou:
Sounds good!

Miserlou:
Poke

TronPaul:
I finally got around to trying to implement this, but I wasn't able to get both single python files and folder packages working at once. I think I need to do more reading on python import conventions. If someone else wants to give this a go, feel free.

tsarpaul:
Hi, I was able to implement the feature and I want to make a pull request.
Would it be OK to override the current way to install a callback since it would be a breaking change?

Miserlou:
What is your proposed change? Is there anyway way to remain backwards-compatible? Feel free to send the PR so I can take a look at what you've got.

PR #509, opened 2016-12-02
Enhacement/Allow importing local files for callbacks #468

tsarpaul:
## Description
This allows for a callback to be set in the working directory.
This is backwards-compatible, the user can still install callbacks over his virtualenv, in the Zappa folder.
Also compatible with Python 3.

## GitHub Issues
#468


Miserlou:
Looks okay, I wish we weren't no-covering those exceptions but I think it's good.

PR #520, opened 2016-12-06
Update description of `log_level` setting

mihow:
## GitHub Issues

The default for `log_level` is "DEBUG" however the description stated "INFO". I updated that and formatted the description to be more consistent with the others. Although the default is also implied in the key/value pair at the beginning of the line and could be removed all together in the comment.


Miserlou:
Ah, okay. I think Travis failed for an unrelated reason but this is good

PR #533, opened 2016-12-12
Cognito user pool authorizer

geeknam:
## Description
- `authorizer.type` in settings specifies type COGNITO_USER_POOL (defaults to TOKEN for backwards compatibility)
- Change the signature of `create_authorizer` to pass in `authorizer` settings

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -781,13 +781,13 @@ class Zappa(object):
         authorizer_resource = None
         if authorizer:
             authorizer_lambda_arn = authorizer.get('arn', lambda_arn)
-            lambda_uri = 'arn:aws:apigateway:' + self.boto_session.region_name + ':lambda:path/2015-03-31/functions/' + authorizer_lambda_arn + '/invocations'
-
-            authorizer_resource = self.create_authorizer(restapi,
-                                                         lambda_uri,
-                                                         "method.request.header." + authorizer.get('token_header', 'Authorization'),
-                                                         authorizer.get('result_ttl', 300),
-                                                         authorizer.get('validation_expression', None))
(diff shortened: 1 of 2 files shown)


Miserlou:
Woah, this is pretty huge an awesome! @aectann , I think this was your function originally, can you comment on this?

geeknam:
Looks like there weren't any unit tests on authorizers. I don't mind setting them up if that's going to bump the coverage :)

aectann:
@Miserlou I have no memory of this. Never used authorisers. I think we need @claudeatsafe in this. He seems to be author of the lines being reworked.

geeknam:
Hope this helps with the review: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-authorizer.html
(cloudformation template ref to Cognito User Pool Authorizer)

This is also working with the project I've set up.

zappa/zappa.py
```diff
@@ -809,24 +809,30 @@ def create_api_gateway_routes(self, lambda_arn, api_name=None, api_key_required=
             self.create_and_setup_cors(restapi, resource, invocations_uri, 1, cors_options)  # pragma: no cover
         return restapi
 
-    def create_authorizer(self, restapi, uri, identity_source, authorizer_result_ttl, identity_validation_expression):
+    def create_authorizer(self, restapi, uri, authorizer):
```
geeknam: `create_authorizer` is only called from `create_api_gateway_routes` so I decided to change the signature (args). Nothing else should break.

Miserlou:
Hey there, sorry for not getting to this yet! It's on my TODO list but time keeps slipping..

geeknam:
@Miserlou no big deal, I'm using my own fork for the moment.

PR #541, opened 2016-12-15
Fix manylinux wheels

maebert:
Wheels would never actually get used because they're installed into the wrong temporary directory.

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -424,24 +424,7 @@ class Zappa(object):
             installed_packages_name_set = [package.project_name.lower() for package in
                                            pip.get_installed_distributions()]
 
-            # First try to use manylinux packages from PyPi..
-            # Related: https://github.com/Miserlou/Zappa/issues/398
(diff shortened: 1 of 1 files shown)


Miserlou:
Thanks for the fix, but I have to request the removal of the preference of manylinux wheels over lambda-packages. Lambda-packages are supposed to include versions of the packages on PyPI that should be smaller and better for Lambda.

Want to write a test for this as well? :)

maebert:
That makes sense - but why do we try wheels first then? We could swap the blocks around, try lambda_packages first, and then a wheel if we don't have a package yet.

Miserlou:
That'd be fine as well. The way it is now they'll be overridden, but we don't have to worry about state as much.

maebert:
Alright, this is updated. However looking at the current test structure, you'd have to refactor A LOT to actually test this behavior - in essence, we'd need smoke tests. I'm working on a PR which does exactly that, but for now this should at least fix wheels.

PR #542, opened 2016-12-15
[WIP] Explicit requirements

maebert:
Tests are yet to come, but please have a look. This includes caching requirements in `.zappa_cache.$HASH`, where `$HASH` is md5 of a colon-separated list of requirements.

Known Todos:

* Does NOT support specifying version constraints yet
* Can’t handle silly packages like PYYaml that will install as `yaml/`
* Doesn’t have tests yet

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,6 +434,7 @@ to change Zappa's behavior. Use these at your own risk!
         "profile_name": "your-profile-name", // AWS profile credentials to use. Default 'default'.
         "project_name": "MyProject", // The name of the project as it appears on AWS. Defaults to a slugified `pwd`.
         "remote_env": "s3://my-project-config-files/filename.json", // optional file in s3 bucket containing a flat json object which will be used to set custom environment variables.
+        "requirements": [], // optional explicit specification of environments (such as ["requests", "bottle"]) - this will override packages installed in the virtual environmen
         "role_name": "MyLambdaRole", // Name of Zappa execution role. Default ZappaExecutionRole. To use a different, pre-existing policy, you must also set manage_roles to false.
         "s3_bucket": "dev-bucket", // Zappa zip bucket,
         "settings_file": "~/Projects/MyApp/settings/dev_settings.py", // Server side settings file location,
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -1408,6 +1408,7 @@ class ZappaCLI(object):
         # Create the zip file
         self.zip_path = self.zappa.create_lambda_zip(
                 self.lambda_name,
+                requirements=self.stage_config.get('requirements', []),
                 handler_file=handler_file,
                 use_precompiled_packages=self.stage_config.get('use_precompiled_packages', True),
                 exclude=self.stage_config.get('exclude', [])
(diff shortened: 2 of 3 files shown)


Miserlou:
Excited for this one, LMK when you're ready for review.

maebert:
This is ready for review, but I might need some help writing a meaningful test for this.

maebert:
@Miserlou How can I help with the review? Anything else we need for this?

smorin:
Hope this is merged soon 👍

mathom:
Hey @maebert I ran into this while testing your PR:

```
Exception in thread Thread-1:                                                                                                          | 2.10M/15.2M [00:00<00:14, 892KB/s]
  File "/usr/local/Cellar/python/2.7.11/Frameworks/Python.framework/Versions/2.7/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/tmp/en/lib/python2.7/site-packages/tqdm/_tqdm.py", line 102, in run
    for instance in self.tqdm_cls._instances:
  File "/tmp/en/bin/../lib/python2.7/_weakrefset.py", line 60, in __iter__
    for itemref in self.data:
RuntimeError: Set changed size during iteration
```

This is on a nearly empty flask test project with no reqs set in settings (using the old venv functionality). It also resulted in a non-functional project with: `[1483582177841] Unable to import module 'handler': No module named werkzeug.wrappers`

I've verified that it depoys with zappa master. Here's my code and config:

```
{
    "dev": {
        "app_function": "my_app.app",
        "s3_bucket": "zappa-*********"
    }
}
```

```
from flask import Flask
app = Flask(__name__)

@app.route("/")
def index():
    return "Hello"

if __name__ == '__main__':
    app.run()
```

smorin:
@maebert Did you get the same error?

maebert:
Hey @mathom , just a quick update: still investigating this. The `RuntimeError` oddly happens when deploying the APIGateway, not during packaging. A bit of a Heisenbug, but I'll track it down eventually.

mathom:
good deal, let me know if you need any more info 👍

Miserlou:
Bumping this

Miserlou:
Rebump

Miserlou:
Bumping this as it keeps coming up in discussions around CI/CD.

Miserlou:
Another I have with this is that this assumes a list in a JSON config, I'd like (prefer) it if this could also take a path to a `requirements.txt` file.

cjerdonek:
I thought of an alternative way of implementing this issue and issue #835 that is different from (and I believe simpler than) PR #542's approach.

What about the idea of, if "requirements_path" is provided in the settings file, then Zappa would create a virtualenv for you, and essentially run itself inside that virtualenv.  This would be simpler to implement and maintain because you'd be able to use Zappa's existing code for deploying from a virtualenv. So the implementation would just be a matter of wrapping Zappa's existing workflow with the pre- and post-steps of creating, running from, and deleting the virtualenv.

As a bonus, it would be good if the created virtualenv didn't _itself_ have to have Zappa installed (unless Zappa does need to be installed on the Lambda server -- I haven't looked into whether it needs to be).

jasonrhaas:
Any traction on this?

Issue #501, opened 2016-11-30
Support AWS-specific Environment Variables


Miserlou:
We already have "local" and (S3) "remote" envvars, so this may be unnecessary, but we could also support [AWS environment variables](https://aws.amazon.com/blogs/aws/new-for-aws-lambda-environment-variables-and-serverless-application-model/) for completeness.

atif1996:
Actually I think considering switching to the new system may be advantageous. 

Basically the same package could be run in test and then deployed to prod. Build once, and then deploy.

That said, one gets the same effect by not using the environments in Zappa, and setting them in AWS with an API call.

Miserlou:
I actually see that as a disadvantage. With Zappa-local envs, you can do something like

```
dev:
  environment_variables: 
    apikey: foo 
prod:
  environment_variables: 
    apikey: bar 
```

I don't think you can do that with AWS vars, can you?

Similarly, we also want to avoid any vendor-specific lock-in, so I think this should be prefixed as `aws_environment_variables`.

horenmar:
I might be missing something because I just started testing using Zappa, but if Zappa creates a lambda per env, then you can also have separated AWS Lambda env variables, they are not shared between lambdas.

Miserlou:
You can also do that in Zappa with the `extends` syntax.

```
base:
  environment_variables: 
    foo: bar 
dev:
  extends:  base
prod:
  extends:  base
```

atif1996:
So I like how the variables are defined in the settings file. However, I don't know the details of how they make it into dev and prod. I thought you have to use Zappa to deploy to both and that each package would be different.

In a corporate environment, it would be preferable if the artifact built by Jenkins, using Zappa, was send to dev, and the same artifact, un-modified as sent to production. You could also envision a case where you had it in prod, in one account and later move it. 

This is how people use Docker now. You create the container and the same container is deployed to dev and prod, but it fed different environment variables to target dev vs production resources.
The environment may change, but the binary shouldn't have to.

Red adding vendor lock-in, if we are using environment variables, the same can be defined in the container/instance where we could as a Flask app. In google, I think you could target google app engine as a flask app. Not sure how Microsoft's Function as a service works.

Miserlou:
Yeah, that's a fair point. You can achieve that with S3-remote environment variables now: https://github.com/Miserlou/Zappa#remote-environment-variables

[...]

PR #549, opened 2016-12-18
Add Lambda native env vars

geeknam:
## Description
Add AWS Lambda native environment variables

Related to: https://github.com/Miserlou/Zappa/issues/501

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -471,7 +471,8 @@ class ZappaCLI(object):
                                                        description=self.lambda_description,
                                                        vpc_config=self.vpc_config,
                                                        timeout=self.timeout_seconds,
-                                                       memory_size=self.memory_size)
+                                                       memory_size=self.memory_size,
+                                                       environment_variables=self.environment_variables)
 
         # Schedule events for this deployment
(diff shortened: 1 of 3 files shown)


Miserlou:
This looks good, but it doesn't seem connected the the CLI yet - should I expect that in addition this commit?

geeknam:
@Miserlou Zappa settings already includes environment_variables in `stage_config`. I'm getting the config from here: https://github.com/geeknam/Zappa/blob/6136f0548f6b42653f2495e880c7522ca5bd638f/zappa/cli.py#L1436

Not sure if that's what you mean

Miserlou:
Ah, so the problem there is that it's a conflict with the existing environment variables, which should be explicitly separated.

Please move _these_ environment variables to the vendor-prefixed `aws_environment_variables` and update the documentation accordingly.

Miserlou:
Also, the really interesting thing about AWS env vars is the use of KMS. Does this PR address that?

jonatasbaldin:
@Miserlou could we close this since its merged?

jneves:
Already merged.

PR #557, opened 2016-12-20
Fix the returned order of events in logs

pjz:
## Description
`zappa tail` can currently show old logs misordered.  This patch:

   1.  sorts the received events
   2. fetches *all* the logs; previous code wasn't dealing with the return of a nextToken

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -1818,14 +1818,19 @@ class Zappa(object):
 
         all_streams = streams['logStreams']
         all_names = [stream['logStreamName'] for stream in all_streams]
-        response = self.logs_client.filter_log_events(
-            logGroupName=log_name,
-            logStreamNames=all_names,
-            filterPattern=filter_pattern,
-            limit=limit
-        )
+        fle_args = {'logGroupName':log_name,
+                    'logStreamNames':all_names,
+                    'filterPattern':filter_pattern,
+                    'limit':limit
+                    }
+        events = []
+        response = {}
+        while not response or 'nextToken' in response:
+            if 'nextToken' in response: fle_args['nextToken'] = response['nextToken']
+            response = self.logs_client.filter_log_events(**fle_args)
+            events += response['events']
 
-        return response['events']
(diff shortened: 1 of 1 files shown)


Miserlou:
Groovy

pjz:
Hate to say it, but I think fixing the nextToken bits might have broken live updates. Can someone test?

Miserlou:
Reverting until confirmed fixed, with test.

Would also appreciate it if you could keep the style as it was rather than using kwargs, which make the code less readable, and please try to use explicit variable names rather than abbreviations (ex 'fle_args')

pjz:
I didn't use them frivolously; replacing them means duplicating the call (and args, etc) to `filter_log_events` literally with some guard code that looks like:
```python
if nextToken:
    response = self.logs_client.filter_log_events(
             logGroupName=log_name,
             logStreamNames=all_names,
             filterPattern=filter_pattern,
             limit=limit,
             nextToken=nextToken
         )
else:
    response = self.logs_client.filter_log_events(
             logGroupName=log_name,
             logStreamNames=all_names,
             filterPattern=filter_pattern,
             limit=limit
         )
```
...which to me is less readable than using kwargs, as well as being more susceptible to future error (a fix that gets put in one call but not the other).  Your call though, I guess.

Miserlou:
Is `logs_client` not smart enough to handle `nextToken=None`? Seems like the entire conditional should be unnecessary

pjz:
It apparently isn't.  Trying got me:
```
ParamValidationError: Parameter validation failed:
Invalid type for parameter nextToken, value: None, type: <type 'NoneType'>, valid types: <type 'basestring'>
```
I wonder what happens if it gets handed a wrong or empty string?  :thinking:

pjz:
...apparently not allowed. Even [core AWS docs](http://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_FilterLogEvents.html ) say "Length Constraints: Minimum length of 1."

PR #561, opened 2016-12-21
Fix the returned order of events in logs

pjz:
Fix returned order of events in logs
Also make checking for new logs much more efficient via new `startTime` arg to `Zappa.fetch_logs()`

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -669,26 +669,20 @@ class ZappaCLI(object):
         """
         Tail this function's logs.
 
+        if keep_open, do so repeatedly, printing any new logs
         """
 
         try:
-            # Tail the available logs
-            all_logs = self.zappa.fetch_logs(self.lambda_name)
-            self.print_logs(all_logs, colorize, http)
-
-            # Keep polling, and print any new logs.
-            loop = True
-            while loop:
-                all_logs_again = self.zappa.fetch_logs(self.lambda_name)
-                new_logs = []
-                for log in all_logs_again:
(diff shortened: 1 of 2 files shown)


pjz:
This is tested to work in the running-tail case as well; I think it actually worked before but was just so slow it appeared to not be working :)  Adding `startTime` to `fetch_logs()` let the search for new logs be much much more efficient.

kapilt:
seems like an improvement, but there's some oddity here, if i have a lambda function deployed for a month, i don't want by default the last month of logs, i'd prefer a more sensible default like the last hr or 5m, with cli  param to get older logs against a window. apex defaults to last 5 minutes which seems sensible.

pjz:
Well, 'beginning of time' of course really means 'beginning of what's stored in cloudwatch logs', which Zappa defaults to setting up as 'keep forever' but you can change the retention policy on to whatever you want - I set it to a week on my dev server and a month on my production server

Also: 'beginning of time' is what the current behavior is - I was more interested in 1) fixing the ordering, which was just plain broken in some cases before but speed considerations had me rewrite it to 2) add the startTime arg to the `fetch_logs` method.  The enhancement you want is a good idea, but IMHO beyond the scope of what I was trying to accomplish here.

pjz:
@Miserlou this look okay?

Miserlou:
Finally getting to this, sorry for the delay.

Looks good, but takes a long time to call `tail`, so I suggest we implement #586.

Issue #564, opened 2016-12-22
zappa invoke


bashu:
```shell
zappa invoke staging "print 1 + 2 + 3" --raw                   (feature/lambda✱)
Calling invoke for environment staging..
START RequestId: abababf4-c81d-11e6-a731-612b3622cb56 Version: $LATEST
need more than 1 value to unpack: ValueError
Traceback (most recent call last):
  File "/var/task/handler.py", line 479, in lambda_handler
    return LambdaHandler.lambda_handler(event, context)
  File "/var/task/handler.py", line 192, in lambda_handler
    return handler.handler(event, context)
  File "/var/task/handler.py", line 309, in handler
    app_function = self.import_module_and_get_function(whole_function)
  File "/var/task/handler.py", line 182, in import_module_and_get_function
    module, function = whole_function.rsplit('.', 1)
ValueError: need more than 1 value to unpack

END RequestId: abababf4-c81d-11e6-a731-612b3622cb56
REPORT RequestId: abababf4-c81d-11e6-a731-612b3622cb56	Duration: 0.46 ms	Billed Duration: 100 ms 	Memory Size: 512 MB	Max Memory Used: 54 MB
```

bashu:
💩

Miserlou:
Ugh, I assume this is the result of that PR I didn't test thoroughly enough.

Miserlou:
Yeah, that's the case. The new ArgumentParser is apparently conditional to ordering. Goddammit, gotta revert all this.

bjinwright:
This is still happening for me

vigilantzealot:
Still happening to me as well! When will the fix be pushed upstream?

dangayle:
Ditto

parroyo:
It's working fine on the latest version (0.34.0)

Miserlou:
Sweet

TammoHeeren:
This showed up for me again today. I am on 0.35.2. 

Log output:
```
need more than 1 value to unpack: ValueError
Traceback (most recent call last):
File "/var/task/handler.py", line 520, in lambda_handler
return LambdaHandler.lambda_handler(event, context)
File "/var/task/handler.py", line 233, in lambda_handler
return handler.handler(event, context)
File "/var/task/handler.py", line 392, in handler
app_function = self.import_module_and_get_function(whole_function)
File "/var/task/handler.py", line 223, in import_module_and_get_function
module, function = whole_function.rsplit('.', 1)
[...]

PR #566, opened 2016-12-23
Fix argument order for invoke and manage commands

rmoe:
## Description
The invoke and manage commands now work correctly
when flags like --raw and --all are specified after
the management command or function to invoke.

## GitHub Issues
#564


Miserlou:
Yay :)

PR #577, opened 2016-12-31
Removed imp

tsarpaul:
## Description
Very similar pull request to #513, this time regarding prebuild_script
Replaced imp with pkgutils for future compatibility with Python 3

## GitHub Issues
#6

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -1788,42 +1788,36 @@ class ZappaCLI(object):
 
         """
 
-        def execute_prebuild_script(self):
-            """
-            Parse and execute the prebuild_script from the zappa_settings.
+        (pb_mod_path, pb_func) = self.prebuild_script.rsplit('.', 1)
 
-            """
+        try:  # Prefer callback in working directory
+            if pb_mod_path.count('.') >= 1:  # Callback function is nested in a folder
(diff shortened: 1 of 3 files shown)


tsarpaul:
Not sure what to do with Travis, tasks are ok however,
> The command "coverage combine --append" exited with 1.
fails the jobs.

zappa/cli.py
```diff
@@ -1789,18 +1788,36 @@ def execute_prebuild_script(self):
...
 
-        # Execute it
-        prebuild_function()
+        if not hasattr(module_, pb_func):  # pragma: no cover
+            raise ClickException(click.style("Failed ", fg="red") + 'to ' + click.style(
+                "find prebuild script ", bold=True) + 'function: "{pb_func}" '.format(
+                pb_func=click.style(pb_func, bold=True)) + 'in module "{pb_mod_path}"'.format(
+                pb_mod_path=pb_mod_path))
 
+        getattr(module_, pb_func)()  # Call the function
```
dacut: Can we split this out, as in the original? Having the `getattr()` on a separate line from the invocation makes it easier when debugging from a stack trace to see if it was the getattr() that failed or the invocation that failed.

tsarpaul: Sure can 👍

Issue #556, opened 2016-12-20
if zip file is bigger than 50mb don't bother to upload


kapilt:
i happened to have a big virtualenv when i installed zappa, it seems to pull everything in the venv by default which resulted in a big zip, ideally before going to upload that it could just barf and say its not supported. even better would be just pulling deps from requirements.txt file or dependency.

chromakey:
Your situation sounds different than mine, but I use Zappa for a lot of Flask deployments where the static folder gets sucked into the zip and uploaded with the rest of the materials.  This frequently pushes my zip well over 50MB.  It doesn't seem to impact me at all and if Zappa puked at a 51MB zip, I'd have a lot of trouble with my existing workflows.  Maybe we can create some flags around size limits or dependencies?

Miserlou:
Hey @kapilt ! I think we met after your great talk at Serverless London..

You may be very interested in this commit that will hopefully be merged soon: https://github.com/Miserlou/Zappa/pull/548

kapilt:
yeah its actually on the basis of extracted size that it barfed not zip size. @Miserlou indeed we did, thanks, playing around with apigw approaches at the moment.

kapilt:
@Miserlou i had a look at #548, afaics that's the wrong solution, because i think the goal is actually just making zappa zipped lambdas smaller by default with the correct dependency stack.

ie. is the goal about keeping the size small ? or supporting large projects with over 250mb of runtime dependencies?

for keeping the size small which i think is the more universal win, conceptually, the solution in that PR feels like the wrong solution and just adds complexity and cold start latency, the right solution imo, would focus on keeping the zappa handler size minimal and getting a better default exclude. the fixed overhead on zappa handler should only be 1mb uncompressed (werkzeug and requestlogger) ~ +310k on the zip

afaics keeping the lambda size small needs a few things
 
 - [ ] better logical separation of runtime vs cli deps on zappa
 - [ ] use a better default exclude that takes into account the lambda env (no dist-info, or dups of lambda env modules)
 - [ ] zappa handler deps should be minimal, the callback for cli.certify seems to be the worse offender
 - [ ] always serialize config files as json in the lambda (add to lambda, to avoid carrying toml, yaml deps forward)
 - [ ] use explicit dependencies from a requirements.txt file instead of copying the whole virtualenv

[...]

PR #581, opened 2017-01-04
get rid of duplicates and extraneous libraries

kapilt:
they are either already present in the python lambda environment or extraneous there (dist-info).

For a simple bottle app, this shrinks the **compressed** zip from 9.3mb to 2.3mb.

first part from #556, lots more low hanging fruit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -398,7 +398,7 @@ to change Zappa's behavior. Use these at your own risk!
             }
         ],
         "exception_handler": "your_module.report_exception", // function that will be invoked in case Zappa sees an unhandled exception raised from your code
-        "exclude": ["*.gz", "*.rar"], // A list of regex patterns to exclude from the archive. To exclude boto3 and botocore (available in an older version on Lambda), add "boto3*" and "botocore*".
+        "exclude": ["*.gz", "*.rar"], // A list of regex patterns to exclude from the archive. By default Zappa excludes common python packages available in a lambda environment (botocore, boto3, concurrent.futures, jmespath, six.py, dateutil).
         "extends": "stage_name", // Duplicate and extend another stage's settings. For example, `dev-asia` could extend from `dev-common` with a different `s3_bucket` value.
         "http_methods": ["GET", "POST"], // HTTP Methods to route,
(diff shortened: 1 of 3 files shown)


kapilt:
travis ci setup seems to have been broken for a while... also getting errors running tests locally (slow) and error when AWS_DEFAULT_REGION is set and does not equal eu-west-1 ...

Miserlou:
Thanks for this PR, sorry for the New Years absence, gonna try to fix the Coveralls/Travis issue before I tackle the PR backlog - https://github.com/Miserlou/Zappa/issues/582

Miserlou:
Okay, builds fixed. Stupid Coverage.

It looks like your solution should actually just be added in here: https://github.com/Miserlou/Zappa/blob/master/zappa/zappa.py#L197 rather than in the CLI. Make sense?

Happy new year!

kapilt:
i was thinking about that, the issue is that if a user explicitly wants a more recent version of boto then in the lambda env, preserving it a a default config option for the user gives them the choice, but happy to move it over to be always on thing.

Miserlou:
Our general philosophy is be practical and sane by default, and for anything else require the user to be completely explicit.

Miserlou:
With the notable caveat that such abilities should be adequately documented. :D

kapilt:
right but isn't what was done practical and sane by default, with the option that gives the user the choice to override, by just making it the default value for the exclude setting?. putting it in zappa.py ZIP_EXCLUDES removes the choice since then it becomes a non overridable hard-coded list. anyways happy to do it either way, your call.

Miserlou:
[From a discussion on Slack, it appears that ZIP_EXCLUDES isn't replaced by the supplied `excludes`, but actually it's appended.]

Miserlou:
I'm going to merge this in now because you've been really patient, but I'm going to manually take out the dist-info because it turns out we actually need that for any resources which use pkg_resources, ex https://github.com/Miserlou/Zappa/pull/56/commits/6c90d2eb01f0fc4c7017022fbbbd336357569246

PR #589, opened 2017-01-10
Accept AWS_SESSION_TOKEN when executing via an IAM role.

dacut:
## Description
When executing via an IAM role (e.g. on an EC2 instance or within a Lambda function), Zappa fails to make API calls (and swallows the error: "The security token included in the request is invalid").

Digging into zappa.py, it has some special logic looking for `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` and only creates a session with these pieces, dropping the `AWS_SESSION_TOKEN` on the floor. That third piece is essential for IAM roles.

(I'm not sure why the logic is this way -- this is pretty much what Boto does by default, after all -- but I'm guessing it's like this for a reason, perhaps related to the region code.)

## GitHub Issues
The attached pull request adds `aws_session_token` into the Boto3 `Session()` creation call.


Miserlou:
Great!

Issue #61, opened 2016-04-16
Add 'package' command


Miserlou:
It could be useful to have Zappa be a more general Lambda utility - for instance, for automatically using lambda-packages and minifying, etc.

ubiquitousthey:
I started a fork to add the package and/or upload functionality.  I got stuck where I needed to figure how how to do the settings.  The settings file really needs to change so that one can define most of the environment settings on individual functions with an additional handler setting.  How do you see this working? I am thinking something like:
```
{
    "dev": { // The name of your environment
       "app" : {
         "s3_bucket": "lmbda", // The name of your S3 bucket
         ... // Various lambda attributes
         "app_function": "your_module.app" // The python path to your WSGI application function.
       },
       "my_function" : {
         "s3_bucket": "lmbda", // The name of your S3 bucket
        ... //Various Lambda attributes
         "handler": "module.function" // The handler for the lambda function
       }
    }
}
```
I think it would be pretty easy to handle both the old-style and new-style.  If there is an "app" key in the dictionary, load as new-style, if not load as old-style.

Your thoughts?

Miserlou:
Hi ubiquitousthey - thanks for bringing this up.

I think that might be over-engineering it a bit. I don't think an environment would have a both an app_function and a handler - just one or the other, with app_function taking precedent over handler. Does that make sense, or is there a scenario I'm missing?

ubiquitousthey:
I have an system I am working on that will have a web component that I would develop with zappa/flask.  It also will have a few lambda functions that get triggered by SES and a timer.  I was thinking it would have a configuration like the one above.

In this case I was thinking one could issue:
`zappa dev deploy` or
`zappa dev deploy app` or
`zappa dev deploy my_ses_handler`

This could also be accomplished with an environment per function or app as well, and that could leave the format pretty much the same and simplify the changes to the code, but it would push toward environments that would result in command like:

`zappa app-dev deploy` and
`zappa my_ses_handler-dev deploy`

which would result in some uglier function and apig names.

Maybe the the idea would be to have one settings file per app or function?

Miserlou:
I was thinking that it would be a settings entry for each function.

Although, to be honest, I think my longer term vision would be that non-APIGW lambda functions could simply be wrapped as decorators as a way to replace Celery, etc, so you could do something like

    @zappa(event_source="ses:blah")
    def mail(event):
        return True

or for scheduled events:

    @zappa(timing="0 0 0 * * *")
[...]

PR #592, opened 2017-01-11
initial build command

michi88:
## Description
POC to only build the project locally
We probably can add some more logic here. Maybe a custom package name and optionally not zip the package.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/61

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,6 +61,9 @@ docs/_build/
 # PyBuilder
 target/
 
+# pyenv
+.python-version
+
 # PyCharm
 #  Project settings from the PyCharm IDE are stored in the .idea folder
 .idea/
--- a/README.md
+++ b/README.md
@@ -177,11 +177,11 @@ You can also `rollback` the deployed code to a previous version by supplying the
 
     $ zappa rollback production -n 3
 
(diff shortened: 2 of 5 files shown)


Miserlou:
Looks good, can you change 'build' to 'package' per 61? Thanks!

michi88:
@Miserlou changed it to `package` and added tests. No idea why coverage still dropped 0.01%...

Miserlou:
Hey @michi88 - thanks for this!

Just tried it out and I don't think I disagree with the behavior here. Why would this command respect `delete_local_zip` (which defaults to True) - this means that you create a package, but immediately delete it by default, which I wasn't expecting and seems bizarre.

Can you make this so that 'package' _doesn't_ respect delete local zip, and update the documentation accordingly? Then I can merge.

Miserlou:
Also, you can just ignore the weird tiny percentage coverage changes.

michi88:
@Miserlou Yeah, I was in doubt about that. I agree you would expect the package to be kept by default. Will change it.

michi88:
@Miserlou changed accordingly

Issue #283, opened 2016-08-25
remote exception logging with flask


jayfk:
_Continuing from https://github.com/Miserlou/Zappa/pull/257_..

It looks like flask on zappa is currently unable to log to remote exception loggers/aggregators. I've tried Sentry, Opbeat and Raygun.

It has probably something to do with the following exception that is raised during exception handling.

``` python
[1472050994820] Traceback (most recent call last):
[1472050994820] File "/usr/lib64/python2.7/logging/__init__.py", line 880, in emit
[1472050994821] stream.write(fs % msg)
[1472050994821] File "/private/var/folders/82/bhs05x6n72s3bl5xdpm_nzsw0000gn/T/pip-build-9rkaZs/Werkzeug/werkzeug/local.py", line 343, in __getattr__
[1472050994821] AttributeError: 'str' object has no attribute 'write'
[1472050994821] Logged from file app.py, line 1587
```

From what it looks like, flasks `log_exception` function tries to write to _something_ in werkzeug that is a string but should be some kind of stream.

I believe the error get's swallowed somewhere by Zappa, continuing with handling the original exception.
## Workaround

Use logging via mail for now:

``` python
if not app.debug:
    import logging
    from logging.handlers import SMTPHandler
    mail_handler = SMTPHandler(
        mailhost='your.host.com',
        fromaddr='postmaster@your.host.com',
        toaddrs=['you@gmail.com'],
        subject='Error on your app',
        credentials=('username', 'password')
    )
    mail_handler.setLevel(logging.ERROR)
    app.logger.addHandler(mail_handler)
```
# Reproducing the issue
## requirements (sentry)

```
zappa # latest, installed from source
flask==0.11.1
```
## Settings

``` javascript
{
    "dev":{
        "app_function": "app.app", 
        "s3_bucket": "foo",
        "exception_handler": "app.unhandled_exceptions"
    }
}
```
## Minimal flask app (app.py)

``` python
from flask import Flask
from raven import Client

app = Flask(__name__)
SENTRY_DSN = 'https://user:pass@app.getsentry.com/id' # free account at getsentry.com
from raven.contrib.flask import Sentry
sentry = Sentry(app, dsn='YOUR_DSN_HERE')

@app.route('/', methods=['GET'])
def index():
    # this throws an exception that should be sent to sentry
    open("/non-existent.txt")
    return "Hi", 200

# We only need this for local development.
if __name__ == '__main__':
    app.run(debug=True)
```
[...]

PR #599, opened 2017-01-14
Initialize wsgi.errors to sys.stderr.

dacut:
## Description
Zappa's wsgi.errors environment variable is set to a static string; this should be a stream (sys.stderr for Lambda).

This is causing the strange 'str' object has no attribute 'write' issue.

## GitHub Issues
Fixes https://github.com/Miserlou/Zappa/issues/283

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,6 +1,7 @@
 coveralls==1.1
 coverage==4.3.1
 Django==1.10.5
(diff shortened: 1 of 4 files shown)


Miserlou:
AND TESTS? Fantastic! :D

jayfk:
BAM, thanks @dacut!

PR #600, opened 2017-01-16
Support AWS-specific environment variables

longzhi:
## Description
Support AWS-specific environment variables.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/501


Miserlou:
Related: https://github.com/Miserlou/Zappa/pull/549

Miserlou:
For both of those, I only want to accept the PR when KMS support is enabled and documented.

kvey:
This appears to be in master

Miserlou:
@kvey - the environment variables in Zappa were built before AWS supported their own environment variables, but they function the same. The only reason to swap would be to support KMS, but this PR doesn't support that.

richiverse:
Nice! Actually needed something like this lately

PR #601, opened 2017-01-16
add instructions for remote exception logging

jayfk:
## GitHub Issues

https://github.com/Miserlou/Zappa/issues/283

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,7 @@
     - [Catching Unhandled Exceptions](#catching-unhandled-exceptions)
     - [Using Custom AWS IAM Roles and Policies](#using-custom-aws-iam-roles-and-policies)
     - [Globally Available Server-less Architectures](#globally-available-server-less-architectures)
+    - [Remote Exception Logging](#remote-exception-logging)
 - [Zappa Guides](#zappa-guides)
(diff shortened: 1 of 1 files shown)


Miserlou:
Could you refactor this to combine it with the information in the overlooked section here: https://github.com/Miserlou/Zappa#catching-unhandled-exceptions

scoates:
Found this today.

- It can't possibly actually need [blink](https://pypi.python.org/pypi/blink), right? Maybe [blinker](https://pypi.python.org/pypi/blinker)?
- there's [code](https://github.com/getsentry/raven-python/blob/bb341218ccafe83d734144bcd8287e7ca1f907eb/raven/conf/remote.py#L27-L31) in Raven that detects lambda and avoids the threaded http client, so is it still necessary to set `requests+https`?
- if `requests+https` *is* needed, it looks like [specifying the transport in the DSN is now deprecated](https://github.com/getsentry/raven-python/blob/bb341218ccafe83d734144bcd8287e7ca1f907eb/raven/conf/remote.py#L87-L88)

Any ideas about the current state, @jayfk?

jayfk:
Looks like this is now working by default! :)

Issue #617, opened 2017-01-21
zappa status dev should print out an error requiring "lambda:ListVersionsByFunction"


richiverse:
## Context
```bash
zappa status dev
```

returns 
```
Error: No Lambda test-dev detected in us-east-1 - have you deployed yet?
```

 

It's a bug simply for the fact that it silenty ignores the error in zappa.py and should at the list print the error out which will include the permissions needed

## Expected Behavior

get a status for a deployed and live function
## Actual Behavior

get error message above
## Possible Fix

get_lambda_function_versions should not silently fail or atleast print something out to the console for the user to go and fix policy.
## Steps to Reproduce

1. remove policy lambda:ListVersionsByFunction
2. run zappa status dev
3.
4.

## Context

Figuring out minimum policy required for different commands
## Your Environment

* Zappa version used:
* Operating System and Python version:
* Link to your project (optional):
* Your `zappa_settings.py`:

PR #622, opened 2017-01-22
Fixes #613

### LINKED ISSUE #613: Tried creating a deploy group/policy from scratch and ran into issues
There is an issue with deploys where it can get stuck in a state where it will give the following error.

ClientError: An error occurred (ResourceConflictException) when calling the CreateFunction operation: Function already exist: test-dev

Deleting 
 

## Actual Behavior

I will be introducing a PR soon with the fix. It will require the "lambda:GetFunction" permission.

Create a profile with the following policies attached:
"iam:PassRole"
"s3:GetObject"
"lambda:CreateFunction"
"iam:GetRole"
"s3:PutObject"
"s3:CreateBucket"

now try zappa deploy dev a couple of times.

## Context

I'm trying to figure out minimum amount of policy requirements per zappa commands.
## Your Environment

* Zappa version used: latest from pip
* Operating System and Python version: ubuntu 16.04 and python 2.7.12 on pyenv
* Link to your project (optional):
* Your `zappa_settings.py`: 
I'm using cookie cutter settings and app.py

richiverse:
## Description
Fixed an issue for me where adding permissions 1 at a time through the deploy/undeploy cycles got stuck. This unstuck it for me. 

## GitHub Issues

Fixes #613 and #617

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -756,7 +756,8 @@ class Zappa(object):
             return response.get('Versions', [])
         # reference: https://github.com/Miserlou/Zappa/issues/617
         except botocore.exceptions.ClientError as e:
-            return e.response['Error']['Message']
+            print('You might need "lambda:ListVersionsByFunction" policy enabled')
+            return [] # e.response['Error']['Message']
 
     def delete_lambda_function(self, function_name):
         """
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -757,9 +757,7 @@ class Zappa(object):
             )
             return response.get('Versions', [])
         except Exception:
-            # reference: https://github.com/Miserlou/Zappa/issues/617
-            print('You might need "lambda:ListVersionsByFunction" policy enabled')
-            return [] # e.response['Error']['Message']
+            return []
 
     def delete_lambda_function(self, function_name):
         """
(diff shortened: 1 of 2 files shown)


richiverse:
@Miserlou ready for merge. all tests pass on travis

zappa/zappa.py
```diff
@@ -736,16 +736,30 @@ def rollback_lambda_function_version(self, function_name, versions_back=1, publi
...
         try:
-            response = self.lambda_client.list_versions_by_function(FunctionName=function_name)
+            response = self.lambda_client.list_versions_by_function(
+                FunctionName=function_name
+            )
             return response.get('Versions', [])
         except Exception:
-            return []
+            # reference: https://github.com/Miserlou/Zappa/issues/617
+            print('You might need "lambda:ListVersionsByFunction" policy enabled')
```
Miserlou: I don't agree with this change.

Not everybody needs this policy, it's far more likely that you're just trying to call status on a function that doesn't exist rather than one you don't have the ability to access. If we're going to state that it's for permission reasons, we should do it when it is actually for permission reasons.

richiverse:
removed the print statement for now. It's called in 3 places in the cli and probably still needs a better test to make it clear that if you don't have that policy, then that is why you are blocked.

PR #636, opened 2017-01-30
Switch middleware to avoid cookie packing, and use case mangling instead

bbangert:
## Description

By mangling the case of the 'Set-Cookie' header, API Gateway will retain them all and include them as 'set-cookie' headers. This replaces the cookie packing which had issues with strict cookie parsers.

### DIFF
--- a/zappa/middleware.py
+++ b/zappa/middleware.py
@@ -1,195 +1,42 @@
-import base58
-import json
-import time
-
-from werkzeug.http import parse_cookie, dump_cookie
 from werkzeug.wsgi import ClosingIterator
 
-REDIRECT_HTML = """<!DOCTYPE HTML>
-<html lang="en-US">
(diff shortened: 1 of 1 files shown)


Miserlou:
Extremely excited about this feature! Out with the old hacks, in with the new ones.

Any chance you can repeal and replace the tests for this? Should actually increase overall coverage. It'd be great if you could also throw in some docstrings to the functions explaining what's happening here, and refer back to that pretty code you found on SO for proper attribution as well. 😸

michi88:
I can confirm that it's working. Getting multiple 'set-cookie' headers back. Whohoo!

michi88:
BTW, used a django app to test this.

mp3il:
Just tested it with my django project, works perfectly :D

PR #639, opened 2017-01-31
log exception when get_lambda_function_versions fails

michi88:
## Description
I propose we at least log the exception here in order for devs to know why it fails.

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -842,7 +842,8 @@ class Zappa(object):
                 FunctionName=function_name
             )
             return response.get('Versions', [])
-        except Exception:
+        except Exception as e:
+            logger.exception('Could not list lambda versions: {}'.format(e.message))
             return []
 
     def delete_lambda_function(self, function_name):


Miserlou:
What happens if you call `status` on an undeployed function with this?

michi88:
@Miserlou sorry, didn't notice your comment...

Nothing will be logged to the console in that case unless boto lambda client is unable to authenticate for example.

This will 'fix' the common error we get when using docker to build / use zappa. When a mac sleeps for too long the time is out of sync in docker and boto complains with a handy exception telling you this. But only if we log the message here...

Miserlou:
No, I just checked this and it fails as I expected for `status`:

```
ERROR:Could not list lambda versions: An error occurred (ResourceNotFoundException) when calling the ListVersionsByFunction operation: Function not found: arn:aws:lambda:us-east-1:123:function:baby-flask-rev
  File "/Users/rjones/Projects/Zappa/lib/Zappa/zappa/zappa.py", line 849, in get_lambda_function_versions
    FunctionName=function_name
  File "/Users/rjones/Projects/Zappa/baby_flask/env/lib/python2.7/site-packages/botocore/client.py", line 253, in _api_call
  File "/Users/rjones/Projects/Zappa/baby_flask/env/lib/python2.7/site-packages/botocore/client.py", line 543, in _make_api_call
    raise error_class(parsed_response, operation_name)
ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the ListVersionsByFunction operation: Function not found: arn:aws:lambda:us-east-1:123:function:baby-flask-rev
Error: No Lambda baby-flask-rev detected in us-east-1 - have you deployed yet?
```

Please update this code to handle ResourceNotFoundExceptions separately, in which case there is no need to log, from other exceptions.

michi88:
Have been really busy the last 2 weeks. Will fix this in the coming days.

Issue #641, opened 2017-02-01
Still issues with the slim_hanlder key


alexlokhov:
I tested the latest Zappa 0.35. With the slim_hanlder key enabled I now get two zip file. The handler zip file contains most of the packages from the main zip file (but not all). Because of this, the handler package is still too big.

I believe the handler file was supposed to have a minimal set of packages so that it is small enough to fit on Lambda. I have also tried running Zappa to create local zip files (`zappa package dev`) and check what's inside. The hanlder zip file contains 75% of all packages, including all the heavy ones: scipy, sklearn, pandas, numpy, etc.

## Actual Behavior
```
 zappa update dev
Calling update for environment dev..
Downloading and installing dependencies..
100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 47/47 [00:06<00:00,  6.74pkg/s]
Packaging project as zip..
Downloading and installing dependencies..
100%|███████████████████████████████████████████████████████████████████████████████████████████████████| 47/47 [00:06<00:00,  7.44pkg/s]
Packaging project as zip..
Uploading credit-score-ap-dev-1485982804.zip (91.7MiB)..
100%|███████████████████████████████████████████████████████████████████████████████████████████████| 96.2M/96.2M [00:02<00:00, 33.4MB/s]
Uploading handler_credit-score-ap-dev-1485982858.zip (81.0MiB)..
100%|███████████████████████████████████████████████████████████████████████████████████████████████| 84.9M/84.9M [00:01<00:00, 55.4MB/s]
Updating Lambda function code..
Oh no! An error occurred! :(

==============

Traceback (most recent call last):
  File "/root/credit_score_api/env/local/lib/python2.7/site-packages/zappa/cli.py", line 2057, in handle
    sys.exit(cli.handle())
  File "/root/credit_score_api/env/local/lib/python2.7/site-packages/zappa/cli.py", line 399, in handle
    self.dispatch_command(self.command, environment)
  File "/root/credit_score_api/env/local/lib/python2.7/site-packages/zappa/cli.py", line 435, in dispatch_command
    self.update()
  File "/root/credit_score_api/env/local/lib/python2.7/site-packages/zappa/cli.py", line 695, in update
    self.s3_bucket_name, handler_file, self.lambda_name)
  File "/root/credit_score_api/env/local/lib/python2.7/site-packages/zappa/zappa.py", line 755, in update_lambda_function
    Publish=publish
[...]

PR #642, opened 2017-02-01
Only get the prepackaged lambdas listed as installed in site_package …

mcrowson:
…or site_packages64. The pip call gets all installed packages, but when slim_handler is used we don't need everything

## Description
The lambda_packages was getting pulled in according to pip, but we are faking a venv for the slim handler and specifying its packages in site_packages. 

## GitHub Issues

#641

### DIFF
--- a/zappa/zappa.py
+++ b/zappa/zappa.py
@@ -493,7 +493,8 @@ class Zappa(object):
         if use_precompiled_packages:
             print("Downloading and installing dependencies..")
             installed_packages_name_set = [package.project_name.lower() for package in
-                                           pip.get_installed_distributions()]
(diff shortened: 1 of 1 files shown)


Miserlou:
Ah, does this resolve the conflict in https://github.com/Miserlou/Zappa/commit/80e4dd4ab36ca24c23e8ea09c5b832ba34c8f992

mcrowson:
Ah, not sure. I was wondering why it started acting like that. What was 80e4dd4 for?

Miserlou:
https://github.com/Miserlou/Zappa/pull/637

You two work it out!

mcrowson:
Incoming commit should be a little more explicit in what I was hoping to accomplish

Issue #644, opened 2017-02-02
Package not resolved using lambda_packages because package.project_name not in the list "package_to_keep"


FrederikP:
## Context

 

Yesterday I added python-ldap to the lambda-packages. Thanks for the quick merge and release, by the way.

Now I was looking to test this and ran into a problem. The package is not resolved when trying to update the environment. I have python-ldap installed in my active virtualenv.

There is a list called "package_to_keep" in zappa.py. It's filled using the listdir command on the site-packages dirs. A few lines later in the "use_precompiled_packages" block, this list is used to filter the installed distributions:

```
installed_packages_name_set = [package.project_name.lower() for package in
    pip.get_installed_distributions() if package.project_name in package_to_keep]
```

The issue is that package_to_keep only contains ['ldapurl.pyo', '_ldap.so', 'ldap', 'python_ldap-2.4.29.dist-info', 'ldapurl.py', 'ldapurl.pyc'] (excluding other packages stuff), but the package.project_name.lower() value is: 'python-ldap'.

This means that the package is not considered as a possible package that needs to be looked for in the replacement process. 

## Expected Behavior

I'd expect zappa to include the precompiled python-ldap package when updating the deployment, because python-ldap is installed in the venv used when calling zappa. 

## Actual Behavior

The python-ldap package is not included. Leading to "No module named ldap: ImportError" in my application running on lambda.

## Possible Fix

Not sure. Maybe find an alternative or fallback, because names of directories and files in the site-packages folders are not necessarily equal to those used as the project-name pip-wise.

## Steps to Reproduce

1. Use python-ldap as dependency
2. zappa deploy
3. import ldap in code
4. -> Import error

## Your Environment

* Zappa version used: 0.35.0
* Operating System and Python version: Ubuntu 16.04, Python 2.7

FrederikP:
I added a fix for this issue in https://github.com/Miserlou/Zappa/issues/645 .

Not sure if that's the most elegant way to do it, but it worked for me, and maybe it can at least help with the actual solution.

Miserlou:
Merged and pushed in 0.35.1 - thanks!

PR #645, opened 2017-02-02
Added fallback for project_name not in package_to_keep

FrederikP:
## Description
For the issue linked below, where a package like python-ldap doesn't get resolved in the lambda-packages resolving step, due to there being no listdir entry corresponding to the pip-wise project name.

It adds another check if the package is part of the site-packages by using pips package location entry.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/644


FrederikP:
not sure why CI is failing. didn't expect this impact.

PR #657, opened 2017-02-09
binary support

wobeng:

### DIFF
--- a/tests/placebo/TestZappa.test_cli_aws/apigateway.GetRestApi_1.json
+++ b/tests/placebo/TestZappa.test_cli_aws/apigateway.GetRestApi_1.json
@@ -1,0 +1,33 @@
+{
+    "status_code": 200, 
+    "data": {
+        "description": "Created automatically by Zappa.", 
+        "ResponseMetadata": {
+            "RetryAttempts": 0, 
+            "HTTPStatusCode": 200, 
+            "RequestId": "6c275db7-ef2b-11e6-a318-5fc3640c993c", 
+            "HTTPHeaders": {
+                "date": "Fri, 10 Feb 2017 00:54:05 GMT", 
+                "x-amzn-requestid": "6c275db7-ef2b-11e6-a318-5fc3640c993c", 
+                "content-length": "145", 
+                "content-type": "application/json", 
+                "connection": "keep-alive"
+            }
+        }, 
+        "createdDate": {
+            "hour": 20, 
+            "__class__": "datetime", 
+            "month": 2, 
+            "second": 13, 
+            "microsecond": 0, 
+            "year": 2017, 
+            "day": 9, 
+            "minute": 24
+        }, 
+        "binaryMediaTypes": [
+            "*/*"
+        ], 
+        "id": "9q10y8x8j9", 
+        "name": "image-fly-test5"
+    }
+}
--- a/tests/placebo/TestZappa.test_cli_aws/apigateway.GetRestApi_2.json
+++ b/tests/placebo/TestZappa.test_cli_aws/apigateway.GetRestApi_2.json
@@ -1,0 +1,30 @@
+{
+    "status_code": 200, 
+    "data": {
+        "id": "9q10y8x8j9", 
+        "name": "image-fly-test5", 
+        "description": "Created automatically by Zappa.", 
+        "ResponseMetadata": {
+            "RetryAttempts": 0, 
+            "HTTPStatusCode": 200, 
+            "RequestId": "79948f9c-ef2b-11e6-854a-99ec105212e0", 
+            "HTTPHeaders": {
+                "date": "Fri, 10 Feb 2017 00:54:28 GMT", 
+                "x-amzn-requestid": "79948f9c-ef2b-11e6-854a-99ec105212e0", 
+                "content-length": "118", 
+                "content-type": "application/json", 
+                "connection": "keep-alive"
+            }
+        }, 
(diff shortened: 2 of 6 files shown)


Miserlou:
Very exciting, can you remove the print statements please?

wobeng:
Done

wobeng:
just need to add

"binary_support" : true

and make sure you  base64.b64encode data you want to encode

Miserlou:
One of the tests is failing, can you look into it? https://travis-ci.org/Miserlou/Zappa/jobs/200061554

mcrowson:
It looks like the Response object we get back might have a content_encoding setting we could use to dynamically determine if we need to set this. Not sure how it all works, but if it is possible to get this from the app we could avoid the setting and everything outside of the handler.py

Response from App:
https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L447

Response has ResponseDescriptors Mixin with content_encoding
http://werkzeug.pocoo.org/docs/0.11/wrappers/#werkzeug.wrappers.CommonResponseDescriptorsMixin

Miserlou:
I like that kind of thinking to make it more transparent - could we use MIMEtype at the middleware layer?

Miserlou:
Boom baby!

mcrowson:
2 Things:
1. Document this in the README
2. Is this for applications that deliver JUST binary data? If i turn this setting on for a simple flask hello-world app would it break things?

Miserlou:
That's what I was hoping to discover to day. Ideally this support would be on a per-request or per-route basis, but having all-or-nothing is better than nothing-or-nothing for now.

wobeng:
works with Content-type. So simple flask hello-world app should not break

mcrowson:
What are the additional paths added/removed for in the APIGW?

wobeng:
So basically how it works is you have to tell api gateway what to treat as binary files. 

I do that with the add and remove function if binary_support is true or false

The possible options can be something like this: image/* or image/jpeg or */* .  I use */* because different browsers sends different Accepts values

So when you request with Accept image/jpeg and the response content type is image/jpeg, it will serve it as binary.

Miserlou:
We should expand that criteria (or make it user-definable), there are many other non-image binary types we may want to serve, ex `application/pdf`, `audio/wav`, etc.

wobeng:
Those are just examples..I think other binary media should work well

wobeng:
making user-definable will be an issue because remember that apigw determines what to treat as binary depending on accept http value. Different browsers sends different accepts so to make sure it works for all you need the gree

wobeng:
oops, I closed the pull request by mistake

Miserlou:
I get a `{"message": "Internal server error"}` error when using this for normal HTTP endpoints.

wobeng:
the latest push should fix that issue

wobeng:
Any update on this?

Miserlou:
This still doesn't appear to work for either plaintext or binary endpoints.

```
[1487011732947] An error occurred during JSON serialization of response: 'utf8' codec can't decode byte 0xff in position 0: invalid start byte
  File "/usr/lib64/python2.7/json/__init__.py", line 251, in dumps
  sort_keys=sort_keys, **kw).encode(obj)
  File "/usr/lib64/python2.7/json/encoder.py", line 207, in encode
  chunks = self.iterencode(o, _one_shot=True)
  File "/usr/lib64/python2.7/json/encoder.py", line 270, in iterencode
  return _iterencode(o, 0)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xff in position 0: invalid start byte
```

For:

```python
@app.route('/logo.jpg')
def logo():
    """Serves the logo image."""

    with open("logo.jpg", 'rb') as bites:
        return send_file(io.BytesIO(bites.read()),
                         attachment_filename='logo.jpeg',
                         mimetype='image/jpg')
```

wobeng:
remember that if you want to use binary, you have to base 64 encode it. Take a look at https://github.com/wobeng/zappa_resize_image_on_fly/blob/master/app.py#L37

Miserlou:
Surely this should be done at the middleware layer, not the application layer?

This is also doesn't explain why non binary endpoints don't work.

Miserlou:
We had a discussion on Slack about the MIMETYPE checking. Any progress? :D

wobeng:
by tomm you will get a push..promise..:)

Miserlou:
:man_dancing: :man_dancing: :man_dancing:

wobeng:
Updated to auto detect and base64encode it.

Examples:

https://github.com/wobeng/zappa_resize_image_on_fly/blob/master/app.py#L15 --> https://jyhmaij1l4.execute-api.us-east-1.amazonaws.com/dev/world

https://github.com/wobeng/zappa_resize_image_on_fly/blob/master/app.py#L23 -->
https://jyhmaij1l4.execute-api.us-east-1.amazonaws.com/dev/example

https://github.com/wobeng/zappa_resize_image_on_fly/blob/master/app.py#L56 -->
https://jyhmaij1l4.execute-api.us-east-1.amazonaws.com/dev/image/example/200x200.jpg

Miserlou:
Works great!

wobeng:
:man_dancing: :man_dancing: :man_dancing::man_dancing: :man_dancing: :man_dancing::man_dancing: :man_dancing: :man_dancing:

Miserlou:
Thank you for your contribution to Zappa!

PR #662, opened 2017-02-11
Fix unschedule events with trimmed names

### LINKED ISSUE #650: Too long scheduled event names result in errors updating and/or duplicate events
## Context
There seems to be a 64 char limit to schedule event names. If the combination of the project name plus stage name plus event function exceed 64 chars, the logic that creates the name differs from the logic that unschedules the event.

This results in an error updating - the event will fail to create with a LimitExceededException: An error occurred (LimitExceededException) when calling the PutTargets operation: The requested resource exceeds the maximum number allowed.

## Expected Behavior
Events should be removed and recreated when updating.

## Actual Behavior
The event isn't deleted and the update fails

## Possible Fix
Looks like the issue is related to zappa.py:

```
    def get_event_name(lambda_name, name):
        """
        Returns an AWS-valid Lambda event name.

        """
        return '{prefix:.{width}}-{postfix}'.format(prefix=lambda_name, width=max(0, 63 - len(name)), postfix=name)[:64]
```

If the name is too long, the prefix gets truncated.

`zappa.unschedule_events` - lists events using the lambda_name prefix (the project name plus '-' plus stage name. `self.events_client.list_rules(NamePrefix=lambda_name)` - however the name doesn't always have that prefix due the way they are created.

## Steps to Reproduce
1. Create a project - `'bugproject'`
2. Configure a scheduled event with name -` 'modulename.realllllllyyyyyylongmethodnameeeeee'`
3. Deploy to a stage - `'stagename'`
4. Note that `zappa.get_event_name` will generate `'bugproject-stagen-modulename.realllllllyyyyyylongmethodnameeeeee'` as the name
5. Update the stage. Observe error because `zappa.unschedule_events` assumes the prefix will be `bugproject.stagename`, but it is actually `bugproject-stagen`.

## Context
Resulted in inability to update.

parroyo:
## Description
Use boto3 events method **list_rule_names_by_target** using the lambda Arn instead of **list_rules** filtering by name prefix. This will find all the scheduled events, even the ones that have a trimmed name. 

Updated the placebo files in order to match the new calls

## GitHub Issues
Fix #650


Miserlou:
Tested seems to work nicely, thanks!

Issue #521, opened 2016-12-06
Zappa shouldn't sluggify non-default project names


horenmar:
Configuration option `project_name` defaults to sluggified pwd (which I hope means the current dir, not the entire path, but thats beside the point), but it shouldn't sluggify custom project names.

Alternatively, it should be at least noted in documentation.

Miserlou:
What are you trying to use?

It's just trying to force a valid name but there are probably better ways of doing that than what's available now.

horenmar:
PascalCase.

`ZappaTestArch` in `dev` config is deployed as `zappatestarch-dev` and I would like it to be deployed as `ZappaTestArch-dev`

Miserlou:
Ahhh. You're probably right, it should just let the user shoot themselves in the foot if that's what they want.

horenmar:
Since there is a PR open that would fix this, I'd like to bump the issue.

vlcinsky:
+1. It would make our lambda functions in AWS console a bit better readable, currently zappa managed ones break our naming conventions.

I checked the pull request code and it seems fine to me.

The new version would in fact allow to shoot in direction, the user aimed really (now it hits sometime a bit lower :-)).

If we care about preventing dangerous names, we could extend the test in place, which actually tests only for the length exceeding 16 chars, it could get extra check for other cases. 

AWS Lambda [spec states](http://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#API_CreateFunction_RequestSyntax)

Type: String

Length Constraints: Minimum length of 1. Maximum length of 140.

Pattern: ```(arn:aws:lambda:)?([a-z]{2}-[a-z]+-\d{1}:)?(\d{12}:)?(function:)?([a-zA-Z0-9-_]+)(:(\$LATEST|[a-zA-Z0-9-_]+))?```

I could make PR for that if it seems reasonable solution.

Miserlou:
+1 to matching the real validation, PR welcome.

vlcinsky:
My proposal for validating lambda function name:

The pattern breaks down to following parts::


    # optional prefix 'arn:aws:lambda:'
    (arn:aws:lambda:)?
    # optional region name, e.g. 'eu-west-1'
    ([a-z]{2}-[a-z]+-\d{1}:)?
    # optional account id, e.g.: '123456789012'
    (\d{12}:)?
    # optional 'function:' string
    (function:)?
    # obligatory: actual name of the Lambda function
[...]

PR #664, opened 2017-02-11
Validate custom project names based on actual API rules

horenmar:
## Description
I implemented validation of project names based on rules from AWS lambda spec, according to @vlcinsky analysis.

## GitHub Issues
#521

### DIFF
--- a/tests/name_scenarios.json
+++ b/tests/name_scenarios.json
@@ -1,0 +1,66 @@
+[
+  {
+    "id": "integer",
+    "value": 12,
+    "is_valid": false
+  },
+  {
+    "id": "float",
+    "value": 3.1415,
+    "is_valid": false
+  },
+  {
+    "id": "list",
+    "value": [
+      "a",
+      1,
+      null
+    ],
+    "is_valid": false
+  },
+  {
+    "id": null,
+    "value": null,
+    "is_valid": false
+  },
+  {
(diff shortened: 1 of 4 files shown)


Miserlou:
This is a great PR!

Before I merge, can I get some clarification on the '_' character. The previous behavior disallowed it, but this change allows it. Is it valid or not. I seem to remember there was some reason we had to remove '_'.

vlcinsky:
The [AWS spec](http://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#API_CreateFunction_RequestSyntax) clearly allows use of "_" (underscore).

Node.js variant of zappa (serverless) did not allow underscore what was [considered a bug](https://github.com/serverless/serverless/issues/2433) and now they allow it.

I can imagine some restriction on using underscore when the name would be processed by slugify, but as we validate the name after it was possibly slugified, such restriction would not apply to us.

In short: I think underscore shall be allowed.

horenmar:
The regex in `CreateFunction` documentation for function name is `[a-zA-Z0-9-_]+`, which does imply that underscore is allowed.

Miserlou:
Okay, this seems to work, so I'm going to merge it now.

I am worried this will cause some unknown effects that I don't remember, and it will be a bit tricky for some users during upgrade, but it's the right thing to do.

PR #676, opened 2017-02-19
Binary support when response body is not set.

cameronmaske:
## Description
Came across this enabling binary support for an oauth login process. I believe at some point in that request chain, a blank response to returned and redirected. 
Here, due to https://github.com/cameronmaske/Zappa/blob/9966573da5ee41b8359c52f4e9907fe58adf41c5/zappa/handler.py#L446 it's possible that the `zappa_returndict` `body` key is not set. 
In such a case, when attempting to be b64 encoded, an error will be raised. 
I added a simple try/except to pass over this and continue the request.


Miserlou:
Fixed this in a different way, thanks for reporting though.

Issue #603, opened 2017-01-17
Make it easier to port Celery projects / queue background tasks


michi88:
## Context
I have quite some Celery / django-q projects I would like to move to Zappa. 
Now one can argue that you don't need any of the tasks libs like celery / django-q as lambda's can easily be invoked on a scheduled interval or based on S3 / SNS events. True, but this doesn't mean the api's (of celery for example) to create tasks aren't handy. Think; `@task()` decorator and `some_func.delay()` to queue the task.

I mostly use celery to invoke some background task to be handled later (sending email, updating search indexes etc.). What I would like to see is that I can still use the Celery task creation api in order for my app to be still portable if I want to move away from lambda (something Zappa promises).

## Possible Fix
As celery already can use SQS as its queue we maybe can add a lambda handler to SQS events and execute the task.
I've been digging in the Celery codebase to search for the point where the task data (coming from the queue) is transformed and in the end executed by the worker. Because that is what we would need to do in a lambda handler.
Not easy to find, it's here:
https://github.com/celery/celery/blob/master/celery/worker/components.py#L229
Definitely not straightforward to get that working.

Another option would be to have a lambda 'celery worker' to run on a schedule every second (or whatever you want), have it connect to the SQS queue, pull a task (if available), execute it and exit. Far from ideal but workers can be told to only execute 1 task and exit (I'm told, haven't confirmed this yet). If that works we could also get around the 1 second schedule by having the lambda celery worker start on an SQS event but have if pull the task in a celery native fashion, instead of using the actual SQS message data when invoked.

Having said that I don't think we should try to make Zappa 'compatible' with Celery / django-q and whatever other framework there is. Celery though is a contender in my view as it's widely used.

The last option (except for not helping dev's on this topic at all) I see is creating helpers in zappa natively to replace the celery / django-q framework api's. Think;
```python
task_id = zappa.queue.delay('module.my_func', *args, **kwargs)
# that would upload the tasks definition to for example S3
# a zappa framework native handler could execute the task based on the S3 event
# when finished the task result is uploaded back to S3 again.
task_result = zappa.queue.get_result(task_id, remove=True)  
# if finished, returns the task result and deletes the S3 task entry (if you want to), else None
```

If that existed I wouldn't have brought this up and I would happily write logic to use zappa vs celery based on the platform I'm running on.

[...]

PR #694, opened 2017-02-27
[WIP DO NOT MERGE YET] Add SNS as event source for async tasks.

geeknam:
## Description
Just proposing an approach to having async tasks. The implementation is not fully completed.

## TODO
[ ] Undeploy SNS
[ ] Write unit tests
[ ] Refactor create_async_sns_topic, feel kinda not clean to keep adding code to cli.py

## Assumptions

SNS Topic ARN is constructed based on:
- AWS_REGION (lambda reserved env var)
- AWS_LAMBDA_FUNCTION_NAME (lambda reserved env var)
- Derived ACCOUNT_ID (queried from currently running Lambda)
- Topic always ends with `zappa-async`

### How to use

1. in zappa_settings add: `"async_enabled": true`
2. Decorate your func with `@task`

```python
from zappa.async import task
@task
def intensive_func(*args, **kwargs):
     this_takes_a_long_time()
```

3. Invoke the func: `intensive_func.delay(*args, **kwargs)`

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/603

### DIFF
--- a/zappa/async.py
+++ b/zappa/async.py
@@ -1,0 +1,88 @@
+import os
+import json
+import importlib
+import inspect
+import boto3
+
+AWS_REGION = os.environ.get('AWS_REGION')
+AWS_LAMBDA_FUNCTION_NAME = os.environ.get('AWS_LAMBDA_FUNCTION_NAME')
+
+sts = boto3.client('sts')
+AWS_ACCOUNT_ID = sts.get_caller_identity()['Account']
+ASYNC_SNS_ARN = 'arn:aws:sns:{region}:{account}:{lambda_name}-zappa-async'.format(
+    region=AWS_REGION, account=AWS_ACCOUNT_ID,
+    lambda_name=AWS_LAMBDA_FUNCTION_NAME
+)
+
+
+def import_and_get_task(task_path):
+    """
+    Given a modular path to a function, import that module
+    and return the function.
+    """
+    module, function = task_path.rsplit('.', 1)
+    app_module = importlib.import_module(module)
+    app_function = getattr(app_module, function)
+    return app_function
+
+
+def route_task(event, context):
+    """
+    Gets SNS Message, deserialises the message,
+    imports the function, calls the function with args
+    """
+    record = event['Records'][0]
+    message = json.loads(
+        record['Sns']['Message']
+    )
+    func = import_and_get_task(message['task_path'])
+    return func(
+        *message['args'], **message['kwargs']
+    )
+
+
+def send_async_task(task_path, *args, **kwargs):
+    """
+    Send a SNS message to a specified SNS topic
+    Serialise the func path and arguments
+    """
+    client = boto3.client('sns')
+    message = {
+        'task_path': task_path,
+        'args': args,
+        'kwargs': kwargs
+    }
+    return client.publish(
+        TargetArn=ASYNC_SNS_ARN, Message=json.dumps(message),
+    )
+
+
+def task():
+    """
+    Async task decorator for a function.
+    Serialises and dispatches the task to SNS.
+    Lambda subscribes to SNS topic and gets this message
+    Lambda routes the message to the same function
+    Example:
+        @task()
+        def my_async_func(*args, **kwargs):
+            dosomething()
+        my_async_func.delay(*args, **kwargs)
+    """
+    def _delay(func):
+        def _delay_inner(*args, **kwargs):
+            module_path = inspect.getmodule(func).__name__
+            task_path = '{module_path}.{func_name}'.format(
+                module_path=module_path,
+                func_name=func.__name__
+            )
+            if ASYNC_SNS_ARN:
+                return send_async_task(task_path, *args, **kwargs)
+            return func(*args, **kwargs)
+        return _delay_inner
+
+    def _wrap(func):
+        func.delay = _delay(func)
+        return func
+
+    return _wrap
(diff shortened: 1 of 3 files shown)


Miserlou:
Finally getting around to this - looks really great!

Per this implementation, I am right in think that there is only a single "topic"? I was imagining that there would be a separate topic created for each function. Or am I misreading this? Perhaps task could take an optional name argument.

geeknam:
@Miserlou 
> I am right in think that there is only a single "topic"?
Correct, only 1 SNS topic gets created for all async tasks. SNS hits the Lambda router func `route_task` which routes to the corresponding task - unlike traditional celery in which you can indicate the queue (normally you have low / high priority queues)

stevezau:
Ah, this is what i was looking for. Great.

Question, if we wanted to record the event state.. IE: Trigger async event and get a status of the event i'd need to manually generate the unique job numbers and store it in a db or something where right??

aehlke:
I still feel like this might be misguided since Lambda already has `InvocationType=Event` for async execution; there's some unresolved discussion at the end of https://github.com/Miserlou/Zappa/issues/603 around whether `InvocationType=Event` has shortcomings that are best overcome by working around it via SNS, or through other means more in line with AWS's intended usage

Miserlou:
From discussion in Slack related to direct-invocation branch off of this branch.

I think that having multiple invocation sources would be fine, the configuration would look something like:

```
dev:
    async_source: sns # Can be one of ['sns', 'sqs', 'lambda']. Defaults to 'lambda'
    async_resources: true # Create async queue resources. Default true
    async_arn: arn:blah:blah:blah # if not async_resources, specify your own queue/sns resource
```
For something like

```
@task()
def async_me(a, b, c):
    long_task(a, b, c)

@app.route('/short')
def short():
    async_me(A, B, C)
    return "Doing it!"
```

With the option of `zappa.async.run(func, args, kwargs)` for non-decorated functions, and immediate (blocking) execution for non-Lambda deployments.

We should also get rid of `delay` entirely, we can simply execute the decorated function or use `run`. Delay is a misnomer unless we can devise a way to schedule a single execution at a desired time. Celery compatibility is not a design goal here.

aehlke:
Thanks for pasting. I guess I'll join Slack for more discussion. Personally, without yet seeing what benefits SNS provides that we lack with Lambda's built-in async invocation, it would seem bad to just provide both if one doesn't offer compelling value over the other - it increases the surface area and complexity of Zappa.

Re: `delay`, it's sometimes useful in Celery to be able to call an async function synchronously, in which case you'd just call it plainly without `delay`/`apply_async`. That's one use-case of `delay`. The other benefit is that it makes it explicit that the function call will trigger async behavior, without having to name all your functions as such. Agreed it makes sense to rethink the API and naming though.

Miserlou:
Agreed, it's just the name I object to.

michi88:
Awesome! Looks good :)

First off, unfortunately I haven't been able to contribute as much as I would've like lately. Sorry about that.

Some thoughts:

I agree it's useful to be able to call the function normally. `async` or `apply_async` probably makes more sense. Also, if we add the actual delay option later we could add that as an argument to `async(delay='something')`, which actually makes more sense than having the func called `delay` I think. Maybe we can use [SQS Delay Queues](http://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-delay-queues.html) but I would say that's V2.

One thing to think about are the arguments we might be adding to `async` as when we are going to use them for things like delay, and maybe more (think for example `async_source ` or  a different `lambda_name` when you want your task to run with different memory requirements), we should also provide two optional keyword arguments `args` and `kwargs` for cases where the function itself has a `delay` argument for its own functionality already. So `zappa.async.run(func, args=(), kwargs={}, delay=100)` or `func.async(args=(), kwargs={}, delay=100)`. When the `args` and `kwargs` keywords are not used we just pop the zappa async invocation specific arguments from the regular keyword arguments. Maybe this is over-engineering at this point though... just venting some thoughts here :)

As suggested a really nice and useful feature would be if we could return a `task_id` that can be used to poll for the status of the task and fetch the result returned by the func (if any). This would require us to create some task result backend store somewhere. S3 would be an obvious contender here for simplicity and cost. Definitely no critical feature though at this point as devs can easily create this functionality in their own DB's. It's just something I would be using when available.

All in all it's great that we have this going! Kudos @geeknam and @schuyler1d for https://github.com/Miserlou/Zappa/pull/732!!

aehlke:
+1 to deferring the work on deferred execution. That's one of Celery's weakest areas, a source of a lot of complexity and it harms visibility into its operation. Would be good to take that slowly and make sure we do it right rather than try to fit it into the v1 and have regular async suffer for it.

schuyler1d:
Just to update this parallel PR. https://github.com/Miserlou/Zappa/pull/732
Now does the following:
* json serilalization of args/kwargs along with raising an exception if the payload size will be too big
* switches to `func.async()` instead of "delay"
* implements zappa.async.run(func, args, kwargs)

Miserlou:
So we've got two branches here, both of which are starting to look pretty okay. I'd like to get this merged sooner rather than later so it doesn't start to diverge from master - what's the game plan here? Are either of these ready to be merged? Both? One rather than the other?

stevezau:
Any idea when this can be merged? I have a real use case for it for a new app i'm building.

Miserlou:
@stevezau - this was merged into master today! It'll likely land in pip sometime in the next few days! :-D

stevezau:
@Miserlou timing!

Great, i'll start to test it :) 

My goal it to have some tracking id but for now i'll just use redis and manually do it within my long running functions..

Miserlou:
This code is all brand-spanking-new, so please feel free to send open issues and send PRs if it's lacking in any way and you can make improvements!

stevezau:
Will do, probably won't get to it till another day or two.

PR #711, opened 2017-03-10
slim_handler=true change working directory

### LINKED ISSUE #702: slim_handler breaks file paths & read permission
<!--- Provide a general summary of the issue in the Title above -->
## Context
I'm attempting to read a small data file from the root project directory at runtime. It works locally and on lambda when slim_handler is false but fails on lambda when slim_handler is true.

After identifying that the paths are impacted during runtime, I updated the path to match the runtime environment (for debug purposes only). Unfortunately, it looks like read permission to the unzipped project is not granted because it throws an Errno 13, IO Error when attempting to read the file.

<!--- Provide a more detailed introduction to the issue itself, and why you consider it to be a bug -->
<!--- Also, please make sure that you are running Zappa _from a virtual environment_ and are using Python 2.7 --> 

## Expected Behavior
File paths remain relative to development project & read permission is remains true.
<!--- Tell us what should happen -->

## Actual Behavior
File paths become relative to the handler. Read access not granted to projects unzipped by slim_handler
<!--- Tell us what happens instead -->

## Possible Fix
Not sure if this is fixable without a major change to the way slim_handler works. An update to the documentation explaining the issue and a method to solve it would be awesome.
<!--- Not obligatory, but suggest a fix or reason for the bug -->

## Steps to Reproduce
<!--- Provide a link to a live example, or an unambiguous set of steps to -->
<!--- reproduce this bug include code to reproduce, if relevant -->
1. Add a file to the root directory with some arbitrary text in it e.g. 
```shell
echo "text from file" > mytext.txt
```

2. Update the documentation's hello world example by appending the text from mytext.txt
```python
# my_app.py
from flask import Flask                                                                                                                                                                                     
app = Flask(__name__)                                                                                                                                                                                       
                                                                                                                                                                                                             
@app.route('/')                                                                                                                                                                                             
def hello():                                                                                                                                                                                                
    with open('mytext.txt', 'r') as w:                                                                                                                                                                      
        txt = w.read()                                                                                                                                                                                      
    return 'Hello, Zappa' + txt                                                                                                                                                                             
                                                                                                                                                                                                            
if __name__ == '__main__':                                                                                                                                                                                  
    app.run()      
```
## Your Environment
<!--- Include as many relevant details about the environment you experienced the bug in -->
* Zappa version used: 0.38.0
* Operating System and Python version: Ubuntu 16.04, Python 2.7.12
* Your `zappa_settings.py`: 
```javascript
{
    "dev": {
        "app_function": "my_app.app", 
        "s3_bucket": "zappa-test-dev-uid-0000121",
        "slim_handler": true,
        "debug" : true
    }
}
```

Thanks!

olirice:
When using slim_handler=true, set working directory to the project's root folder to preserve relative file paths.

## Description
slim_handler=true sets file paths relative to the handler. The 1-line update sets the working directory to the project root.

## GitHub Issues
Resolves #702

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -168,6 +168,9 @@ class LambdaHandler(object):
 
         # Add to project path
         sys.path.insert(0, project_folder)
+        
+        # Change working directory to project folder
+        os.chdir(project_folder)
         return True
 
     def load_remote_settings(self, remote_bucket, remote_file):


Miserlou:
Grand

PR #732, opened 2017-03-21
Async for lambda and SNS -- tackling #603

schuyler1d:
## Description
(based on @geeknam 's work) For SNS, you need to set `"async_resource": "sns"` (If you don't want it to auto-create then do `"async_resources": false`
The rest is documented at the top of zappa/async.py

This rebases @geeknam 's work, and then adds a commit.

I added a bit of abstraction for the async stuff, so that the caller can find out what happened, and if we want to add, e.g. fallback options into the `task()` call (e.g. fallback to synchronous, etc), then it's ready for such things.

This code is untested, so that's what I'm going to do next, but feedback (or testing) welcome.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/603

### DIFF
--- a/zappa/async.py
+++ b/zappa/async.py
@@ -1,0 +1,88 @@
+import os
+import json
+import importlib
+import inspect
+import boto3
+
+AWS_REGION = os.environ.get('AWS_REGION')
+AWS_LAMBDA_FUNCTION_NAME = os.environ.get('AWS_LAMBDA_FUNCTION_NAME')
+
+sts = boto3.client('sts')
+AWS_ACCOUNT_ID = sts.get_caller_identity()['Account']
+ASYNC_SNS_ARN = 'arn:aws:sns:{region}:{account}:{lambda_name}-zappa-async'.format(
+    region=AWS_REGION, account=AWS_ACCOUNT_ID,
+    lambda_name=AWS_LAMBDA_FUNCTION_NAME
+)
+
+
+def import_and_get_task(task_path):
+    """
+    Given a modular path to a function, import that module
+    and return the function.
+    """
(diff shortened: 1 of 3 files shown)


mcrowson:
I like this too. What is the point of getting the region and function out of the environment though?

schuyler1d:
@mcrowson How would you recommend that I get it?  My use case is calling code in my own app (and thus the same lambda function as is running).

mcrowson:
The LambdaAsyncResponse gets instantiated and used within lambda, but we want to use the SNS topic that was created locally during the deploy/update. So rather than trying to recreate that ARN, just save the SNS ARN to the zappa_settings.py file during the deploy process. Then once you're in lambda you can get it without knowing the region. 

Any other approach I can think of either requires setting the region manually like you're doing, or guessing it based on the lambda but that feels wrong.

Miserlou:
What's the sitrep on these PRs? Would really love to get this stuff merged in as soon as possible so we don't loose momentum. (They don't need to be perfect and I'm happy to keep working on them myself, so "works for me" could be a good enough state.) @schuyler1d @geeknam

geeknam:
I'd merge @schuyler1d PR, lgtm

schuyler1d:
@Miserlou  I haven't had a chance to actually test this yet, even in practice.  However I think it's pretty close.  I have a project late this week that will probably give me an opportunity to do some actual testing, though that will be for the Lambda part.  Still very interested in some others testing their use-cases, especially with SNS.

@mcrowson ah, that's very helpful -- I see in the handler now how zappa_settings.py will have the details I need, and I can import them directly.

PR #734, opened 2017-03-22
More robust guide for newbies getting on Zappa

edgarroman:
I had originally put together a collection of examples, but narfman0 already had that here:https://github.com/narfman0/zappa-examples/

So I am trying to put together a series of walkthroughs for common use cases for folks new to zappa.  Including:

- Basic setup
- Core Django installation
- Hosting static files
- Connecting to a database

With some background information and links to AWS concepts like IAM, VPC Networking, RDS creation, etc

Eventually want to add 
- Using cloudfront to manage urls and domains
- Using SSL
- Async tasks (knowing that it's in flux now)
- Uploading files
- User login

Guide is hosted on github here: https://edgarroman.github.io/zappa-django-guide/


Miserlou:
Really excellent stuff!

PR #735, opened 2017-03-22
Fix API Gateway test button (the bolt button)

alanjds:
## Description

The API Gateway have a test button, but it always fails because it sends an event with ``headers:null``. This PR fixes by falling back the ``null`` with an empty dict on python side.

A test was baked from the logs, replacing sensive information of my account.

Screenshot of the test button:
screen shot 2017-03-22 at 17 12 44

Partial log of the error:
```
Wed Mar 22 19:27:35 UTC 2017 : Method response body after transformations: {
    "message": "An uncaught exception happened while servicing this request. You can investigate this with the `zappa tail` command.", 
    "traceback": [
        "Traceback (most recent call last):\n", 
        "  File \"/var/task/handler.py\", line 441, in handler\n    binary_support=settings.BINARY_SUPPORT\n", 
        "  File \"/Users/alanjds/.virtualenvs/zappa/lib/python2.7/site-packages/zappa/wsgi.py\", line 49, in create_wsgi_request\n", 
        "AttributeError: 'NoneType' object has no attribute 'keys'\n"
    ]
}
```

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -544,6 +544,51 @@ class TestZappa(unittest.TestCase):
         response_tuple = collections.namedtuple('Response', ['status_code', 'content'])
         response = response_tuple(200, 'hello')
 
+
+    def test_wsgi_from_apigateway_testbutton(self):
+        """
+        API Gateway resources have a "test bolt" button on methods.
+        This button sends some empty dicts as 'null' instead of '{}'.
+        """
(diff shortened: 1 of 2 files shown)


Miserlou:
Cool - I don't think I've ever actually hit that button, but this looks like a good commit. Thanks!

Issue #738, opened 2017-03-23
Add an entry in zappa_settings to configure the Dead Letter Queue to which triple failed async invokes are sent to


chuckoy:
## Context

 
There's a feature in AWS Lambda where if an async event fails to execute properly after three tries you can opt to send it to a Dead Letter Queue. [AWS docs](http://docs.aws.amazon.com/lambda/latest/dg/retries-on-errors.html)
It would be nice to have an entry in `zappa_settings.json` where if we specified the ARN of the SQS queue/SNS topic Zappa would hook the lambda function to that DLQ. My specific use case is for replacing celery with Zappa.

## Possible Fix

Something along the lines of: `"dead_letter_queue": "sqs_or_sns_arn"` seems to be the most straightforward implementation. I'm happy to start a pull request for this once I've managed to grok exactly how `zappa_settings.json` works. It seems fairly simple and beginner-friendly.

chuckoy:
I've decided to use the dictionary format config that the boto3 client uses instead of <key>: <ARN>(String) to be consistent.

`"dead_letter_config": {
    "target_arn": "<arn here">
}`

chuckoy:
Just popped up in my mind. We could also include an option where the user could opt to deploy/create a new SQS/SNS queue/topic upon deploy so everything's done from Zappa.

I'll have to study the codebase even further for this, though. This is the current limit of my limited knowledge of Zappa.

Miserlou:
Great, looking forward to reviewing and merging this tomorrow. Auto-queue creation stuff relates to https://github.com/Miserlou/Zappa/issues/603

Miserlou:
Fixed by PR.

PR #740, opened 2017-03-23
Add config for dead letter queue

chuckoy:
## Description
Add config option in `zappa_settings.json` to allow configuration of DLQ when Lambda async call fails thrice.

I've decided to use the dictionary format config that the boto3 client uses instead of `<key>: <ARN>(String)` to be consistent.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/738 Add an entry in zappa_settings to configure the Dead Letter Queue to which triple failed async invokes are sent to

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -167,3 +167,20 @@ Now in your application you can use:
 
     import os
     db_string = os.environ.get('DB_CONNECTION_STRING')
+
+Linking to a Dead Letter Queue
+==============================
+
+If you want to utilise `AWS Lambda's Dead Letter Queue feature <http://docs.aws.amazon.com/lambda/latest/dg/dlq.html>`_, you can set them in your ``zappa_settings.json``:
+
+::
+ 
+    {
+        "dev": {
+            ...
+            "dead_letter_arn": "your_sns_or_sqs_arn"
+        },
+        ...
+    }
+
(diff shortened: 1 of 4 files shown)


chuckoy:
Am I correct in that if the boto3 Lambda client is passed an empty dictionary in `DeadLetterConfig` it will simply ignore it? I assumed so from how `vpc_config` was implemented in the code.

chuckoy:
Why did the coverage for `tests_middleware` decrease? I didn't touch `middleware.py` or `wsgi.py`. I can't imagine having that much of an effect for `test_handler` too.

Should I rebase my default value empty dict to the first commit?

Miserlou:
Awesome stuff!

I wouldn't worry about Coveralls here, it's being really funky lately.

This looks okay, but since it's only a single value in the config dictionary, I'd prefer if we just used `dead_letter_arn` in the settings then built the dictionary ourselves, just to make it more simple to configure and use.

It'd also be great if you could add this setting into the `README.md` documentation, along with a link on how to use this feature, or maybe even a paragraph in the Readme on how to use it.

Think you can do that?

chuckoy:
I've added a paragraph in `docs/advanced.rst` as well as a link to the AWS docs for it.

Is how I'm building the dictionary and passing the argument the right way to do it?

chuckoy:
Currently working on adding automated topic/queue creation

chuckoy:
It's not as trivial as I thought. I'd rather get this merged in and create a different pull request for dlq resource creation.

Issue #707, opened 2017-03-09
Feature request - allow specifying usage plan


danielwhatmuff:
## Context

When using an API key which has a usage plan, it is required that new API stages are added to the usage plan before requests are allowed.

It would be good to be able to specify a usage plan ID and then zappa creates the association on deploy. Also removes the association on undeploy.

Miserlou:
+1, want to PR?

danielwhatmuff:
$ aws apigateway update-usage-plan --usage-plan-id APIID --patch-operations '[{"op":"add","path":"/apiStages","value":"APIID:sandbox"}]'
$ aws apigateway create-base-path-mapping --domain-name CUSTOMDOMAIN --rest-api-id APIID --stage STAGENAME --base-path ""

danielwhatmuff:
These are the commands Im running each time I undeploy / deploy with an API key, to re-associate the new API. Yea I'll see if I can find time to do a PR.

danielwhatmuff:
Updated to include custom domain and base path mapping too.

danielwhatmuff:
PR created, but need to do some testing... https://github.com/Miserlou/Zappa/pull/743

romabysen:
FWIW, this the post hook I've been using for this:
```python
import boto3


def post(zappa_cli):
    if zappa_cli.command == 'deploy':
        api_id = zappa_cli.zappa.get_api_id(zappa_cli.lambda_name)
        session = boto3.session.Session(profile_name=zappa_cli.profile_name,
                                        region_name=zappa_cli.aws_region)
        client = session.client('apigateway')
        patch_ops = [
            {
                'op': 'add',
                'path': '/apiStages',
                'value': '{}:{}'.format(api_id, zappa_cli.api_stage)
            }
        ]
        plan_id = zappa_cli.environment_variables['USAGE_PLAN_ID']
        client.update_usage_plan(usagePlanId=plan_id, patchOperations=patch_ops)
```

alfonsomhc:
[...]

PR #743, opened 2017-03-28
Adding attachment of API stage to an existing usage plan and custom domain

danielwhatmuff:
## Description

Haven't tested this yet, but would appreciate feedback...

This PR is for allowing new settings to specify 1. an existing usage plan ID and 2. an existing API gateway custom domain and a base path mapping. When specified prior to zappa deploy it should attach each to the newly created API and both should be removed as part of the undeploy.

## GitHub Issues
Related - https://github.com/Miserlou/Zappa/issues/707


danielwhatmuff:
Closing to get the test passing locally...

nickovs:
Was there ever any progress made in adding this functionality? It would be very useful to have this.

PR #756, opened 2017-03-29
allows wrapped function to be called synchronously as well with func.sync()

schuyler1d:
…sync() -- also removes need to trace stack to route things

## Description
* Use case: func = zappa.async.task(func) called synchronously.  now doable with func.sync()
* Also removes stack tracing requirement to route properly

## GitHub Issues
#603 evolution

### DIFF
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -7,7 +7,6 @@ def hello_world(environ, start_response):
         subject = escape(parameters['subject'][0])
     else:
         subject = 'World'
-    schedule_me()
     start_response('200 OK', [('Content-Type', 'text/html')])
     return ['''Hello {subject!s}
     Hello {subject!s}!
--- a/tests/test_app.py
+++ b/tests/test_app.py
@@ -1,3 +1,4 @@
+from zappa.async import task
 from cgi import parse_qs, escape
 
 def hello_world(environ, start_response):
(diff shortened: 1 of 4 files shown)


schuyler1d:
I'm closing this, because the main use case was being able to call task.sync() but I realized that I can just call task.func().

I still think it's better not to look up the call stack and better to call the function directly, but that would be a smaller patch

schuyler1d:
I'm closing this, because the main use case was being able to call task.sync() but I realized that I can just call task.func().

I still think it's better not to look up the call stack and better to call the function directly, but that would be a smaller patch

Miserlou:
@schuyler1d - do you mind submitting that as an isolated patch? In Slack, we realized an important use case that's currently broken that would be fixed by avoiding the stack inspection: being able to call async tasks from _other_ async tasks currently won't work (but will with `run`).

michi88:
also, when I do `run(create_some_stuff, *args, **kwargs, service='sns')` and `create_some_stuff` is a `@task_sns` decorated function the run command doesn't work.

Error is: `AttributeError: 'task_sns' object has no attribute '__name__'`. 

Which happens in `zappa.async.get_func_task_path`.

When the `create_some_stuff` task is not decorated we can currently create tasks with `run` from within other tasks.

michi88:
Oh, and when using `zappa invoke env 'some_func_that_calls_a_task_decorated_func'` the function is also executed synchronously.

Issue #744, opened 2017-03-28
Project directories with the same name as the virtualenv are excluded from project zip, even if they're not virtualenvs


evandrone:
If there's a directory in the project directory with the same name as the virtualenv, that directory will be excluded from the lambda zipfile. 

That makes sense if that directory *is* the virtualenv, --this is the default behavior with `virtualenv`, putting the environment directly in the project file.  However, if your virtualenv is stored someplace else -- default behavior for Virtualenvwrapper's `mkvirtualenv`, the wrong directory could be excluded.  

## Possible Fix

PR will be added shortly. zappa.py line 390 
goes from 
`excludes = ZIP_EXCLUDES + exclude + [split_venv[-1]]`
to 
`excludes = ZIP_EXCLUDES + exclude + [venv]`

## Steps to Reproduce

Running this script creates two zip files in `~/Desktop/samename_test`.  The first file, created under a virtualenv with the same name as a submodule of the project, excludes that directory.  The second, created with a differently named virtualenv, contains the `samename` directory.
```shell
#!/usr/bin/env bash

mkdir ~/Desktop/samename_test
cd ~/Desktop/samename_test

# ====================================================
# = # trivial Flask app with dependency on submodule =
# ====================================================
cat > my_app.py <<'EOM'
from flask import Flask
app = Flask(__name__)
from samename import samename_app

@app.route('/')
def index():
    samename_app.print_hello()
    return ("Hello, world!", 200)
    
# We only need this for local development.
if __name__ == '__main__':
    app.run()    

EOM

# ===========================================
# = Make submodule and put a function in it =
# ===========================================
mkdir samename
touch samename/__init__.py
cat > samename/samename_app.py <<'EOM' 
def print_hello():
    print("Hello from samename.samename_app")

EOM

# ========================
# = Basic Zappa settings =
# ========================
cat > zappa_settings.yml <<'EOM' 
dev: 
    project_name: samename_test_proj
    lambda_description: zappa bug replication
    s3_bucket: zappa.samename.bug
    aws_region: us-west-1
    
EOM

# =========================================================
# = Make a virtualenv with the same name as the submodule =
# =========================================================
mkdir -p ~/Desktop/venvs

virtualenv ~/Desktop/venvs/samename
. ~/Desktop/venvs/samename/bin/activate
pip install flask zappa
zappa package dev
# Zip file won't contain a `samename` directory

# ==============================================================
# = Make a virtualenv with a different name from the submodule
# ==============================================================
[...]

PR #757, opened 2017-03-29
Pr/package excludes venv

evandrone:
## Description
Fixes bug introduced in PR 745.  `shutil.ignore_patterns()` doesn't work with absolute paths. 

Iff  the virtualenv is inside the project directory, exclude the virtualenv from the zipped lambda package

Note that this isn't a perfect solution.  If someone had a project dir with a package called `venv_name` in the project directory and a virtualenv named `venv_name` inside that (`project_dir/venv_name/venv_name`),  then this would exclude valid user code in `project_dir/venv_name`.  

But... that would be a strange thing to do and the cure seemed like more complication than that peculiar edge case was worth.  

## GitHub Issues

This is a revised solution for #744


Miserlou:
Would you mind rebasing this so that I can merge it in cleanly?

evandrone:
No problem. Done.

evandrone:
BTW, I'd be happy to wrote some tests against issue #744, but in my (cursory) look around the tests dir, I didn't see how a virtualenv gets set up for the tests.  Can you point me to where that happens?

jneves:
@evandrone can you update this so I can run a few tests on it? Thanks!

Issue #766, opened 2017-04-03
Zappa doesn't play well with chargebee-python


FrederikP:
## Context
I'd like to use the chargebee client library (https://pypi.python.org/pypi/chargebee/2.2.7) in my zappa app. It doesn't work and I think it's due to them storing their ssl cert in a subfolder of the library.

## Expected Behavior
I want to be able to use the chargebee client lib without getting an error.

## Actual Behavior
An error is thrown:
AttributeError: 'module' object has no attribute 'CERT_NONE'
  File "tmp/pip-build-tjL06K/chargebee/chargebee/models/subscription.py", line 36, in list
  File "tmp/pip-build-tjL06K/chargebee/chargebee/request.py", line 13, in send_list_request
  File "tmp/pip-build-tjL06K/chargebee/chargebee/request.py", line 23, in send
  File "tmp/pip-build-tjL06K/chargebee/chargebee/http_request.py", line 35, in request
  File "tmp/pip-build-tjL06K/chargebee/chargebee/compat.py", line 52, in set_cert

My theory is that the "import ssl" statement in the chargebee library actually picks up its own subfolder ("ssl") which contains the certificate and the __init__.py that Zappa generating while packaging, instead of picking up the standard ssl package.

## Possible Fix
Right now https://github.com/Miserlou/Zappa/blob/master/zappa/zappa.py#L541 creates an __init__.py file if not available in the folder, while going through all folders. If we could only add this file, if there are other python files available, maybe that would fix it.

## Steps to Reproduce
1. create test chargebee account
2. import chargebee library
3. configure chargebee library with api key
4. call any resource using the client

## Your Environment
* Zappa version used: 0.40.0
* Operating System and Python version: ubuntu, python 2.7

Miserlou:
Thanks for this bug report! Maybe you can use `"delete_zip": false` to inspect the package that it produces and compare it to your client package?

FrederikP:
I actually did that and there was an init file in the zip in that subfolder, while there is none in my local venv. On Apr 3, 2017 3:22 PM, Rich Jones <notifications@github.com> wrote:Thanks for this bug report! Maybe you can use "delete_zip": false to inspect the package that it produces and compare it to your client package?

FrederikP:
This is what I get:

```
$ ls /venv/lib/python2.7/site-packages/chargebee/ssl/
ca-certs.crt

$ unzip -l app-dev-1491223335.zip | grep chargebee/ssl
[...]

PR #769, opened 2017-04-04
Don't always create a __init__.py file for any folder where it doesn't exist

FrederikP:
## Description
 
As described in issue 766 the way `__init__.py` files are created in libraries can lead to problems. In my case a folder named ssl becomes a python module overwriting the standard ssl module and thus breaking the library I'm trying to use.
This change disables the creation of `__init__.py` files if there are no sub folders and no .py/.pyc files in the folder we are looking at.
As I was not sure why the `__init__.py` files are even created, I don't know if this change is satisfactory. For me this would fix my problems with the client library.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/766

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -30,7 +30,7 @@ from zappa.letsencrypt import get_cert_and_update_domain, create_domain_key, cre
 from zappa.util import (detect_django_settings, copytree, detect_flask_apps,
                         add_event_source, remove_event_source,
                         get_event_source_status, parse_s3_url, human_size, string_to_timestamp,
-                        validate_name, InvalidAwsLambdaName)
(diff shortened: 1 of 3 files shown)


Miserlou:
Hey @FrederikP - this looks like a good solution, can you please move that function to util, add docstrings and a test, and then add a comment linking back to the ticket, as per https://raw.githubusercontent.com/Miserlou/Zappa/master/.github/PULL_REQUEST_TEMPLATE.md - thanks!

FrederikP:
ok will do.
ignore the last commit. didn't see your comment

FrederikP:
@Miserlou Is that okay now?

Miserlou:
Looks good now, thanks!

Issue #773, opened 2017-04-05
https connection is not reflected in wsgi environment


oliviersels:
Using zappa with a custom domain and https connection is not reflected in the wsgi environment. The values for `SERVER_PORT` and `wsgi.url_scheme` are still 80 and http. This can be solved by looking at the X-Forwarded-Port and X-Forwarded-Proto header.

https://github.com/Miserlou/Zappa/blob/36ff4941f6ad37276944f16c12e101e461f9767e/zappa/wsgi.py#L84

I will open a pull request with a fix.

See http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html for more information

Miserlou:
Merged and fixed! Thank you!

Issue #772, opened 2017-04-05
wsgi remote_addr can be user spoofed


oliviersels:
### Summary
The `REMOTE_ADDR` wsgi environment value can be easily spoofed by a user. Zappa takes the first value from the `X-Forwarded-For` header. This should be changed to the second to last one.

https://github.com/Miserlou/Zappa/blob/36ff4941f6ad37276944f16c12e101e461f9767e/zappa/wsgi.py#L71-L73

I will open a pull request with a fix.

### Explanation
AWS adds ips to the back of the existing `X-Forwarded-For` header. The last one is the ip of the cloudfront proxy server while the second to last is the real client ip. Everything before that can be user supplied.

Currently if a client already sends a `X-Forwarded-For` header that value will then be used by Zappa as the remote_addr instead of the client ip.

See http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/x-forwarded-headers.html for more information

aehlke:
This was merged, so this can now be closed?

Miserlou:
Yep

PR #774, opened 2017-04-05
Wsgi environment improvements

oliviersels:
This includes fixes for:
 - #772 wsgi remote_addr can be user spoofed
 - #773 https connection is not reflected in wsgi environment


Miserlou:
Thanks, I've wanted to fix that remote IP one for a long time. Looks good.

PR #775, opened 2017-04-05
all func.sync() in funcs decorated by async.task

schuyler1d:
also, make async.task a function to pass inspect.getargspec, and allow calling synchronously with func.sync()

## GitHub Issues
#603 related


mcrowson:
if you want to test calling lambda, you can set AWS_REGION and AWS_LAMBDA_FUNCTION_NAME as environment variables, and use placebo to mock some boto calls.

schuyler1d:
fixes #834 
some explicit feedback of what more we need on this would be helpful

Miserlou:
This is the third time I'm looking at it and I think I'm getting closer to wrapping my head around it.

`_run_async` is still missing a docstring. More generally though, I'd like to see the parent comment replaced with something _explanatory_ about the whole code path. I just think it looks too much like voodoo. (I'm not normally strict about this, but if we could use [Google Style Python Docstrings](http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) here that'd help me out.)

Sorry for being a stickler on this one, I'm sure your solution is good, it just that I'm kind of stupid and it needs to be stupid-proof.

schuyler1d:
amen to making things clear.  I added a 'lifetime of an async function' document on the header, which I hope de-voodoos some of it (and also helps future changes, including future-me :-)

Miserlou:
Okay, this is a really good looking PR now!

I'm going to manually test this, but I think we should be able to get this merged in shortly. Thanks for being so accommodating @schuyler1d

Issue #780, opened 2017-04-07
NoRegionError from boto on init when no default profile in aws credentials file


pdpol:
Running `zappa init` from a Django project gave me the following traceback. Easy workaround was to change the AWS profile I intended to deploy the project to to `default`, but I personally try not to keep a `default` profile in my credentials file to be explicit about which AWS env I'm working in.

```
Traceback (most recent call last):
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/zappa/cli.py", line 2187, in handle
    sys.exit(cli.handle())
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/zappa/cli.py", line 385, in handle
    self.init()
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/zappa/cli.py", line 1192, in init
    self.check_venv()
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/zappa/cli.py", line 2154, in check_venv
    temp_zappa = Zappa()
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/zappa/zappa.py", line 232, in __init__
    self.lambda_client = self.boto_session.client('lambda', config=long_config)
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/boto3/session.py", line 263, in client
    aws_session_token=aws_session_token, config=config)
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/botocore/session.py", line 836, in create_client
    client_config=config, api_version=api_version)
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/botocore/client.py", line 70, in create_client
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/botocore/client.py", line 224, in _get_client_args
    verify, credentials, scoped_config, client_config, endpoint_bridge)
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/botocore/args.py", line 45, in get_client_args
    endpoint_url, is_secure, scoped_config)
  File "/Users/pdpol/.virtualenvs/weather/lib/python2.7/site-packages/botocore/args.py", line 103, in compute_client_args
    service_name, region_name, endpoint_url, is_secure)
[...]

PR #785, opened 2017-04-09
201704 init venv check fix

pdpol:
Addressing a couple things as per [this issue](https://github.com/Miserlou/Zappa/issues/780).

* Made `get_current_venv` a static method so that init doesn't need to create a temp Zappa instance just to call that
* Added a profile detection bit in init

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -906,7 +906,7 @@ class TestZappa(unittest.TestCase):
                 }
             } },
             { 'profiles' : {
-                'default' : {
+                'radical' : {
                     'region' : 'us-east-1'
                 },
                 'another' : {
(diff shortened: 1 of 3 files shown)


Miserlou:
Hey @pdpol - thanks submitting this!

The changes look good, but I'd prefer it if the behavior will only emerge if there are multiple profiles or no profile at all. It is unnecessary to ask if the choice of `default` is okay:

```
AWS Lambda and API Gateway are only available in certain regions. Let's check to make sure you have a profile set up in one that will work.

We only found one profile: default. Is this okay? (default 'y') [y/n]: y
```

Also, please format the text so that there is no extra linebreak, so that each "section" of `init` corresponds to a different field of the generated config.

Thank you for participating in this fun media project!

pdpol:
@Miserlou Cool, PR updated with changes.

Issue #776, opened 2017-04-05
when slim_handler is used MySQLdb cannot be found by Django


timj98:
this issue is similar to #755 
when using the zappa_settings.json configuration to turn on "use_precompiled_packages":true and then turning on "slim_handler": true I get error below: 

`1491418915817] Error loading MySQLdb module: libmysqlclient.so.18: cannot open shared object file: No such file or directory: ImproperlyConfigured
Traceback (most recent call last):
  File "/var/task/handler.py", line 508, in lambda_handler
  return LambdaHandler.lambda_handler(event, context)
  File "/var/task/handler.py", line 239, in lambda_handler
  handler = cls()
  File "/var/task/handler.py", line 142, in __init__
  wsgi_app_function = get_django_wsgi(self.settings.DJANGO_SETTINGS)
  File "/var/task/django_zappa_app.py", line 20, in get_django_wsgi
  return get_wsgi_application()
  File "/private/var/folders/7c/lt6hvzr97z919lwy5yxyfs7w0000gn/T/pip-build-YiIZfn/django/django/core/wsgi.py", line 14, in get_wsgi_application
  File "/tmp/dailyfantasyser/django/__init__.py", line 18, in setup
  apps.populate(settings.INSTALLED_APPS)
  File "/tmp/dailyfantasyser/django/apps/registry.py", line 108, in populate
  app_config.import_models(all_models)
  File "/tmp/dailyfantasyser/django/apps/config.py", line 198, in import_models
  self.models_module = import_module(models_module_name)
  File "/usr/lib64/python2.7/importlib/__init__.py", line 37, in import_module
  __import__(name)
  File "/private/var/folders/7c/lt6hvzr97z919lwy5yxyfs7w0000gn/T/pip-build-YiIZfn/django/django/contrib/auth/models.py", line 41, in <module>
  File "/private/var/folders/7c/lt6hvzr97z919lwy5yxyfs7w0000gn/T/pip-build-YiIZfn/django/django/db/models/base.py", line 139, in __new__
  File "/private/var/folders/7c/lt6hvzr97z919lwy5yxyfs7w0000gn/T/pip-build-YiIZfn/django/django/db/models/base.py", line 324, in add_to_class
  File "/private/var/folders/7c/lt6hvzr97z919lwy5yxyfs7w0000gn/T/pip-build-YiIZfn/django/django/db/models/options.py", line 250, in contribute_to_class
[...]

PR #789, opened 2017-04-13
#776 fix for libmysqlclient.so.18 error when using slim_handler

timj98:
Added include to zappa_settings
Added ctypes to handler to load the selected libraries directly to
PYTHONPATH

## Description
 
This PR fixes the issue where Django using MySQLdb looks for libmysqlclient.so.18 and could not find it in any path when using the slim_handler.  I creates a setting in the zappa_settings file for the specific library to add at handler init. 

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/776

### DIFF
--- a/README.md
+++ b/README.md
@@ -569,6 +569,7 @@ to change Zappa's behavior. Use these at your own risk!
             "Resource": "*"
         }],
         "iam_authorization": true, // optional, use IAM to require request signing. Default false. Note that enabling this will override the authorizer configuration.
+        "include": ["your_special_library_to_load_at_handler_init"], // load special libraries into PYTHONPATH at handler init that certain modules cannot find on path
         "authorizer": {
             "function": "your_module.your_auth_function", // Local function to run for token validation. For more information about the function see below.
             "arn": "arn:aws:lambda:<region>:<account_id>:function:<function_name>", // Existing Lambda function to run for token validation.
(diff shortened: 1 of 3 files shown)


Miserlou:
Okay, this looks like we're on the right track.

Two things: 

1. Can we include some sane defaults for INCLUDE so that MySQL and others affected by this in lambda-packages all work out of the box?

2. Please change the `print` statements to `print()` so that we don't have to update them when P3 drops.

timj98:
That make sense, I’ll add those tonight.  I will include the defaults for MySQL-Python and mysqlclient.  I was concerned about the coveralls test and resultant reduction in coverage.  

TimJ

Miserlou:
I've merged this in now, thank you for going in on this!

Just as a style point, you included a line of _exceptionally_ ugly code:

```
settings_s += "INCLUDE=[" + ','.join(("'","'")).join(self.stage_config.get('include', [])).join(("'","'")) + "]\n"
```
Since it's already a list, we can simply cast it to a string:

```
include = self.stage_config.get('include', [])
if len(include) >= 1:
    settings_s += "INCLUDE=" = str(include) + '\n'
```
I've gone ahead and changed this already, but in future it's generally good to try to keep the code as short, readable and WTF-free as possible!

Thanks again,
R

timj98:
Thanks.  I appreciate the opportunity and the feedback. 

TimJ

Issue #721, opened 2017-03-15
Attempt to Import Event Sources


mcrowson:
## Context

 
Feature Request to warn about typos.

In the zappa settings for events,  If there are typos in event source function names, such as my recent use of a dash instead of an underscore, the error isn't apparent until zappa responds to the event and is unable to import the function used in response to the event.

## Expected Behavior

For each event source in the stage, import the function to ensure it is reachable. Do this prior to uploading the zip. If the function is not reachable, raise an exception and message for the user.

"You have an event setup to call {{ function_name }}, but that that function cannot be imported. Please check for typos and ensure that the function is reachable."

## Actual Behavior

Assumes function names are reachable.

Miserlou:
Related: https://github.com/Miserlou/Zappa/issues/78

bcongdon:
Does Zappa do this type of check on event and exception handlers? If not, that might be good to do as well.

mcrowson:
It does not currently. That was my use case as well.

Issue #78, opened 2016-05-11
Add a pre-deployment sanity check


Miserlou:
Just display warnings if the app_module doesn't exist to avoid the

     UnboundLocalError: local variable 'app_module' referenced before assignment 

that happens from putting the wrong values in zappa_settings.

tipu:
on it.

tipu:
I'm going to put the addition for this in `zappa/cli.py:create_package()`. Does this seem like a suitable place?

Miserlou:
This would probably be better to but in zappa.py so that it could be used by other clients. We should expect that this function may grow over time.

tipu:
@Miserlou `zappa/zappa.py` appears to be unaware of the `zappa_settings` reference that `zappa/cli` holds onto. Should I inject the `zappa_settings` into `zappa/zappa.py` or should I implement it in the `update` method within ZappaCLI?

collingreen:
It would be helpful to check types on config values -- for example, memory_size, parameter_depth, timout_seconds, etc all need to be numbers, not strings. If they are strings, zappa will zip your project, upload it, then fail with an error:
> Parameter validation failed:
Invalid type for parameter MemorySize, value: 128, type: <type 'unicode'>, valid types: <type 'int'>, <type 'long'>

Miserlou:
This recently happened to somebody in IRC:

```
[1471899440322] No module named your_module
[1471899440322] No module named your_module: ImportError
Traceback (most recent call last):
 File "/var/task/handler.py", line 345, in lambda_handler
   return LambdaHandler.lambda_handler(event, context)
 File "/var/task/handler.py", line 122, in lambda_handler
   return cls().handler(event, context)
 File "/var/task/handler.py", line 317, in handler
   raise e
ImportError: No module named your_module
```

Could we check if the specified module exists before uploading?

PR #802, opened 2017-04-22
Sanity checking for function imports

bcongdon:
## Description
 
* This PR adds import-checking for user-specified functions. Before uploading, try to import / check that `app_function`, `event` functions, and `exception_handler` functions can be imported and are accessible. Presents an error message if unable to import.
* Adds test for the above changes

## GitHub Issues

#721
#78


Miserlou:
I like it

Issue #800, opened 2017-04-22
Error loading psycopg2 module on Python 3.6 (when on Lambda)


nikbora:
## Context
Shiny new Zappa 0.41 and Python 3.6

Django==1.11
djangorestframework==3.6.2
psycopg2==2.6.1
zappa==0.41.0
...

## Expected Behavior

It works fine locally, but on Lambda the following happens:

```
[1492873312435] File "/var/task/django_zappa_app.py", line 20, in get_django_wsgi
[1492873312435] return get_wsgi_application()
[1492873312435] File "/var/task/django/core/wsgi.py", line 13, in get_wsgi_application
[1492873312435] django.setup(set_prefix=False)
[1492873312435] File "/var/task/django/__init__.py", line 27, in setup
[1492873312435] apps.populate(settings.INSTALLED_APPS)
[1492873312435] File "/var/task/django/apps/registry.py", line 108, in populate
[1492873312435] app_config.import_models()
[1492873312435] File "/var/task/django/apps/config.py", line 202, in import_models
[1492873312436] self.models_module = import_module(models_module_name)
[1492873312436] File "/var/lang/lib/python3.6/importlib/__init__.py", line 126, in import_module
[1492873312436] return _bootstrap._gcd_import(name[level:], package, level)
[1492873312436] File "<frozen importlib._bootstrap>", line 978, in _gcd_import
[1492873312436] File "<frozen importlib._bootstrap>", line 961, in _find_and_load
[1492873312436] File "<frozen importlib._bootstrap>", line 950, in _find_and_load_unlocked
[1492873312436] File "<frozen importlib._bootstrap>", line 655, in _load_unlocked
[1492873312436] File "<frozen importlib._bootstrap_external>", line 678, in exec_module
[1492873312436] File "<frozen importlib._bootstrap>", line 205, in _call_with_frames_removed
[1492873312436] File "/var/task/django/contrib/auth/models.py", line 4, in <module>
[1492873312436] from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
[1492873312436] File "/var/task/django/contrib/auth/base_user.py", line 52, in <module>
[1492873312436] class AbstractBaseUser(models.Model):
[1492873312436] File "/var/task/django/db/models/base.py", line 124, in __new__
[1492873312436] new_class.add_to_class('_meta', Options(meta, app_label))
[1492873312436] File "/var/task/django/db/models/base.py", line 330, in add_to_class
[...]

PR #803, opened 2017-04-23
Support for manylinux wheels Python 3.6 package downloading.

nikbora:
## Description

Since AWS Lambda and Zappa now support Python 3.6 this change makes sure that the correct version of the manylinux-wheels package is downloaded depending on the target runtime. 

## GitHub Issues

Addresses https://github.com/Miserlou/Zappa/issues/800.
A separate change will be required once work to fix lambda-packages to be runtime aware is done here https://github.com/Miserlou/lambda-packages/issues/43.


Miserlou:
👍

Issue #635, opened 2017-01-30
Allow user to exclude packages from .py trimming


bbangert:
## Context

Zappa excludes .py files as it packages up Python directories if a .pyc/.pyo exists. However some tools need to find the .py file, in my case Pyramid's plugin extension locator.

## Expected Behavior
The zipped package should have .py files if needed as some Python packages expect to find them.

## Actual Behavior
All .py files are removed.

## Possible Fix
Provide a zappa settings option to specify packages that should be excluded from this process.

## Steps to Reproduce
1. Package up anything that has .pyc files
2. Look at the resulting zip that has no .py files.

Miserlou:
What should the interface to this be?

bbangert:
Probably something like:
```json
{
    "exclude_py_removal": ["pyramid_mako"]
}

mcrowson:
Why are the .py files removed in the first place? Space savings?

PR #804, opened 2017-04-23
Setting for excluding files from .py trimming

bcongdon:
## Description
 
* Adds new `exclude_py_removal` setting. Files included in this list will have their `.py` files included in the Lambda Zip file even if a `.pyo` or `.pyc` file of the same name exists.
* Adds a test for the above changes

## GitHub Issues

#635


bcongdon:
I'm definitely open to changing the `exclude_py_removal` setting name. Might be a bit confusing as it's a double-negative... Maybe, `force_py_inclusion`?

PR #807, opened 2017-04-25
Revert "Sanity checking for function imports"

Miserlou:
I shouldn't have merged this without testing more thoroughly.

Doesn't work for non-package Flask apps.


Miserlou:
@bcongdon - Had to revert this, want to try again? 

Fails for working applications with `Error: You have tried to reference the module app when importing app.app, but this module cannot be imported. Please check for typos and ensure that this module is reachable.`.

bcongdon:
Sure, I'll look into it

bcongdon:
I wasn't able to recreate this completely. However, I did notice that if the app.py file is reachable, but its dependencies aren't installed, an ImportError is still raised - causing that error message to be triggered.

In your working app, does `python -c "from app import app"` return successfully?

When I used a completely barebones app file:
```
from flask import Flask
app = Flask(__name__)


@app.route('/')
def hello_world():
    return 'Hello, World!'

if __name__ == '__main__':
    app.run()
```

... and basic zappa settings ...
```
{
    "dev": {
        "app_function": "app.app", 
        "aws_region": "us-east-1", 
        "profile_name": "default", 
        "s3_bucket": "zappa-qv489ct0v"
    }
}
```
and `flask` was present in the virtualenv, I'm not getting an error. I wonder if there's something more going on here?

PR #811, opened 2017-04-26
Force fetching pandoc

fifthecho:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
+venv/
+.venv/
 build/
 develop-eggs/
 dist/
@@ -77,3 +79,9 @@ target/
 *~
 tests/zappa_settings.json
 tests/example_template_outputs
+
+# VS Code stuff
+.vscode
+
+# Pandoc temporary files
+pandoc*
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,6 +10,7 @@ hjson==2.0.2
 jmespath==0.9.2
 kappa==0.6.0
 lambda-packages==0.14.2
+pypandoc>=1.4
 python-dateutil==2.6.0
 python-slugify==1.2.4
 PyYAML==3.12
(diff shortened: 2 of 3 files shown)


gdw2:
Not working for me.  Don't you need to downlaod pandoc before you try to import it?

```
(env36_64) C:\Users\gwarner\zappa\Zappa>pip install .
Processing c:\users\gwarner\zappa\zappa
    Complete output from command python setup.py egg_info:
      File "C:\Users\gwarner\AppData\Local\Temp\pip-3q4ev8qg-build\setup.py", line 7, in <module>
        from pypandoc import convert
    ModuleNotFoundError: No module named 'pypandoc'

    During handling of the above exception, another exception occurred:

      File "<string>", line 1, in <module>
      File "C:\Users\gwarner\AppData\Local\Temp\pip-3q4ev8qg-build\setup.py", line 15, in <module>
        README = open(os.path.join(os.path.dirname(__file__), 'README.md')).read()
      File "c:\users\gwarner\zappa\env36_64\lib\encodings\cp1252.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 48630: character maps to <undefined>
```

fifthecho:
@gdw2 https://github.com/Miserlou/Zappa/pull/811/files#diff-2eeaed663bd0d25b7e608891384b7298
That should automatically download pandoc when building.

gdw2:
But how do you ensure that the `pypandoc` module is installed before you call `download_pandoc()` (which comes from the `pypandoc` module)?

gdw2:
Should the `requirements.txt` be read and processed before `download_pandoc()`?

Miserlou:
To be honest, I don't think this is the right solution to the problem.

Literally the only reason this requirement is even included is so that when I publish it, the Markdown README looks nice on PyPI, which only supports RST for now.

There is no need to install pypanddoc/pandoc for every single Zappa user, literally only me. I'd rather just figure out what's changed that's causing the exception handler to break.

Thanks for your contribution, but I have to decline this PR.

Issue #809, opened 2017-04-26
Issue installing 0.41.1 on Windows


fifthecho:
## Context
When installing in Python 3.6 on Windows, install fails with:

```
Collecting zappa
  Using cached zappa-0.41.1.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "C:\Users\fifthecho\AppData\Local\Temp\pip-build-fp131vse\zappa\setup.py", line 7, in <module>
        from pypandoc import convert
    ModuleNotFoundError: No module named 'pypandoc'

    During handling of the above exception, another exception occurred:

    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\fifthecho\AppData\Local\Temp\pip-build-fp131vse\zappa\setup.py", line 10, in <module>
        README = open(os.path.join(os.path.dirname(__file__), 'README.md')).read()
      File "c:\python36\lib\encodings\cp1252.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 48630: character maps to <undefined>
```

## Your Environment

* Zappa version used: 0.41.1
* Operating System and Python version: Windows 10, Python 3.6.1

Miserlou:
Ah fuck.

fifthecho:
It's worth noting that Zappa 0.41.0 installs fine in Windows.

Miserlou:
Works on OSX/Py2.7. 

I don't have a Windows machine, can you see what changed? Something in the README file that Windows doesn't like?

gdw2:
I'm seeing this too.  I'll see what I can find.

fifthecho:
https://github.com/Miserlou/Zappa/pull/811 should fix this. It's using pypandoc to download pandoc and make sure it's in place. Also, pypandoc wasn't in the requirements so that was causing the issues as well.

gdw2:
For me, it seems that removing this line with the Japanese characters fixes the install:
https://github.com/Miserlou/Zappa/blame/master/README.md#L919

Not sure why this is appearing just now.

selepo:
When trying to install Zappa on Windows, I get another traceback, but still pandoc that seems to be the problem: 

`Collecting zappa
  Using cached zappa-0.41.1.tar.gz
    Complete output from command python setup.py egg_info:
[...]

PR #820, opened 2017-05-04
use io.open() and specify encoding utf-8 to fix python3/windows encod…

dazcode:
## Description
 
Fixes the windows/python3 decoding error when installing setup.py . This issue is caused by the differences in how Python2/Python3 handle data. The fix is compatible with Python 2 and 3 on windows and linux.

more info on this error:
https://ncoghlan-devs-python-notes.readthedocs.io/en/latest/python3/text_file_processing.html#what-changed-in-python-3

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/809

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,13 +1,14 @@
 import os
 import sys
 from setuptools import setup
+from io import open
 
 # Set external files
 try:
     from pypandoc import convert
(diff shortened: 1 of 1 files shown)


Miserlou:
Tusen takk.

Issue #816, opened 2017-05-02
Ability to provide EventPattern


takeda:
It looks like implementing this through zappa is currently not possible: https://aws.amazon.com/blogs/compute/building-a-dynamic-dns-for-route-53-using-cloudwatch-events-and-lambda/

Particularly equivalent of this command:
`aws events put-rule --event-pattern "{\"source\":[\"aws.ec2\"],\"detail-type\":[\"EC2 Instance State-change Notification\"],\"detail\":{\"state\":[\"running\",\"shutting-down\",\"stopped\"]}}" --state ENABLED --name ec2_lambda_ddns_rule`

Miserlou:
Why not possible? These should all be supported: http://docs.aws.amazon.com/lambda/latest/dg/invoking-lambda-function.html#supported-event-source-cloudwatch-events

Feel free to re-open if not, but I think it is.

takeda:
@Miserlou could you show how to define this in Zappa? I could not find this anywhere in the documentation.

I ended up modifying the code to get the functionality, but if there is a way to do it I would prefer to do it properly.

Miserlou:
Hm. How was the code modified?

From my phone..

takeda:
Added my changes (it's not suitable to be merged yet)

mjschultz:
@Miserlou as far as I can tell this is not supported by Zappa right now; at least for EC2 events from Cloudwatch Events.

I was _kind of_ able to get it working by using the following:
```
        "events": [{
            "function": "my_lambda.ec2_event",
            "event_source": {
                "arn": "arn:aws:events:us-east-1:<account-id>:rule/ec2-notifier",
                "enabled": true,
                "pattern": "{\"source\": [\"aws.ec2\"], \"detail-type\": [\"EC2 Instance State-change Notification\"]}",
                "events": []
            },
        }]
```

This allows kappa to create/update/delete the event (I believe). However, I had to tweak the zappa code that defines [`funk.name`](https://github.com/Miserlou/Zappa/blob/6ab48b0/zappa/utilities.py#L266) as the full `lambda_arn` to just the actual lambda name (the full ARN is too long and contains invalid characters for the CloudwatchEvents trigger). I just changed it to `lambda_arn.split(':')[-1]`, not sure if that's acceptable in the general case.

[...]

PR #823, opened 2017-05-06
Partial support for EventPattern.

takeda:
## Description
This is unfinished code for adding support for EventPattern, currently the handler also needs some work, I ended using the lambda_handler so I have more control of how things work.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/816

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -345,18 +345,18 @@ class LambdaHandler(object):
 
         # This is the result of a keep alive, recertify
         # or scheduled event.
-        if event.get('detail-type') == u'Scheduled Event':
+        if 'source' in event:
+            if event.get('detail-type') == u'Scheduled Event':
+                whole_function = event['resources'][0].split('/')[-1].split('-')[-1]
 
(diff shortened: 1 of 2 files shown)


zappa/handler.py
```diff
@@ -345,18 +345,18 @@ def handler(self, event, context):
 
         # This is the result of a keep alive, recertify
         # or scheduled event.
-        if event.get('detail-type') == u'Scheduled Event':
+        if 'source' in event:
```
jneves: Can you explain this change? I feel like I'm missing something here, as I don't see the change as being related to the PR.

takeda: This was submitted almost 2 years ago, so my memory is hazy.

From what I remember the event that I was looking for was handled incorrectly. I think distinguishing factor was that "source" key was set same as in "Scheduled Event" but the detail-type was different. I think this was the unfinished part where proper handling of different types supposed to be implemented. Previously the code just failed back to responding to http requests.

This PR can probably be closed, for my needs I realized that writing my own handler was actually way simpler than using a generalized module like zappa :)

Issue #841, opened 2017-05-10
Zappa installs two versions of pillow


bxm156:
## Context
Despite having Pillow==3.4.2 in my env. Zappa installs version 4.1.1 from manylinux. I added debug statements durning the package install process and discovered:
```
Calling update for environment dev..
Downloading and installing dependencies..
Installing lxml from lambda-packages
Installing psycopg2 from lambda-packages
Installing Pillow from lambda-packages
pyterminalsize from manylinux
Installing pillow from manylinux
Packaging project as zip..
```
I checked the package.zip, and sure enough PIL/__init__.py was version 4.1.1. It looks like zappa doesn't consider 'Pillow' and 'pillow' to be the same package, so it is re-installed in the manylinux step, and its instal, and overwrites the previous version.

After some more investigation in core.py, I realized the check for packages to install via manylinux might be invalid.  It checks to see if the name of the package to be installed is in lambda_packages, but the case is wrong.

## Expected Behavior
I would expect zappa not to even consider needing to download 'pillow' as a pre-compiled package, since its not in my virtualenv

## Your Environment
* Zappa version used: 0.42.2
* Operating System and Python version: py2.7

bxm156:
I tried the following fix, which fixed the duplicate package install issue for me:

https://github.com/bxm156/Zappa/commit/aa09e68ae75f12bf135957c28388374069de0a09

nikbora:
Hey @bxm156, I believe this is related to the fact that Zappa is ignoring version numbers of packages when looking at both lambda-packages and wheels. I have a PR for it here, do you mind checking out my branch of Zappa to see if this fixes your issue here? (i.e. correct pillow is downloaded). Maybe can kill multiple issues with one PR.

https://github.com/Miserlou/Zappa/pull/843

bxm156:
Closed by https://github.com/Miserlou/Zappa/pull/843

PR #843, opened 2017-05-11
More binary package fixes

nikbora:
* Zappa now checks package version when deciding if it has the correct lambda-package, or which wheels version to pull down. Before you would silently get the latest version no matter what version you had in your virtual env. Fixes https://github.com/Miserlou/Zappa/issues/800 (again) and probably https://github.com/Miserlou/Zappa/issues/841

* Changed how the progress bars work when downloading manylinux packages. You now get one per package. There is also some logging to let you know whether a lambda-packages or wheels version of the package is used, along with which version. Here an example:

```
zappa package test
Calling package for environment test..
Downloading and installing dependencies..
Downloading cffi-1.10.0-cp27-cp27mu-manylinux1_x86_64.whl
100%|████████████████████████████████████████████████████████████████████████████████████████████████| 393K/393K [00:00<00:00, 2.34MB/s]
Using lambda-packages binary for psycopg2 2.6.1
Downloading grpcio-1.3.0-cp27-cp27mu-manylinux1_x86_64.whl
100%|██████████████████████████████████████████████████████████████████████████████████████████████| 5.31M/5.31M [00:01<00:00, 4.49MB/s]
Downloading protobuf-3.3.0-cp27-cp27mu-manylinux1_x86_64.whl
100%|██████████████████████████████████████████████████████████████████████████████████████████████| 5.72M/5.72M [00:01<00:00, 4.08MB/s]
Packaging project as zip..
```

* Refactored the binary package code into one loop.

* Changed a few unit tests for manylinux wheels and lambda packages that were in fact doing nothing 🙀 

* Tested on both Python 2.7 and Python 3.6

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -470,23 +470,15 @@ class Zappa(object):
 
         copy_tree(temp_package_path, temp_project_path, update=True)
 
-        package_to_keep = []
-        if os.path.isdir(site_packages):
-            package_to_keep += os.listdir(site_packages)
-        if os.path.isdir(site_packages_64):
-            package_to_keep += os.listdir(site_packages_64)
-
         # Then the pre-compiled packages..
         if use_precompiled_packages:
             print("Downloading and installing dependencies..")
-            installed_packages = {package.project_name.lower(): package.version for package in
-                                  pip.get_installed_distributions() if package.project_name in package_to_keep or
-                                  package.location in [site_packages, site_packages_64]}
+            installed_packages = self.get_installed_packages(site_packages, site_packages_64)
 
             try:
                 for installed_package_name, installed_package_version in installed_packages.items():
 
-                    if self.should_use_lambda_package(installed_package_name, installed_package_version, lambda_packages):
+                    if self.should_use_lambda_package(installed_package_name, installed_package_version):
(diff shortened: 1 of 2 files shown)


bxm156:
What do you think about making the progress bars per package only show if the user passes '-v' ? I have a lot of packages and this can get really long.

nikbora:
Not a bad shout. I shall see how easy it would be to pass a verbose logging
flag around. @miserlou any thoughts on this?

Miserlou:
You can use `--quiet` to silence output.

zappa/core.py
```diff
@@ -588,20 +589,44 @@ def splitpath(path):
...
+            return False
+
+        # Binaries can be compiled for different package versions
+        # Related: https://github.com/Miserlou/Zappa/issues/800
+        if package_version != lambda_package_details[self.runtime]['version']:
+            return False
+
+        return True
+
+    def download_url_with_progress(self, url, file_stream):
```
Miserlou: Docstring please!

nikbora:
@Miserlou 👍 

I also discovered some unit tests were not quite doing what they were meant to so also changed those and added a couple more.

PR #852, opened 2017-05-15
Remove unused exceptions.

bhodorog:
Purely code cleanup. I was just browsing the code and realized this most likely forgotten code so I decided to PR for its removal.

I tracked down the commits which should have probably removed it to at least c8addf5 and 24e4438


Miserlou:
rm'ing dead code always appreciated.

Issue #851, opened 2017-05-14
Feature: keep_warm can take integer values


mcrowson:
The keep_warm feature is a true/false and it can keep your zappa application warm on 1 concurrent lambda. It would be nice to pass the number of concurrent lambdas you would like to keep warm.

## Expected Behavior

keep_warm takes a value, such as 100, and keeps 100 lambdas warm for you.

## Actual Behavior

keep_warm just takes true/false and either keeps 1 lambda warm or 0.

Miserlou:
I don't think that it really works that way. If it does, I don't know how to keep that many warm simultaneously, I thought it was sort of an all-or-nothing deal. There is the `keep_warm_expression` for setting the rate.

mcrowson:
I haven't looked at the code yet, but my gut is telling me to do it with multi threading lambda requests with a slight time.sleep so that you know each thread is hitting a new lambda.

mcrowson:
Another approach would be to have the keep_warm handler recursively call keep_warm lambda handlers with the requestResponse event type until we reach the desired number.

nikbora:
I am doing this at the moment by calling several async @task functions at the same time rather than using threads. They all hit a `/ping` like endpoint in the app (via API Gateway). In this endpoint there is a fake wait of a few hundred ms. This ensures that more than one Lambda is instantiated as the requests can't all be served by the same lambda at the same time. I am not sure how else to 'force' a cold start of more than one container.

I also had some non-conclusive evidence that a scheduled keeper warmer keeps a different container warm to the one that an APIG request would hit (i.e. even though keep warm just ran, an APIG request would cause a cold start). Related to: https://github.com/Miserlou/Zappa/issues/418

matse004:
see also an attempt here: #1790

tim-schier-frontier:
There is a way to spool up a guaranteed number of Lambda instances, but it doesn't solve for the inherent Lambda cost of n+1 cold starts. 
To run n Lambdas, you need to have n concurrent Lambda requests executing at once. To guarantee you have n Lambdas executing, you need to kick them off not as the default async execution requests (i.e. kick-off and forget), but synchronous ones (i.e. kick-off and wait for the response).

I wrote a little function that invokes itself up to n times, each time waiting for the subsequent invocation to finish before it finishes itself (it's recursive).

[...]

PR #861, opened 2017-05-17
Multiple keep warm

mcrowson:
## Description
 
The `keep_warm` setting now takes an integer for the number of concurrent containers to keep warm. This is done by the keep_warm_callback in the handler.py file using a thred pool to call zappa async tasks that each initialize the application.

Those initialized applications then wait 30 seconds before returning to allow for other other lambdas to cold start rather than reuse the newly warmed container.

Somewhat arbitrary values were set for the ThreadPool size and the sleep size. If the user has specified a `timeout_seconds` < 30, it will exit sooner. 

The time to run the keep_warm_callback with 200 as the value was 15 seconds. If the time.sleep is too low, then it is possible for some percentage of the keep_warm_lambda_initializers to use warm containers. 

## GitHub Issues

#851

### DIFF
--- a/README.md
+++ b/README.md
@@ -593,7 +593,7 @@ to change Zappa's behavior. Use these at your own risk!
             "token_source": "Authorization", // Optional. Default 'Authorization'. The name of a custom authorization header containing the token that clients submit as part of their requests.
             "validation_expression": "^Bearer \\w+$", // Optional. A validation expression for the incoming token, specify a regular expression.
         },
-        "keep_warm": true, // Create CloudWatch events to keep the server warm. Default true.
+        "keep_warm": 1, // Create CloudWatch events to keep the server warm. Default 1 lambda instance.
         "keep_warm_expression": "rate(4 minutes)", // How often to execute the keep-warm, in cron and rate format. Default 4 minutes.
         "lambda_description": "Your Description", // However you want to describe your project for the AWS console. Default "Zappa Deployment".
         "lambda_handler": "your_custom_handler", // The name of Lambda handler. Default: handler.lambda_handler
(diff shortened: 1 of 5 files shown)


mcrowson:
Following a discussion with @nikbora on slack, now setting the sleep time dynamically so it is least intrusive.

kevinluvian:
is there any update to this @mcrowson ?

mcrowson:
Not on my end. Feel free to take a stab at a PR on this. The challenge is going to be on making sure your concurrently making the requests and getting a cold lambda each time.

jneves:
Does this still make sense now that we have an upstream solution with provisioned concurrency? https://aws.amazon.com/pt/about-aws/whats-new/2019/12/aws-lambda-announces-provisioned-concurrency/

nikbora:
Probably not? However this does give you a cheaper way of trying to keep a
few instances warm as you are not paying for their "uptime" (provisioned is
pretty pricey https://aws.amazon.com/lambda/pricing/).

TamasNo1:
FWIW I tried to set up provisioned concurrency manually for our Zappa application, but it did not deliver the results expected:
- I manually had to create a new ALIAS for the latest Lambda version
- I had to update API Gateway to point to FUNCTION_NAME:ALIAS instead of just FUNCTION_NAME
- Set provisioned concurrency to 20 for ALIAS and wait until it became ready
- Run `ab -n200 -c10 https://MY_ENDPOINT` (concurrency of 10)

In the result, I still saw > 10s for p95, so cold start still occurred, even though I've set my concurrency to twice the amount I used to test it with. When I ran the test again, right after I got the results, the cold starts were gone. This suggests that provisioned concurrency is not a good enough solution to keep Zappa functions warm.

Issue #574, opened 2016-12-29
Cache Precompiled Packages for Faster Subsequent Updates


mcrowson:
https://github.com/Miserlou/Zappa/blob/569a576c7d4f5f3959bf7c29a53694be84b2207a/zappa/zappa.py#L422

When using precompiled packages, it downloads them each time an update is called. During development, the updates could happen frequently. So saving these precompiled packages locally between updates could speed things up.

Miserlou:
Was there ever an open PR for this? I feel like there was and its gone..

mcrowson:
I don't remember seeing one.

PR #862, opened 2017-05-17
Cached packages

mcrowson:
## Description
 
When manylinux wheels are downloaded, they are now saved into the local tmp working space. Subsequent calls to update/deploy then use those cached copies. 

## GitHub Issues

#574


Miserlou:
I had to merge in Nik's PR first but I'd like to get this one in too, any chance you'd mind rebaseing?

mcrowson:
With this new approach it looks like we are actually hitting pypi twice. Once in the elif, and then again when we go to download it.

PR #871, opened 2017-05-19
Pull zappa's version into its own python module.

bhodorog:
Yet another simple PR, but quite effective. 

Currently there is no clean, simple way to know what version of Zappa is contained in a AWS Lamda tarball which can be exported using AWS Console. 

Having the version as part of the python code is one of the recommended options by [Python Packaging User Guide](https://packaging.python.org/single_source_version/) and will enable to determine what Zappa version was running on that specific tarball. Once this PR is merged use one of the followings:
  - the simplest way is to just read the zappa/__init__.pyc file from the tarball. Although bytecode, the version will be easy to read.
```shell
$ cat zappa/__init__.pyc

C�Yc@sdZedkreGHndS(s0.42.0__main__N(t
```
  - or, from the directory containing the unpacked zappa package
```shell
$ python
>> import zappa
>> zappa.__version__
0.42.0
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,6 +2,7 @@ import os
 import sys
 from setuptools import setup
 from io import open
+from zappa import __version__
 
 # Set external files
 try:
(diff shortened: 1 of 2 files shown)


Miserlou:
I'd rather do this in the package __init__.py rather than than exec and a file for just that one value. #6 here https://packaging.python.org/single_source_version/

bhodorog:
I've changed it to [no. 6](https://packaging.python.org/single_source_version/).

Issue #818, opened 2017-05-02
`zappa certify` results in "No module named 'letsencrypt'" (with fix)


dtenenba:
When I type `zappa certify` I get the stack trace below. 

## Context

 

You need to update your template--it should be OK to use python 3 now! And in fact I am. I wonder if that is what flushed this bug out. 

## Expected Behavior

No stack trace

## Actual Behavior

```
(Werkzeug 0.12.1 (/Users/dtenenba/envs/lambdagoodies/lib/python3.5/site-packages), Requirement.parse('Werkzeug==0.12'), {'zappa'})
Calling certify for environment dev..
Are you sure you want to certify? [y/n] y
Certifying domain sub.domain.tld [ACTUAL DOMAIN CENSORED]..
Oh no! An error occurred! :(

==============

Traceback (most recent call last):
  File "/Users/dtenenba/envs/lambdagoodies/lib/python3.5/site-packages/zappa/cli.py", line 2333, in handle
    sys.exit(cli.handle())
  File "/Users/dtenenba/envs/lambdagoodies/lib/python3.5/site-packages/zappa/cli.py", line 450, in handle
    self.dispatch_command(self.command, environment)
  File "/Users/dtenenba/envs/lambdagoodies/lib/python3.5/site-packages/zappa/cli.py", line 545, in dispatch_command
    manual=self.vargs['manual']
  File "/Users/dtenenba/envs/lambdagoodies/lib/python3.5/site-packages/zappa/cli.py", line 1565, in certify
    from letsencrypt import get_cert_and_update_domain, cleanup
ImportError: No module named 'letsencrypt'

==============

Need help? Found a bug? Let us know! :D
File bug reports on GitHub here: https://github.com/Miserlou/Zappa
And join our Slack channel here: https://slack.zappa.io
Love!,
 ~ Team Zappa!
```

## Possible Fix

```diff
diff --git a/zappa/cli.py b/zappa/cli.py
index 68da8f9..01a6e20 100644
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -1435,7 +1435,7 @@ class ZappaCLI(object):
 
         # Let's Encrypt
         if not cert_location and not cert_arn:
-            from letsencrypt import get_cert_and_update_domain, cleanup
+            from zappa.letsencrypt import get_cert_and_update_domain, cleanup
             cert_success = get_cert_and_update_domain(
                     self.zappa,
                     self.lambda_name,

```

[...]

PR #874, opened 2017-05-21
fix compatibility for py2/py3 in certify command

brad-alexander:
## Description
Fixes issue https://github.com/Miserlou/Zappa/issues/818

- Updates import of zappa.letsencrypt to use explicit relative imports as previously it fails in python2
- Coerce keyauthorization to str as python3 was sending a bytes object to json raising "is not JSON serializable"

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/818


Miserlou:
Awesome

Issue #885, opened 2017-05-24
Automatic creation of __init__.py breaks import of Flask-User


marcoesposito1988:
## Context

 

I tried to use Zappa with my Flask app, which makes use of [Flask-User](https://github.com/lingthio/Flask-User).

## Expected Behavior

The app should have worked.

## Actual Behavior

The app fails with the following stack trace:
```
[1495617599312] File "/tmp/web/app/__init__.py", line 22, in <module>
[1495617599312] from flask_user import SQLAlchemyAdapter, UserManager, current_user
[1495617599312] File "/tmp/web/flask_user/__init__.py", line 13, in <module>
[1495617599312] from . import forms
[1495617599312] File "/tmp/web/flask_user/forms.py", line 19, in <module>
[1495617599332] from .translations import lazy_gettext as _
[1495617599332] ImportError: cannot import name 'lazy_gettext'
```

## Possible Fix

This is because the library has a "translations" package (folder without an `__init__.py`) containing folders each holding babel files, sitting next to a "translations.py" single file module. When the library is imported, the "from .translations import lazy_gettext" statement confuses the folder with the file due to the `__init__.py` erroneously created by Zappa during packaging.

This is analogous to #766, but the fix in #769 does not apply since the folder is not empty. 

If it is fine, I would work on a merge request adding a constraint that the `__init__.py` file should not be created when a file with the same name as the examined folder lies next to it.

Do you foresee any problem with this approach?

PR #888, opened 2017-05-25
Avoid creating __init__.py in directory with same name as module

marcoesposito1988:
## Description
 

As discussed in #885, if a library contains a folder (e.g. `lib/translations`) in the same folder as a module with the same name (e.g. `lib/translation.py`), creating an `__init__.py` file in it (e.g. `lib/translations/__init__.py`) can break relative imports (e.g. `from .translations import *`).

This patch inserts a check before creating the file to verify if in the parent directory there is a file named in the same way with extension .py or .pyc.

However, this is apparently not enough: the subdirectories should also not contain any init file, otherwise the directory is still considered a namespace package (if I remember correctly the terminology).

So I also modified the check introduced in #769 to look recursively for .py or .pyc files before creating an init file. If a directory does not contain .py or .pyc or any of its subfolders, it will not get the init file.

This seems to solve my issue with Flask-User (now I get another problem with Crypto which seems to be unrelated). Before proceeding with extensive testing, I would welcome any feedback on the approach.

Thanks in advance and also for your awesome job!

## GitHub Issues

Solves #885

### DIFF
--- a/tests/data/test1.py
+++ b/tests/data/test1.py
@@ -1,0 +1,6 @@
+
+# this is a module
+
+
+def greet():
+    print('There is more stupidity than hydrogen in the universe, and it has a longer shelf life.')
--- a/tests/data/test1/this_dir_is_not_a_module.txt
+++ b/tests/data/test1/this_dir_is_not_a_module.txt
@@ -1,0 +1,9 @@
+# test1 is a directory without any .py files in it, so a __init__.py file should NOT be created in it
+
+Information is not knowledge.
+Knowledge is not wisdom.
+Wisdom is not truth.
+Truth is not beauty.
+Beauty is not love.
+Love is not music.
+Music is THE BEST.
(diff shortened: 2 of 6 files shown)


Miserlou:
I'd like to merge this, but it looks like one test related to this change is failing. Can you update the PR to make sure the test is working properly?

marcoesposito1988:
Hi @Miserlou , thanks for the feedback!

Sure, but that will be hard since there are calls to os.path.exist and so on.

What about creating a directory with test data?

marcoesposito1988:
I added the tests with the required test data, and refactored those of `utilities.contains_python_files_or_subdirs()` to use it. I also refactored the flow to avoid scanning a directory twice.

I still have to test this by deploying a package, though.

marcoesposito1988:
So I tried this with a [minimal flask application ](http://flask.pocoo.org/docs/0.12/quickstart/) first, then I added flask-user, and it always worked reliably. I think it is ready for merging.

Miserlou:
And a nice test as well!

marcoesposito1988:
:metal:

Issue #244, opened 2016-08-12
Minimum AWS policies for example.


wesleykerr:
I’m looking to kick of a project with Zappa, but per company rules the default policies for AWS access are a bit too lax (allow * is forbidden), so can we get the minimum set of AWS policies to run the example provided?

Miserlou:
Yeah, this is a really good one that definitely needs addressing. I'm actually not super sure. It's very liberal out of the gate, but it absolutely needs a way to run in a minimum-permissions kinda way. Let's see.

For the execution policy, _I think_ this is all that's necessary:

ASSUME_POLICY = """{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Sid": "",
      "Effect": "Allow",
      "Principal": {
        "Service": [
          "apigateway.amazonaws.com",
          "lambda.amazonaws.com",
          "events.amazonaws.com"
        ]
      },
      "Action": "sts:AssumeRole"
    }
  ]
}"""

and for the attach policy:

ATTACH_POLICY = """{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                 "logs:CreateLogGroup",
                 "logs:CreateLogStream",
                 "logs:PutLogEvents",
                 "logs:DescribeLogStreams"
            ],
            "Resource": "arn:aws:logs:YOUR_LOG_RESOURCE_ARN"
        },
        {
            "Effect": "Allow",
            "Action": [
                "lambda:InvokeFunction"
            ],
            "Resource": [
                "arn:aws:lambda:YOUR_LAMBDA_FUNCTION_ARN"
            ]
        },
    ]
}""" 

Plus permissions for whatever AWS resources you're using on top of that.

Don't quote me on that, but that's my first guess. Leaving this ticket open until we can figure this out and get it documented.

Miserlou:
Related https://github.com/Miserlou/Zappa/issues/241

wesleykerr:
Thanks!  I'll start working with that and see where that can get me.

Miserlou:
Great! Please report back with what you find out so we can add it to the documentation.

wesleykerr:
This is what I've learned so far (note: I am not an AWS expert) from adding actions on errors while deploying.

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "iam:GetRole",
                "iam:PutRolePolicy"
            ],
            "Resource": [
                "*"
[...]

PR #894, opened 2017-05-26
first pass at iam policy example for deploy user

Bartvds:
This is basically a collection from #244

- Started from https://github.com/Miserlou/Zappa/issues/244#issuecomment-303697308 @audiolion @MihaZelnik
- Removed "s3:CreateMultipartUpload" because it is not needed
- Added "route53:ListResourceRecordSets" and "apigateway:OPTIONS" for zappa certify @aehlke 
- Added some SNS rules for events @bxm156
- Added some EC2 rules for security groups @eely22
- Added "lambda:InvokeFunction" for command @bartvds

I split it up into statements so people can remove chunks easily. The policy lingo is a bit limited and the only way to comment a Statement is a Sid value (with only CamelCase value).

Anyway, I'm by no means an expert on zappa or IAM and this is cargo culted so please review. 

Maybe this should be provided with some example (patterns) to make it more specific? Put in markdown instead of JSON?

### DIFF
--- a/example/policy/deploy.json
+++ b/example/policy/deploy.json
@@ -25,7 +25,7 @@
             ]
         },
         {
-            "Sid": "CoreLambda",
+            "Sid": "Core",
             "Effect": "Allow",
             "Action": [
                 "lambda:AddPermission",
@@ -46,7 +46,8 @@
                 "cloudformation:UpdateStack",
                 "logs:DescribeLogStreams",
                 "logs:DeleteLogGroup",
-                "logs:FilterLogEvents"
+                "logs:FilterLogEvents",
+                "s3:CreateBucket"
             ],
             "Resource": [
                 "*"
--- a/example/policy/deploy.json
+++ b/example/policy/deploy.json
@@ -1,0 +1,151 @@
+{
+    "Version": "2012-10-17",
+    "Statement": [
+        {
+            "Sid": "RoleManager",
+            "Effect": "Allow",
+            "Action": [
+                "iam:AttachRolePolicy",
+                "iam:CreateRole",
+                "iam:GetRole",
+                "iam:PutRolePolicy"
+            ],
+            "Resource": [
+                "*"
+            ]
+        },
+        {
+            "Sid": "RoleDelegation",
+            "Effect": "Allow",
+            "Action": [
+                "iam:PassRole"
+            ],
+            "Resource": [
+                "arn:aws:iam::<account_id>:role/*-ZappaLambdaExecutionRole"
+            ]
+        },
+        {
+            "Sid": "CoreLambda",
+            "Effect": "Allow",
+            "Action": [
+                "lambda:AddPermission",
+                "lambda:CreateFunction",
+                "lambda:DeleteFunction",
+                "lambda:GetFunction",
+                "lambda:GetPolicy",
+                "lambda:InvokeFunction",
+                "lambda:ListVersionsByFunction",
+                "lambda:RemovePermission",
+                "lambda:UpdateFunctionCode",
+                "lambda:UpdateFunctionConfiguration",
+                "cloudformation:CreateStack",
+                "cloudformation:DeleteStack",
+                "cloudformation:DescribeStackResource",
+                "cloudformation:DescribeStacks",
+                "cloudformation:ListStackResources",
+                "cloudformation:UpdateStack",
+                "logs:DescribeLogStreams",
+                "logs:DeleteLogGroup",
+                "logs:FilterLogEvents"
+            ],
+            "Resource": [
+                "*"
+            ]
+        },
+        {
+            "Sid": "CoreS3List",
+            "Effect": "Allow",
+            "Action": [
+                "s3:ListBucket"
+            ],
+            "Resource": [
+                "arn:aws:s3:::<s3_bucket from zappa_settings.json>"
+            ]
+        },
+        {
+            "Sid": "CoreS3Object",
+            "Effect": "Allow",
+            "Action": [
+                "s3:DeleteObject",
+                "s3:GetObject",
+                "s3:PutObject",
+                "s3:AbortMultipartUpload",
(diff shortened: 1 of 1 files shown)


jakul:
@Bartvds `CreateMultipartUpload` is needed for me!

```$ AWS_PROFILE=oldaccount-craig-deployment-example zappa deploy uatold
Calling deploy for stage uatold..
Warning! You may have a namespace collision with app.wsgi.the_app! You may want to rename that file.
Warning! Your project and virtualenv have the same name! You may want to re-create your venv with a new name, or explicitly define a 'project_name', as this may cause errors.
Downloading and installing dependencies..
 - cffi==1.9.1: Using locally cached manylinux wheel
 - cryptography==1.7.2: Warning! Using precompiled lambda package version 1.4 instead!
 - greenlet==0.4.12: Using locally cached manylinux wheel
 - pymongo==3.4.0: Using locally cached manylinux wheel
Packaging project as zip..
Uploading s-user-old-uatold-1496245451.zip (23.5MiB)..
  0%|                                                                    | 0.00/24.6M [00:00<?, ?B/s]Failed to upload s-user-old-uatold-1496245451.zip to zappa-uploads-uat-old/s-user-old-uatold-1496245451.zip: An error occurred (AccessDenied) when calling the CreateMultipartUpload operation: Access Denied

Error: Unable to upload to S3. Quitting.
```

Perhaps you didn't need it because your sample zip file is small and mine is large?

jakul:
+1 for the rest of the permissions though. I've copied/pasted these into my project and they work fine for all my usecases.

ebertti:
👍

levesquejf:
`lambda:GetFunctionConfiguration` is required to update an existing deployment.

2PacIsAlive:
huh, when I use this policy I get `An error occurred: Maximum policy size of 2048 bytes exceeded for user`

philvarner:
I'm interested in doing something about this within Zappa.  I've been thinking about these things:

1) Improve the documentation such that it's much more apparent that the permissions in the policy given to the default role are far broader than they strictly need to be, and that _most_ production deployments (not just those in a "corporate environment") will probably need to use manually-managed roles configured via 'role_name'. 

2) Get the above example policy committed so at least there are examples to work from that aren't just in github threads (like this one).

3) Precisely document which Zappa features require which permissions in the current policy and the example minimal policy, e.g., if you don't turn X-Ray on, you don't need the X-Ray permissions (that's a pretty obvious one, but many of the other ones aren't -- particularly the ENI ones). Part of this is separating the permissions required for Zappa (e.g., just no-op in the lambda) and which are granted so that most code that uses common AWS services (e.g., dynamodb) will just work without needing to explicitly grant 'extra_permissions'.

4) Convert the existing ASSUME_POLICY and ATTACH_POLICY to troposphere code. One of the primary benefits is simply being able to document exactly why a permission is being added to the policy, without needing to break the JSON string literal up into a bunch of concatenated strings and comments. 

5) As for making this more configurable _within Zappa_, I was thinking that there could be a set of policies instead of a single one (i.e., ATTACH_POLICY) that could be specified in the Zappa config. One option might be a parameter "canned_role" with options "all" (for the existing one, which is the default for backwards compatibility) or "minimal" (just to deploy a no-op handler), and then also add whatever additional permissions necessary for the features you've enabled, e.g., xray.  My feeling is that users should be required to add whatever permissions their code needs via 'extra_permissions' rather than adding the common ones (e.g., dynamodb) by default, since in most prod envs people aren't going to want to grant a single Lambda access to all their dynamo/sqs/s3/etc resources.  Another option could be to specify a function name to pass the troposphere Template object (from the previous item) into, and then you could write whatever code you needed to to modify the permissions without having to dump your 'extra_permissions' JSON into the zappa config file.

philvarner:
Created a PR for the documentation issue (#1 above) https://github.com/Miserlou/Zappa/pull/1233

philvarner:
One change that can be made to this policy:
instead of 
```
"Resource": [
                "arn:aws:iam::<account_id>:role/*-ZappaLambdaExecutionRole"
]
```
you can use Sub and the pseudo parameter for AccountID:

```
"Resource": [
                { "Fn::Sub" : "arn:aws:iam::${AWS::AccountId}:role/*-ZappaLambdaExecutionRole" }
]
```

philvarner:
the apigateway permissions can be reduced to "apigateway:*", as the only one that's missing is HEAD, and adding that doesn't seem to be a concern.

coltyharrison:
When using `zappa cerify`, I ran into this error:
```
botocore.errorfactory.BadRequestException: An error occurred (BadRequestException) when calling the CreateDomainName operation: Unable to associate certificate [CERTIFICATE ARN] with CloudFront. This error may prevent the domain name [DOMAIN NAME] from being used in API Gateway for up to 40 minutes. Please ensure the certificate domain name matches the requested domain name, and that this user has permission to call cloudfront:UpdateDistribution on '*' resources.
```

Can we add `cloudfront:UpdateDistribution` on * resources to this?

dpretty:
Thanks to everyone who's been working on this, it helps a lot!

One issue though is the Action "ec2:DescribeVpcsRequest" appears to be invalid, it's not listed here: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_Operations.html

I assume if this policy has been working for people, then this action can just be omitted.

Pinging #244 (comment https://github.com/Miserlou/Zappa/issues/244#issuecomment-270270950) as that appears to be the original source of "ec2:DescribeVpcsRequest"

yiannistri:
Thanks for all the contributions so far. I've had to set up a VPC recently so I had to add the following permissions to my policy.

```
{
    "Effect": "Allow",
    "Action": [
        "ec2:CreateNetworkInterface",
        "ec2:DeleteNetworkInterface",
        "ec2:DescribeNetworkInterfaces"
    ],
    "Resource":"*"
}
```

Issue #891, opened 2017-05-26
Add --disable-progress command line parameter


ryansmith940:
We use an automated deployment system for running `zappa deploy|update`.  The progress bar in these commands is written to stderr.  Our deployment system notices that something was written to stderr, and flags the deployment with a warning.  To get rid of this warning, I would like to run the Zappa commands with a flag `--disable-progress` so that this warning does not appear.

Miserlou:
Please change to --no-progress, or you can use the already-present --silent.

audiolion:
This sort of exists already, `--quiet` flag will silence output, however, to do progress bars `Zappa` uses [tqdm](https://github.com/tqdm/tqdm) which prints to `sys.stderr` by default. You can change the print location by passing `file=...` to the `tdm()` fn.

You can search for [places in the codebase](https://github.com/Miserlou/Zappa/blob/40e018729e87d06380cb8a8821cec8a16173a04c/zappa/zappa.py#L299) that use tqdm. Basically pass an argument flag to change how the progress bar prints (through `file=..` param) or update the `--quiet` command to disable the `progress.update` callbacks.

PR #895, opened 2017-05-26
Add --disable_progress command line parameter

ryansmith940:
Was needed to prevent zappa commands from writing to stderr.

https://github.com/Miserlou/Zappa/issues/891


audiolion:
You can check https://github.com/tqdm/tqdm/blob/9d7ccc3dfa1a7c37c8e9c48300f36e3466ee4342/tqdm/tests/tests_tqdm.py#L698 for examples of testing disable

Issue #899, opened 2017-05-28
Zappa package command spends lots of time downloading manylinux wheel metadata from Pypi


suriya:
The `zappa package` command is slower than ideal. One reason is that Zappa checks whether a package has a manylinux wheel by downloading the package's Python metadata from PyPI. When a project's virtualenv has a lot of packages, this can consume a lot of time.

## Possible Fix
It would be good to cache the metadata files and reuse them in the next invocation.

## Steps to Reproduce
Have a lot of packages in your virtual environment and run `zappa package`

Miserlou:
Fixed via merge, thank you.

PR #900, opened 2017-05-28
Cache and reuse PyPI metadata of packages

suriya:
This commit is a fix for https://github.com/Miserlou/Zappa/issues/899

The zappa package command is slower than ideal. One reason is that Zappa
checks whether a package has a manylinux wheel by downloading the package's
Python metadata from PyPI. When a project's virtualenv has a lot of
packages, this can consume a lot of time.

This commit introduces a cache for PyPI metadata and uses them in future
invocations.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/899

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -684,16 +684,36 @@ class Zappa(object):
 
         Related: https://github.com/Miserlou/Zappa/issues/398
         Examples here: https://gist.github.com/perrygeo/9545f94eaddec18a65fd7b56880adbae
+
+        This function downloads metadata JSON of `package_name` from Pypi
+        and examines if the package has a manylinux wheel. This function
+        also caches the JSON file so that we don't have to poll Pypi
+        everytime.
         """
-        url = 'https://pypi.python.org/pypi/{}/json'.format(package_name)
-        try:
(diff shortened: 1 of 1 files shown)


Miserlou:
Okay, let's do it! The road to hell is paved with cached invalidation bugs, but I don't see anything wrong with this code. I'm going to hold you responsible if somebody finds that it's wrong though!

suriya:
Thanks for merging.

Does my pull request work in both Python 2 and 3? I did not have the time to make my changes work in both (that's why the test fails). Please make sure it works in both versions.

I have been using it for a few months and the code's working fine in Python 2. I will take responsibility to fix this (in Python 2) if someone complains about broken functionality.

suriya:
In fact, I made the second commit in this pull request https://github.com/Miserlou/Zappa/pull/900/commits to address compatibility with both 2 and 3, but could not succeed.

Please fix it for both versions :)

Miserlou:
Ah, shit, you're right. How did I miss that. Reverting.

suriya:
It's probably 5 minutes effort (for someone who knows how to do it) to make it work in both. I hope you can give it a shot.

Issue #930, opened 2017-06-14
Format and colorize the output of manage and invoke commands


jonatasbaldin:
Hi!

The default output from manage and invoke is just a base64 decoded string, which is really hard to read. I made #931  to trying fix that.

jonatasbaldin:
Merged!

PR #931, opened 2017-06-14
Format and colorize the output of manage and invoke commands

jonatasbaldin:
Hi! This is a solution for #930 

This PR formats correctly the line breaks and prints a colored version of the output. I made this **the default** option, if you want the old version, you must use the `--no-color` parameter to `invoke` or `manage`. This is totally open for discussion, I just made this way because I though that this was nicer.

Here's some screenshots of before and after the PR:

Before (now with a `--no-color` parameter):
Before

After:
After

What do you think about it? Any feedback is appreciated!

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -776,6 +776,30 @@ class TestZappa(unittest.TestCase):
         zappa_cli.print_logs(logs, colorize=True, non_http=False, http=False)
         zappa_cli.check_for_update()
 
+    def test_cli_format_invoke_command(self):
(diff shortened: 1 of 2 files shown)


jonatasbaldin:
hey @Miserlou, could u give a look at this? 😄

Miserlou:
Hey, sorry for the hold up. My concern here is around Py2/3 compatibility, have you tried this patch with  both versions?

jonatasbaldin:
Hey @Miserlou, sorry for the delay too! Tested it is looking good!

Issue #909, opened 2017-06-02
Package command should not require aws cli config/credentials


brianrower:
I've been integrating zappa into an AWS Codebuild/Codepipeline. When running on a codebuild instance, zappa fails with the stack trace shown below. Clearly it's failing initializing boto due to the lack of aws config and credentials files. But when doing a package command, I do not believe any AWS resources are utilized, hence this command should work without the credentials.

### Steps to reproduce locally (without codebuild):
1. cd into a directory with a working zappa build
2. activate the virtual env
3. Run "mv ~/.aws ~/.aws.bak" (so that boto cannot find config and credentials)
3. run "zappa package <env> -o lam.zip"

### Environment:
Ubuntu 16.04 and 14.04
Python 3.5.2
zappa 0.42.0

### Workaround:
Create a  config and credential files in a valid format, but invalid crentials like this:
config:
```
[default]
output = json
region = us-west-2
```

credentials:
```
[default]
aws_access_key_id = XXXXXXXXXXXXXXXXXXXX
aws_secret_access_key = XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```
(note: those are ACTUALLY X's in the file, I'm not trying to hide valid credentials)
and point the environment variables AWS_CONFIG_FILE and AWS_SHARED_CREDENTIALS_FILE to these files. Since it's just an issue with initializing boto and no aws resources are accessed, the zappa package command then succeeds

Stack Trace:
```
Traceback (most recent call last):
File "/codebuild/output/src623405482/src/.python/lib/python3.5/site-packages/zappa/cli.py", line 2353, in handle
sys.exit(cli.handle())
File "/codebuild/output/src623405482/src/.python/lib/python3.5/site-packages/zappa/cli.py", line 456, in handle
self.dispatch_command(self.command, stage)
File "/codebuild/output/src623405482/src/.python/lib/python3.5/site-packages/zappa/cli.py", line 482, in dispatch_command
self.load_settings(self.vargs.get('settings_file'))
File "/codebuild/output/src623405482/src/.python/lib/python3.5/site-packages/zappa/cli.py", line 1801, in load_settings
runtime=self.runtime
File "/codebuild/output/src623405482/src/.python/lib/python3.5/site-packages/zappa/core.py", line 248, in __init__
self.load_credentials(boto_session, profile_name)
[...]

PR #935, opened 2017-06-17
Addressing #909: Don't load credentials for 'package' command

bcongdon:
## Description
 

- Changes the `zappa package` command so it doesn't load AWS credentials.
- Adds test that `zappa package` doesn't load credentials.

## GitHub Issues

#909


Miserlou:
Nice small and clean solution.

Issue #765, opened 2017-04-02
lambda environment variables are removing after zappa update


anush0247:
## Context
I am using environment variables in lambda configuration, when tried to `zappa update` the environment variables got vanished

## Possible Fix

before updating the lambda configuration get_function_configuration, retain the same values if they not changed by zappa configuration

fixed in #663

Miserlou:
Bumping this issue for discussion around what the behavior here should be.

By default, Zappa sets the environment variables based on the stage config.

Should it check if those values have changed, and if so, not update them? Should it only set them on `deploy`? Should it ask? Should it be configurable with a flag? Etc.

nikbora:
AWS environment variables are being cleared since Zappa 0.41.2. It didn't happen in 0.41.1. To be clear we are talking about env vars on Lambda side (i.e. ones setup via AWS Lambda console or AWS CLI).

The reason for this is this change here (as part of the KMS PR): https://github.com/Miserlou/Zappa/pull/600/files#diff-5e6a43a7125a90e37a105fa7924098c8R761

Since we don't use any Zappa settings based env vars but rely on AWS ones it is quite a PITA.

I would like to do the following fix:
1. Before update we download the current AWS env var dictionary.
2. We merge the settings based env vars into those (so anything you have setup locally would overwrite any remote ones on from AWS).
3. Then we upload that.

nikbora:
On Slack there was discussion to splinter environment_variables and aws_environment_variables and only set aws_environment_variables for the key that is present (which means aws env vars one sets via CLI or in AWS console won't just be wiped on each deploy). @Miserlou if this still makes sense I will work on a PR for this.

Miserlou:
That's the solution that I'd like to see, I think.

PR #962, opened 2017-06-26
Separating out native AWS environment variables

nikbora:
## Description

AWS environment vars are now splintered into a separate configuration item.

This addresses https://github.com/Miserlou/Zappa/issues/765 and prevents any AWS environment vars that are set via the AWS console or cli from being wiped if their key name doesn't exist the settings file. Original support for native AWS env vars was added in https://github.com/Miserlou/Zappa/pull/600.

Tested on python 2.7 and python 3.6

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Zappa Changelog
 
+## 0.43.0
+* Add exclude for __pycache__ contents (#943)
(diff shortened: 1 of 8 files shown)


nikbora:
Hey @Miserlou, let me know if you have any thoughts/comments on this whenever you get the chance.

Issue #965, opened 2017-06-27
Current AWS Boto version doesn't support S3 v4 Signatures causing failing remote env


anateus:
## Context
I'm not certain under what conditions S3 requires v4 signatures, but that's the direction things are moving. Zappa currently uses the AWS-provided boto and botocore packages which default to using HMACv1 signatures. There appears to be no way to instruct those packages to use s3v4 as the signature method.

## Expected Behavior
Uploading a JSON file to s3 and setting its path to the `remote_env` field in `zappa_settings.json` should result in the env vars being loaded as if they were provided in the `environment_variables` field.

## Actual Behavior
The app deploys fine, but then fails to load. In the logs we can see that accessing the remote json file fails due to signature error. The presence of `Calculating signature using hmacv1 auth.` in the logs indicates it's not using the v4 signature method.

## Possible Fix

The `botocore` changlog indicates that starting from version 1.5.71, the v4 signature method becomes the default. AWS provides older versions for both Python 2.7 and 3.6.

Was able to fix this by manually modifying `zappa/cli.py` to remove `boto3` and `botocore` from the list of excluded packages and having the latest versions of both in the virtual environment when the app gets deployed.

## Steps to Reproduce

As mentioned above, v4 seems to be the default signature method now, but some regions may allow older ones as well. If the following steps aren't reproducible this may be why.
1. Place a `remote_env.json` file on S3
2. Add it `zappa_settings.json`
3. Deploy the app
4. Attempt to load the app

## Your Environment

* Zappa version used: 0.42.2
* Operating System and Python version: OS X 11.6 Python 3.6.0
* The output of `pip freeze`:
```
alembic==0.9.2
argcomplete==1.8.2
base58==0.2.4
bcrypt==3.1.3
boto3==1.4.4
botocore==1.5.75
certifi==2017.4.17
cffi==1.10.0
chardet==3.0.4
click==6.7
dataset==0.8.0
docutils==0.13.1
durationpy==0.4
Flask==0.12
Flask-Caching==1.2.0
Flask-Migrate==2.0.3
Flask-S3==0.3.3
Flask-Script==2.0.5
Flask-Session==0.3.1
Flask-SQLAlchemy==2.2
future==0.16.0
futures==3.1.1
gunicorn==19.7.0
hjson==2.0.7
idna==2.5
itsdangerous==0.24
Jinja2==2.9.6
jmespath==0.9.3
kappa==0.6.0
lambda-packages==0.15.1
[...]

PR #966, opened 2017-06-27
Added override_default_exclude config option

anateus:
## Description
 
Added a new config parameter `override_default_exclude ` that allows overriding the default excludes, as per the discussion in #965 . The main use is to allow using versions of boto newer than what Amazon provides, but could have additional utility.

## GitHub Issues

#965

### DIFF
--- a/README.md
+++ b/README.md
@@ -599,6 +599,7 @@ to change Zappa's behavior. Use these at your own risk!
         ],
         "exception_handler": "your_module.report_exception", // function that will be invoked in case Zappa sees an unhandled exception raised from your code
         "exclude": ["*.gz", "*.rar"], // A list of regex patterns to exclude from the archive. To exclude boto3 and botocore (available in an older version on Lambda), add "boto3*" and "botocore*".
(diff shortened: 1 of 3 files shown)


Miserlou:
Please change to `excludes_default` and make sure the README stays alphabetized. Looks good otherwise.

anateus:
@Miserlou I know you suggested `excludes_default` in the issue but that name on its own is less directly meaningful than `override_default_exclude`, which only depends on knowing what `exclude` is. Presumably `excludes_default` is a contraction of **excludes default [exclude items]**. Additionally, most other configuration options that are verb phrases are imperative (exclude, include, keep warm), so *override* parallels that pattern.

I'll make the change if still prefer it.

jneves:
@anateus because the overrides would result in extra support (people not finding it easily) we would prefer the change to `excludes_default` please.

mathom:
Is this still an issue? My testing shows that Zappa now packages the version of boto3 you have installed by default.

PR #973, opened 2017-06-29
Update to readme about how to add your zappa stored environment varia…

andytwoods:
…bles to your locally running django app

### DIFF
--- a/README.md
+++ b/README.md
@@ -809,6 +809,20 @@ If your project needs to be aware of the type of environment you're deployed to,
 
 If you are using KMS-encrypted AWS envrionment variables, you can set your KMS Key ARN in the `aws_kms_key_arn` setting.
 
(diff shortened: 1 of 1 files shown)


Miserlou:
This is a nice idea, but it's Windows specific, and I think the majority of our users are on OSX and Linux. Is there are a cross platform alternative to `RUNNING_DEVSERVER = sys.platform.startswith('win')`?

Miserlou:
In fact, I think we can completely de-Djangoize this and have the condition simply be:

```python
if 'SERVERTYPE' in os.environ and os.environ['SERVERTYPE'] == 'AWS Lambda':
```

andytwoods:
Super nice! Afraid I'm rather a newbie when it comes to contributing to other projects so I hope my git actions suffice :)

PR #983, opened 2017-07-03
Added the ability to pass in args and kwargs to the `zappa invoke`

xuru:
## Description
 
Added the ability to pass in arguments to invoked commands from the command line.  For example:

`$ zappa invoke production 'my_app.my_function' one two three=3 name=bob`

Which would get passed to your function in the `event['args']` and `event['kwargs']` dictionary.  For example:

```
def myfunc(event, context):
    print("Args: {}".format(event['args']))
    print("Kwargs: {}".format(event['kwargs']))
```

Would print out

```
Args: ['one', 'two']
Kwargs: {'three': '3', 'name': 'bob'}
```

## GitHub Issues

Couldn't find a ticket for this, but @Miserlou said in slack:
 
> this was assigned to somebody as their school project but they never did it

### DIFF
--- a/README.md
+++ b/README.md
@@ -591,7 +591,7 @@ All of this code is still backwards-compatible with non-Lambda environments - it
 
 ### Running Tasks in a VPC
 
-If you're running Zappa in a VPC you'll need to configure your subnets to allow your lambda to communicate with services inside your VPC as well as the public Internet. A minimal setup requires two subnets.
+If you're running Zappa in a Virtual Private Cloud (VPC), you'll need to configure your subnets to allow your lambda to communicate with services inside your VPC as well as the public Internet. A minimal setup requires two subnets.
 
 In __subnet-a__:
 * Create a NAT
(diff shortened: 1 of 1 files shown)


Miserlou:
The build is not happy!

zappa/handler.py
```diff
@@ -336,9 +336,27 @@ def handler(self, event, context):
 
         # This is a direct command invocation.
         elif event.get('command', None):
+            if not event['command']:
```
pjz: This check is entirely redundant with the previous line.

xuru: Line 338 checks for the existence of the key/value of 'command in the dictionary, but doesn't check if that value is a "false" value (None, empty string, etc).

pjz: `event.get('command')` is the same as `event['command']`, so it is a check for falsity, with a side effect of returning `None` instead of throwing `KeyError`

xuru: Ah, you are correct, sorry, got my mind on other things...

Miserlou:
I'd still like to see this merged! @mizell - can you review and test it for your purposes?

smizell:
@Miserlou I'll test it out either today or tomorrow. This is great!

zappa/handler.py
```diff
@@ -336,9 +336,24 @@ def handler(self, event, context):
 
         # This is a direct command invocation.
         elif event.get('command', None):
-
             whole_function = event['command']
+            args = []
+            kwargs = {}
```
smizell: I think I'd recommend not going the non-standard way of parsing command line keyword arguments. For most command lines, I'd expect something like `zappa invoke "myapp.myfunc" arg1 arg2 --arg3 foo` to provide args as `['arg1', 'arg2']` and kwargs as `{'arg3': 'foo'}`.

Of course this is up to you @xuru and @Miserlou, but I think leaving out kwargs for the first iteration would be best.

xuru: Sorry, i've been pretty busy and couldn't get back to this.  I do have a ticket this sprint to get back to this, so I can help out tomorrow or this weekend.  That said, here is where I left off:  I wasn't able to get this working, and I don't really remember what the issue was.  I'll take a look and let you know when I can.

smizell: @xuru no worries! And thank you for working on this. I'm going to tinker with it probably tomorrow as well.

pjz:
I'd think it needs to be opened as a bug, not re-opened the original feature creation.

xoelop:
I'm having the same problem, can't send args or kwargs to functions using `zappa invoke`. Any progress on this? Thanks!!

Giribhushan:
I also have same problem. Did anyone fix this ?

Issue #984, opened 2017-07-03
Use AWS Credentials and Region When Creating a new Boto Session to Fetch Project Zip


bahoo:
I ran into an issue where [fetching the project zip from S3](https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L165) was failing, with a very unhelpful `400 Bad Request` from S3, and no response in the body to illuminate on the problems.

## Context
I'm using the `us-west-2` region, with `slim_handler` set to `true`.

## Expected Behavior
The handler should download the zip successfully from S3.

## Actual Behavior
The handler tried to download my project zip from S3, and it failed.

## Possible Fix

I eventually discovered it was because of the [`boto`session being initiated](https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L155-L156) above it, without the right kwargs being passed in. Namely, in my case, I needed these set:

- `aws_access_key_id`
- `aws_secret_access_key`
- `region_name`

I *think* `region_name` was the clincher for me (I'm using `us-west-2`), as the access and secret key should be part of the environment variables. But interestingly, if `aws_session_token` was passed in, it would still fail. That was puzzling to me, but I didn't have the bandwidth to unpack further.

I've made these changes, sitting in a branch here: https://github.com/bahoo/Zappa/tree/use_aws_credentials_in_creating_boto_session Happy to add a PR if that seems like a good idea to others.

## Steps to Reproduce
1. Set your AWS region to `us-west-2`
2. Use `slim_handler: true`
3. Deploy

## Your Environment
* Zappa version used: 0.42.0
* Operating System and Python version: macOS Sierra, Python 2.7.12
* The output of `pip freeze`: 
```
appdirs==1.4.3
argcomplete==1.8.2
base58==0.2.4
boto3==1.4.4
botocore==1.5.40
click==6.7
dj-database-url==0.4.2
Django==1.10.6
django-bootstrap3==8.2.1
django-dotenv==1.4.1
docutils==0.13.1
durationpy==0.4
EasyProcess==0.2.3
future==0.16.0
futures==3.1.1
googlemaps==2.4.6
gunicorn==19.7.1
hjson==2.0.7
jmespath==0.9.3
kappa==0.6.0
lambda-packages==0.16.1
packaging==16.8
placebo==0.8.1
psycopg2==2.7.1
pyparsing==2.2.0
python-dateutil==2.6.0
python-slugify==1.2.4
pyunpack==0.1.2
PyYAML==3.12
[...]

PR #986, opened 2017-07-03
Add AWS credentials / region to boto session instantiation when fetch…

bahoo:
## Description
Specifies AWS credentials and region when fetching the project zip via S3.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/984


pjz:
I'm :-1:  on this since Boto implements this mechanism already.  If it's not working, we should figure out what's breaking the environment and/or file a bug with Boto.

bahoo:
Per discussions in #984 this is an edge case I'm running into and coding around. I'm going to close it.

Issue #987, opened 2017-07-04
KeyError when there are no remote aws lambda env vars


yuric:
## Context
PR https://github.com/Miserlou/Zappa/pull/962 was merged into master but it raises KeyError when there are no remote lambda env variables etc or the (["Environment"]) key. My proposed change checks that that the key ["Environment"] exists. 

## Expected Behavior
It should handle the the merging of aws_environement_variables even if the are no env vars in the remote server and the Environment key is missing.

## Actual Behavior
Right it raises a `KeyError` when there are no remote lambda env variables etc and the Environment key is missing.

## Possible Fix
Check for that key (["Environment"]) exists.

## Steps to Reproduce
1. Try `zappa update {env}` where there are no ["Environment"] settings key.

PR #988, opened 2017-07-04
Checks for the key 'Environment' when fetching remote lambda env vars

yuric:
Prevents Zappa from failing to update when the remote lambda function returns no Environment Variables and is otherwise missing the 'Environment' key.

## Description
https://github.com/Miserlou/Zappa/issues/987

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/987
https://github.com/Miserlou/Zappa/pull/962
https://github.com/Miserlou/Zappa/issues/765


yuric:
Not sure how I decreased coverage by 16% with small changes on those 2 files. I am new enough to python and may be missing something obvious. Feel free to point that out. 

I will review this again shortly.

Miserlou:
This is a really nasty bug, but unfortunately it looks like your PR is causing a build failure. Can you try to fix so I can merge?

zappa/core.py
```diff
@@ -918,15 +918,17 @@ def update_lambda_configuration(    self,
...
-            for key, value in lambda_aws_environment_variables.items():
-                if key not in aws_environment_variables:
-                    aws_environment_variables[key] = value
+            # There may not be an Environment key defined.
+            # https://github.com/Miserlou/Zappa/issues/987
+            lambda_aws_environment_variables = self.lambda_client.get_function_configuration(FunctionName=function_name)
+            if "Environment" in lambda_aws_environment_variables:
+                # Get any remote aws lambda env vars so they don't get trashed
+                # Related: https://github.com/Miserlou/Zappa/issues/765
+                lambda_aws_environment_variables["Environment"].get("Variables", {})
```
BrianRower-BR: I think you want to add a "lambda_aws_environment_variables = " to the front of this line, that makes the unit tests pass

collingreen: and probably rename the lambda_aws_environment_variables on line 923 since it is now all of the function config, not just the environment

yuric: good call @BrianRower-BR and @collingreen. I'll update this shortly.

yuric:
I also added a couple test and placebo files hoping to increase test coverage instead of decreasing it. Aside from the tests a2ef4ee was just a cleanup commit on [zappa/core](https://github.com/Miserlou/Zappa/commit/a2ef4ee06628f4a087bd48d6ceab94dfb821d51d#diff-69e19f3663afb330594add0cc19b0a1c) as you can see.

zappa/core.py
pjz: would `lambda_aws_environment_variables.get("Environment", {}).get("Variables", {})` work?

Miserlou:
Let's do it!

nikbora:
Argh, the number of AWS API edge cases is astounding. Thanks @yuric for fixing this, I totally should have caught this initially but was always testing in envs that had some vars it seems.

Issue #992, opened 2017-07-05
Non-verbose error when updating stage from new environment


texonidas:
Issue is being logged for the sake of a pull request. Error logging on failure to get lambda function URL is non-specific.

## Context
When attempting to update an existing stage from a new environment (i.e. moving to continuous integration), a generic error message is logged "Warning! Couldn't get function <function> in <region> - have you deployed yet?"

Upon digging, I discovered that the real issue was that I had forgotten to set up appropriate AWS credentials in the CI environment.

## Expected Behavior
An error message specific to the reason my `zappa update <stage>` had failed.

## Actual Behavior
The generic message "Warning! Couldn't get function <function> in <region> - have you deployed yet?" was displayed (probably covers 99.9% of use cases).

## Possible Fix
Pull request incoming with fix - adding an extra Exception clause.

## Steps to Reproduce
1. Deploy a trivial application.
2. Set up a new environment without AWS credentials.
3. Call `zappa update <stage>`.

PR #993, opened 2017-07-05
Fixes Miserlou/Zappa#992 - except BotoCoreError before general exception in zappa.cli.update

texonidas:
## Description
Except a BotoCoreError instead of a generic Exception for more verbose logging in `zappa.cli.update`.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/992


Miserlou:
I'm not 100% that this is 2/3 compatible but it's better than nothing?

Issue #845, opened 2017-05-11
ImportError: No module named six.moves.http_client


jonathanglima:
I'm migrating a Django application from EC2 to Lambda and I'm configuring it to run with Zappa.

This application uses elasticsearch and I got the error below. All my packages are up to date and I'm running a python 2.7 venv using virtualenvwrapper.

I reached  kennethreitz/requests#3985 and even with a full package update, I still had the error.

I also added "delete_local_zip" to my zappa_settings so I could take a look inside the generated zip and check if the file isn't really there. I wasn't.

> [1494523900501] No module named six.moves.http_client: ImportError
> Traceback (most recent call last):
>   File "/var/task/handler.py", line 514, in lambda_handler
>   return LambdaHandler.lambda_handler(event, context)
>   File "/var/task/handler.py", line 259, in lambda_handler
>   handler = cls()
>   File "/var/task/handler.py", line 162, in __init__
>   wsgi_app_function = get_django_wsgi(self.settings.DJANGO_SETTINGS)
>   File "/var/task/django_zappa_app.py", line 20, in get_django_wsgi
>   return get_wsgi_application()
>   File "/private/var/folders/vv/9gq0jj317ql4dsjhr532k2z40000gq/T/pip-build-erQj35/Django/django/core/wsgi.py", line 13, in get_wsgi_application
>   File "/private/var/folders/vv/9gq0jj317ql4dsjhr532k2z40000gq/T/pip-build-erQj35/Django/django/__init__.py", line 27, in setup
>   File "/private/var/folders/vv/9gq0jj317ql4dsjhr532k2z40000gq/T/pip-build-erQj35/Django/django/apps/registry.py", line 116, in populate
>   File "/private/var/folders/vv/9gq0jj317ql4dsjhr532k2z40000gq/T/pip-build-erQj35/Django/django/contrib/admin/apps.py", line 23, in ready
>   File "/private/var/folders/vv/9gq0jj317ql4dsjhr532k2z40000gq/T/pip-build-erQj35/Django/django/contrib/admin/__init__.py", line 26, in autodiscover
>   File "/private/var/folders/vv/9gq0jj317ql4dsjhr532k2z40000gq/T/pip-build-erQj35/Django/django/utils/module_loading.py", line 50, in autodiscover_modules
>   File "/usr/lib64/python2.7/importlib/__init__.py", line 37, in import_module
>   __import__(name)
[...]

PR #994, opened 2017-07-05
fix/Include six

jakul:
## Description
This PR fixes https://github.com/Miserlou/Zappa/issues/845 by including `six` in the uploaded Lambda. The original problem is that deploying on Python 3.6 gives the following error `ImportError: No module named six.moves.http_client`

Miserlou said that the inclusion of six was the problem in https://github.com/requests/requests/issues/3985#issuecomment-296244685

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/845

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -2042,7 +2042,6 @@ class ZappaCLI(object):
                                         "dateutil",
                                         "botocore",
                                         "s3transfer",
-                                        "six.py",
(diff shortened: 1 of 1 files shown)


Miserlou:
But you've only changed the code if `if sys.version_info[0] < 3:`..?

jakul:
Oops.

PR #1005, opened 2017-07-11
adding warning when timeout is greater than 300

doron2402:


Miserlou:
This is a good thought, but since we'll need this for pretty much all of the commands and not just `update`, I'd say put this into `load_settings` directly when the setting is loaded rather than having a single use function that'd have to be added to all the primary command calls.

doron2402:
agree @Miserlou 
I'll work on a better pr.

Thanks for the quick response

doron2402:
- [ ] validate `timeout_seconds` (min 1 max 300)
- [ ] valid `log_level` 

@Miserlou should we validate all the settings? seems too much, but definitely going to help debug your deployment script?

Miserlou:
We should be doing more validation where we can, but the errors that cascade are generally self-explanatory. The ones like: https://github.com/Miserlou/Zappa/issues/78 are more tricker and more time consuming but I'd really like to be able to handle those.

doron2402:
yeah I was thinking about it making sure the file exists and the handler

doron2402:
@Miserlou Should I validate all the settings? if so, how about breaking into a separate file that validates the settings?

Miserlou:
Validate as much as you can, I suppose, but I don't think it needs to be a separate file, it can just be a function called from load_settings.

Issue #82, opened 2016-05-11
Zappa CLI Python3 support


kienpham2000:
I'm having this error when running zappa master on Python 3.5.1:
```
env/lib/python3.5/site-packages/zappa/cli.py", line 452
    os.chmod(temp_settings.name, 0644)
SyntaxError: invalid token

```

When I switch back to Python 2.7, no problem.

Also looks like the pypi version (0.17.6) is out of date. Can you please start tagging release?

Miserlou:
Yeah, I think dropping any notion of 3 support is probably the best move. It was a nice idea but it's too difficult to maintain the two in parallel given that Lambda still requires 2.7.

kienpham2000:
cool makes sense, I'll use 2.7.

PR #1007, opened 2017-07-16
Remove Python 3 references from documentation

hosay:
I encountered multiple errors when trying to use Python 3 and Zappa CLI, According to this post by Miserlou there is no support for it:
https://github.com/Miserlou/Zappa/issues/82

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ __Awesome!__
 
 ## Installation and Configuration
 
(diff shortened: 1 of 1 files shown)


hosay:
I was using an old version of Zappa (got it from pip last week weirdly), I haven't confirmed python 3 is not working with the latest. Closing for now.

Issue #1020, opened 2017-07-27
Delete <myproject>.zip from /tmp after load to free space (slim_handler=true)


olirice:
## Context

 

Currently, when using slim_handler=true, the project source zip file at `/tmp/my_project.zip` is not deleted after it is extracted.

Projects that download additional dependencies (tensorflow/sklearn models, binaries, video, etc) at runtime can burn through the available 500 mb pretty quickly and throw OSError resource exhausted errors.

What are your thoughts on deleting the project zip file after it's been decompressed? 

I'll submit a PR if approved.

Miserlou:
Fantastic idea

olirice:
I changed a bit more than originally proposed.

In case it isn't clear why I didn't just `os.remove('/tmp/project.zip')` at the end of `zappa.handler.load_remote_project_zip` please see below

tl;dr
> The punchline is that you can always deploy your resource files with `zappa deploy` rather than downloading them from s3 in application code.

**Scenario - Deployment source is ~60MB and resources (models, etc) are ~300MB** 
Lets say, the resulting zip file with source and resources came to 350MB

Currently, the deployment would fail because slim_handler tries to store project.zip (350MB) and the unzipped folder (360MB) in `/tmp` (500MB) at the same time. Bolting `os.remove('/tmp/project.zip')` onto the end of `load_remote_project_zip` doesn't help because the failure occurs before we get that far.

The most common way around that issue is to copy the large resource files to s3 independently and write a bunch of boilerplate in your application to put them in the right place when you have a cold start.

As written, #1022 tries to store `project.zip` in-memory (failing over onto `/tmp`). In that case, if the lambda instance has enough memory, you can deploy any zip file that decompresses into < 500MB. The punchline is that you can always deploy your resource files with `zappa deploy` rather than downloading them from s3 in application code.

Additionally, you know in advance if your deployment is going to fit in Lambda. If your local project folder is < 500MB and the instance's memory size > ~530MB, its going to fit.

Issue #961, opened 2017-06-25
OSError: [Errno 28] No space left on this device on slim_handler=true


abhisuri97:
## Context

When uploading my zipped lambda package with zappa, I get a 'no space left on this device' error after calling `zappa tail`

This is a fairly large project, so it is totally reasonable that Lambda might not be the best fit for it. I need tensorflow dependencies as well as a fairly large model. 

My total unzipped files are 498 MB (very close to 512 MB).

When zappa zips up everything it comes out to 213 MB for the project zip (unzipped is 492 MB) and 9.2 MB for the handler zip (unzipped is 47 MB). 

While I know that 492 is cutting it very close to the 512 /tmp limit, it doesn't go over it...unless the 47 MB from the handler is counted in that as well...

## Expected Behavior

Should be able to upload files in their entirity 

## Actual Behavior

I get this error when visiting the lambda URL:

```
"{'message': 'An uncaught exception happened while servicing this request. You can investigate this with the `zappa tail` command.', 'traceback': ['Traceback (most recent call last):\\n', '  File \"/var/task/handler.py\", line 433, in handler\\n    response = Response.from_app(self.wsgi_app, environ)\\n', '  File \"/var/task/werkzeug/wrappers.py\", line 903, in from_app\\n    return cls(*_run_wsgi_app(app, environ, buffered))\\n', '  File \"/var/task/werkzeug/wrappers.py\", line 57, in _run_wsgi_app\\n    return _run_wsgi_app(*args)\\n', '  File \"/var/task/werkzeug/test.py\", line 884, in run_wsgi_app\\n    app_rv = app(environ, start_response)\\n', \"TypeError: 'NoneType' object is not callable\\n\"]}"
```

and when looking through the tail log i get the OSError in the title.

## Steps to Reproduce

1. unzip the directory i link below, cd into it, activate the virtual environment (venv) and deploy with zappa.

## Your Environment

* Zappa version used: `0.42.2`
* Operating System and Python version: Mac OS 10.12.5, python3.6
* The output of `pip freeze`: 
```
argcomplete==1.8.2
backports.weakref==1.0rc1
base58==0.2.4
bleach==1.5.0
boto3==1.4.4
botocore==1.5.40
certifi==2017.4.17
chardet==3.0.4
click==6.7
docutils==0.13.1
durationpy==0.4
Flask==0.12.2
future==0.16.0
futures==3.1.1
hjson==2.0.7
html5lib==0.9999999
idna==2.5
[...]

Issue #881, opened 2017-05-23
Zipped and unzipped deployment adding upto more than 500Mb and failing deployment


neo01124:
Using "slim_handler": true, the zipped version of my deployment is 120Mb and the unzipped version is 436Mb. The combined size is more than the lambda limit for ephemeral storage which fails the deployment.

A streaming mode for unzipping the file would mitigate this.

mcrowson:
So this is a little tricker than I thought. Because there are multiple files in a zip, we can't exactly stream unzip (from what I can tell reading online). Would the gzip format lend itself better?

If so we run into another challenge implementing this. The current approach for creating the application zip with the slim_handler mirrors the application zip needed by lambda. Lambda requires a zip file. So we'd have to separate functionality there assuming we go gzip with the slim_handler application.

mcrowson:
Putting Slack convo w/ @mbeacom here: The commit above brings the whole response into memory.

bharling:
I think I'm hitting exactly the same issue myself - #924

mcrowson:
Could be made possible by https://github.com/boto/botocore/pull/1195

mbeacom:
I'm thinking I might wait for this PR to be merged and for us to up `botocore`, then commit the necessary changes to `handler.py`.
Would that be preferable or should I take a crack at it now @mcrowson ?

mbeacom:
I'm going to implement this for now.  If they ultimately handle it, we can update accordingly.

nils-braun:
Are there any news on this that I have missed (sorry for my stupid question)? I have seen your branch, @mbeacom, and I have implemented it in a similar and it seems to work (at least in my case). Do you want to do a PR or is there already one?

mbeacom:
@nils-braun The branch I have out currently dumps the zip to memory (can be a problem with large deployments and is restricted to the memory allocated to the function). I haven't submitted a PR for this yet because we're looking to stream the zip file, uncompressing the contents on the fly.  This will free up both memory and disk resources.

nils-braun:
Very good idea - maybe something like https://stackoverflow.com/questions/339053/how-do-you-unzip-very-large-files-in-python may help. I am definitely willing to help you with this PR, if you want or if you do not have the time to work on this now :-)

mcrowson:
[...]

PR #1022, opened 2017-07-28
slim_handler reduce /tmp use and enable significantly larger deployments  

olirice:
## Description
 
Updated slim_handler behavior to:
  - Delete project.zip after it has been extracted
  - Enable significantly larger deployments (see #1020)

Previously:
  - Download project.zip from s3 to /tmp
  - Unzip project.zip to /tmp/project
  - Do not delete project.zip

Now:
  - Download project.zip into an SpooledTemporaryFile
       - SpooledTempfile is in-memory
       - SpooledTempfile has a max_size based on the Lambda function's memory from an env var
       - SpooledTempfile seamlessly dumps to /tmp and continues if max_size is exceeded (safe)
       - SpooledTempfile deletes automatically once it falls out of scope
   - Unzip project.zip to /tmp

## GitHub Issues

#1020
#961
#881

### DIFF
--- a/README.md
+++ b/README.md
@@ -721,7 +721,11 @@ You can also simply handle CORS directly in your application. Your web framework
 
 #### Large Projects
 
-AWS currently limits Lambda zip sizes to 50 megabytes. If your project is larger than that, set `slim_handler: true` in your `zappa_settings.json`. In this case, your fat application package will be replaced with a small handler-only package. The handler file then pulls the rest of the large project down from S3 at run time! The initial load of the large project may add to startup overhead, but the difference should be minimal on a warm lambda function. Note that this will also eat into the _memory_ space of your application function.
+AWS currently limits Lambda zip sizes to 50 megabytes. If your project is larger than that, set `slim_handler: true` in your `zappa_settings.json`. In this case, your heavyweight application package will be replaced with a small handler package. When the Lambda function starts, the handler package downloads your project files from S3, unzips them into `/tmp` storage, and fires up your application! The initial load of the large project may add to startup overhead, but the difference should be minimal on a warm lambda function.
+
+For deployments between 250 and 500 megabytes, the `memory_size` of the Lambda instance will determine if your deployment succeeds. The `/tmp` folder on each Lambda instance has 500 megabytes of free space. When the handler package downloads the zip file containing your project from S3, it attempts to store that file in memory while it unzips it's contents to `/tmp/<your_project>/`. If the `memory_size` of the instance is smaller than the size of your zipped project, the zip file spills onto `/tmp` and eats into the maximum project size that can be unzipped.
+
+As a rule of thumb, if a project directory is between 250 and 500 megabytes on your development machine, set `memory_size` > 650 megabytes and your deployments will never fail due to insufficient resources.
 
 #### Enabling Bash Completion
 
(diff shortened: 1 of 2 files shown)


Miserlou:
This is really great, is there any chance you can also update the README file or write a [blog post](https://github.com/Miserlou/zappa-blog) about this for the release?

Poking @mcrowson for review as well.

olirice:
Great, README updated

The existing blog post on [Large Applications](https://blog.zappa.io/posts/slim-handler) already says that Zappa supports projects up to 500 MB. 

> Zappa deployments now support up to 500M of zipped up Python projects. Simply set “slim_handler”: true in zappa_settings.json and your large projects can now serve up requests from Lambda without a server.

What are you looking for in a blog post?

I could write a walkthrough with code snippets to create and deploy a 500 MB project and explain the memory considerations throughout? "Lets Deploy a 500 MB Project on Lambda" or something similar.

mcrowson:
Love the idea, i just don't think it makes a tremendous difference. It still ends up spilling onto disk and you have the zip size and zip folder to worry about. The only change here then is that we have /tmp space plus ram space to put the zip file and the unzipped contents. 

Setting the RAM size to 650M to account for all of it is over a huge cost increase for the whole application to get it all to fit. I still think a streaming approach is the right long term solution, but this might help the projects that just needed an extra 100M or something to fit it all. For those project zips over 300M though they still might not be big enough to account for the fully unzipped contents in addition to the zip file.

olirice:
@mcrowson
Agreed. Unfortunatey, Zipfile expects random access to the archive so streaming unzip isn't possible. 

@miserlou If I can change the project upload format to .tar.gz for slim handler projects then streaming unzip is totally feasible. 

Objections?

mcrowson:
totally on the same page. I say give it a go packaging with the gzipped tarball and streaming unzip that way

olirice:
@mcrowson streaming .tar.gz from s3 works great.

``` python
# Resources
remote_bucket, remote_file = parse_s3_url(project_zip_path)
s3 = boto_session.resource('s3')

# S3 File Object
remote_project = s3.Object(remote_bucket, remote_file)

# remote_project byte stream
raw_stream = remote_project.get()['Body']._raw_stream

# Create tarfile from byte stream. Note mode = 'r|gz' not 'r:gz' 
# For mode explanation see https://docs.python.org/2/library/tarfile.html
remote_archive = tarfile.open(None, 'r|gz', fileobj=raw_stream)

# Extract as usual
remote_archive.extractall(path=project_folder)
```

It looks like both the cli and core need updating to implement this feature. That's a little more than I intended to bite off. 

Do you know of any contributors who might consider updating the client-side tools?

If not, how about merging the `SpooledTemporaryFile` PR so those of us who insist on abusing Lambda have a simple (albeit costly) solution for 500mb until I have time to loop back around and implement the streaming solution.

mcrowson:
I think @mbeacom offered to do the whole of it as well over on #881

dswah:
I've already found this code to be super useful! 

I'd like to deploy my app with this, but I dont want to manually copy over handler.py every time i reinstall Zappa...

What's the status of this PR?

olirice:
@dswah I believe @mcrowson is waiting on an update wrt @mbeacom's streaming gzip implementation over on #881 before making a recommendation to merge this (not as good) solution.

@mbeacom, if you had any trouble wrestling that s3 file object into a stream (I did) there's a code snippet ^ that might be helpful.

olirice:
Closing due to better solution at #1037

Issue #1035, opened 2017-08-04
Cannot update an API which has custom endpoints and which has CORS enabled


jakul:
## Context
I have a Zappa API to which I add extra endpoints manually and which has CORS enabled. I have extra endpoints because they need a different authorizer than the ones handled by Zappa do.

With the latest master the deploy process crashes during deployment, because my extra endpoints do not have OPTIONS methods, and the CORS code added in https://github.com/Miserlou/Zappa/pull/1011 tries to update the `OPTIONS` method of every resource.

## Expected Behavior
Should be able to update an API which has custom endpoints and which has CORS enabled

## Actual Behavior
An error happens
```
==============

Traceback (most recent call last):
  File "/Users/craig/.virtualenvs/service-user/lib/python3.6/site-packages/zappa/cli.py", line 2489, in handle
    sys.exit(cli.handle())
  File "/Users/craig/.virtualenvs/service-user/lib/python3.6/site-packages/zappa/cli.py", line 473, in handle
    self.dispatch_command(self.command, stage)
  File "/Users/craig/.virtualenvs/service-user/lib/python3.6/site-packages/zappa/cli.py", line 517, in dispatch_command
    self.update()
  File "/Users/craig/.virtualenvs/service-user/lib/python3.6/site-packages/zappa/cli.py", line 904, in update
    self.zappa.add_binary_support(api_id=api_id, cors=self.cors)
  File "/Users/craig/.virtualenvs/service-user/lib/python3.6/site-packages/zappa/core.py", line 1320, in add_binary_support
    for resource_id in resource_ids:
  File "/Users/craig/.virtualenvs/service-user/lib/python3.6/site-packages/botocore/client.py", line 254, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/craig/.virtualenvs/service-user/lib/python3.6/site-packages/botocore/client.py", line 558, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.NotFoundException: An error occurred (NotFoundException) when calling the UpdateIntegration operation: Invalid Method identifier specified

==============
```

## Possible Fix
Only update the content handling on resources with have an OPTIONS method.

## Steps to Reproduce
1. Create and deploy a Zappa API with CORS enabled
2. Add an extra resource using the AWS web console
3. Re-deploy the API using Zappa

## Your Environment

* Zappa version used: Latest `master` -
[...]

PR #1036, opened 2017-08-04
fix/Only update CORS on resources which have an OPTIONS method

jakul:
## Description
Don't try to modify the `OPTIONS` methods on resources which don't have one

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1035

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1309,10 +1309,13 @@ class Zappa(object):
                 )
 
             if cors:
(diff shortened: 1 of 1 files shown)


jakul:
@eely22 FYI

eely22:
Thanks! Definitely not something we have tried (or even knew people did) so thanks for catching that corner case.

Issue #1038, opened 2017-08-05
Feature Request: zappa tail --color


pjf:
When piping tail output to a command, `zappa tail` automatically engages the `--no-color` option. This is usually what the user wants, but some users (ie: me) would love to enforce coloured output even if we're using pipes.

It would be amazing to have a `zappa tail --color` option that forces coloured output, regardless or what the default may otherwise have been.

Many thanks again for an *incredible* tool!

zaraken:
https://github.com/zaraken/Zappa/tree/z-tail-force-color
Tested with python 3.6
Feel free to review, test and give feedback if anything isn't as it should be.

Miserlou:
@zaraken - please send this as a Pull Request!

zaraken:
https://github.com/Miserlou/Zappa/pull/1039

PR #1039, opened 2017-08-07
#1038 Add --force-color option for zappa tail to display colors even if not supported

zaraken:
#1038 Add --force-color option for zappa tail to display colors even if not supported supported/auto-detected. --force-color overrides --no-color. Example usage "zappa tail --force-color | less -r"

Only tested with python 3.6


pjf:
Thank you!!

PR #1049, opened 2017-08-12
Use AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY for async boto Sessions if present

exp0nge:
## Description
 
Use AWS keys for boto Session for async requests.
## GitHub Issues

Miserlou/Zappa/issues/1043

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,6 +495,8 @@ def order_pie():
 
 And that's it! Your API response will return immediately, while the `make_pie` function executes in a completely different Lambda instance.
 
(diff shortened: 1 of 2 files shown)


mjschultz:
`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are set for every single Lambda invocation no matter what, so the `else` condition here will never execute in practice. I suspect this is working for you because you are explicitly setting those to variables via environment variables somewhere?

exp0nge:
I'm using the `zappa_settings.json` `environmental_variables` setting. I'm not sure what's going on with the session but this patch lets me use the async calls.

mjschultz:
I put a comment on your issue since I think that discussion is better put over there.

exp0nge:
Check issue for resolution

PR #877, opened 2017-05-23
add support for multiple expressions when scheduling functions

texonidas:
## Description
Adds a check for an `"expressions"` key when parsing `"events"` function list. If `"expression"` is found, it is inserted into a one element list. The scheduled event generator then iterates through each expression, resulting in `n` iterations for `n` expressions.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/876

Pre-emptive pull request given the relative triviality of the code changes.

PR #1051, opened 2017-08-14
Add index prefix to each subsequent schedule expression for the same event

texonidas:
## Description
This a fix for the bug introduced in PR #877, where each expression would be subsequently written over by the next. This PR generates a unique name for each rule in such a way that it does not interfere with Zappa's method of pulling methods of scheduled events.

```javascript
"events": [
            {
                "function": "application.utils.cron.some_func",
                "expressions": [
                    "cron(*/5 22-23 ? * SUN-THU *)",
                    "cron(*/5 0-8 ? * MON-FRI *)"
                ]
            }
        ]
```

The above snippet results in two new cloudwatch rules, `project-stage-application.utils.cron.some_func` and `project-stage-1-application.utils.cron.some_func`.

In the case of a named event, the resulting rule would be `project-stage-1-func_name-application.utils.cron.some_func`.

Apologies for the bug I introduced, I didn't notice that I had overwritten the schedule for the original rule upon the second.

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -2031,7 +2031,6 @@ class Zappa(object):
             expression = event.get('expression', None) # single expression
             expressions = event.get('expressions', None) # multiple expression
             event_source = event.get('event_source', None)
-            name = self.get_scheduled_event_name(event, function, lambda_name)
             description = event.get('description', function)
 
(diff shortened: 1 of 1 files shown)


Miserlou:
Thank you for taking ownership of this bug, and for providing an explanatory PR!

PR #1056, opened 2017-08-16
Adding XRay support

### LINKED ISSUE #842: AWS X-Ray Integration?
## Feature Request

Integrate support for AWS XRay. https://aws.amazon.com/xray/

> AWS X-Ray helps developers analyze and debug production, distributed applications, such as those built using a microservices architecture. With X-Ray, you can understand how your application and its underlying services are performing to identify and troubleshoot the root cause of performance issues and errors. X-Ray provides an end-to-end view of requests as they travel through your application, and shows a map of your application’s underlying components. You can use X-Ray to analyze both applications in development and in production, from simple three-tier applications to complex microservices applications consisting of thousands of services.

mathom:
## Description
Adding in a setting to enable X-Ray tracing now that there is an official Python SDK.

closes #842

See http://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-python.html for the application side of things

screen shot 2017-08-15 at 10 48 54 pm

screen shot 2017-08-15 at 10 50 01 pm


richiverse:
@mathom 

do you mind adding a section to the README for this? Seems super useful

mathom:
Good idea, I've added a small section to the readme @richiverse

Miserlou:
This is excellent!

I have a small question about the example you give in the documentation:

```python
```python
+from aws_xray_sdk.core import xray_recorder
+from aws_xray_sdk.ext.flask.middleware import XRayMiddleware
+
+app = Flask(__name__)
+
+xray_recorder.configure(service='my_app_name')
+XRayMiddleware(app, xray_recorder)
```

For this example, what is the `app_function`. Does it still work with `app`, or do we need to add a line for `xapp = XRayMiddleware(app, xray_recorder)` and then `app_function: app.xapp`?

mathom:
@Miserlou I decided to leave the actual xray implementation out of Zappa and just activate the required bits on the lambda function. The main reason is the official AWS python library is still in flux and there are a bunch of third party ones people have made.

I do have a generic WSGI shim we could throw on top of the app function, but it would probably have to be toggleable as AWS ships Flask and Django specific shims that would do the same thing.

mathom:
I also have a forum issue up for the fact that their SDK has some goofy behavior inside Lambda: https://forums.aws.amazon.com/thread.jspa?threadID=263192&tstart=0

mathom:
Might need to remove the example or add a my WSGI wrapper if upstream says they won't support it.

mathom:
Guess I never really answered your question on the end:

> For this example, what is the app_function. Does it still work with app, or do we need to add a line for xapp = XRayMiddleware(app, xray_recorder) and then app_function: app.xapp?

It still works with `app` like normal.

PR #1061, opened 2017-08-16
Ability to capture asynchronous responses

### LINKED ISSUE #1060: Ability to capture the responses from async tasks
## Context

The async tasks functionality is really useful, but wouldn't it be more useful if there was a way to capture the responses from functions called asynchronously?

## Expected Behavior

Functions run asynchronously, and their responses are queryable by the calling context and other future requests.

## Actual Behavior

Functions run, but their output is unavailable to the caller, and also unavailable to future app requests.

## Possible Fix

Assign asynchronous tasks an ID, and store the responses in a more-persistent datastore for later retrieval using this ID.

scoates:
## Description

Implements a mechanism (`capture_response`) to optionally capture the result of asynchronous tasks to DynamoDB.

Example (see the README for more):

```python
@app.route('/payload')
def payload():
    delay = request.args.get('delay', 60)
    x = longrunner(delay)
    return "Got response_id: {}".format(x.response_id)

@task(capture_response=True)
def longrunner(delay):
    sleep(float(delay))
    return {'MESSAGE': "It took {} seconds to generate this.".format(delay)}
```

Updates settings to allow specification of table name.

Updates README to document this feature, including an example app.

## GitHub Issues
Closes #1060

## Other

I think this can be considered for merge because the functionality is useful right away, but it's missing a few things, longer term:

- ~~DynamoDB table creation uses static values (`5`) for read + write provisioning, which currently costs "$2.91 / month" according to the console. Would be really nice to make this autoscale, but that's much more complicated.~~
- I did not add a test for this. It sounds extremely difficult to write a test that implements this without actually deploying an app during the test suite run.
- Timeout is still limited to the lambda limit (`timeout_seconds` in zappa settings), so there's a hard limit of 300s for these. The DynamoDB records have a TTL of ~600 (+lag) seconds; this is not configurable at this time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -535,11 +535,11 @@ run(your_function, args, kwargs, service='sns') # Using SNS
 ### Remote Invocations
 
 By default, Zappa will use lambda's current function name and current AWS region. If you wish to invoke a lambda with
-  a different function name/region or invoke your lambda from outside of lambda, you must specify the 
-  `remote_aws_lambda_function_name` and `remote_aws_region` arguments so that the application knows which function and 
-  region to use. For example, if some part of our pizza making application had to live on an EC2 instance, but we 
+  a different function name/region or invoke your lambda from outside of lambda, you must specify the
+  `remote_aws_lambda_function_name` and `remote_aws_region` arguments so that the application knows which function and
+  region to use. For example, if some part of our pizza making application had to live on an EC2 instance, but we
   wished to call the make_pie() function on its own Lambda instance, we would do it as follows:
-  
+
  ```python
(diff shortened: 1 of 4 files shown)


bcongdon:
Great feature!

I'd place a high emphasis on being able to configure the read/write capacity provisioning, though. Based on the experience I've had with Dynamo, 5 read/write is probably overkill for toy examples, and not enough for something that gets a modest amount of load.

scoates:
Agreed. I'll work on this.

scoates:
Does the change in 03314e1 work for you, @bcongdon?

bcongdon:
Yeah, awesome!

scoates:
Ok, @Miserlou, I'm done fiddling for now. Merge at will.

PR #1091, opened 2017-09-08
events take kwargs

mcrowson:
## Description
 
This allows events to take keyword arguments. The event sent to CloudWatch is recreated through the InputTransformer parameter and then modified to include the kwargs argument. 

The kwargs can then be obtained through

`kwargs = event.get('kwargs')`

## GitHub Issues

#359

### DIFF
--- a/README.md
+++ b/README.md
@@ -467,6 +467,26 @@ Similarly, for a [Simple Notification Service](https://aws.amazon.com/sns/) even
        ]
 ```
 
+Events can also take keyword arguments.
+```javascript
+       "events": [
+            {
+                "function": "your_module.your_recurring_function", // The function to execute
+                "kwargs": {"key": "val", "key2": "val2"},  // Keyword arguments to pass. These are available in the event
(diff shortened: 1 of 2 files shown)


mcrowson:
Recreated the events that it normally passes, and added kwargs:
http://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html#schedule_event_type

mcrowson:
Only adds kwargs to Cloudwatch cron-like events. Does not currently do so to events with another AWS source. Not sure how to do that here or if its needed:

https://github.com/Miserlou/Zappa/blob/master/zappa/core.py#L2165

CaseGuide:
If you're poking around here in 2020 and wonder what this looks like in code, the [linked example](https://github.com/Miserlou/Zappa/blob/master/example/zappa_settings.json) still works and [the code is here](
https://github.com/Miserlou/Zappa/blob/542c8c60291ebc5ece03b8dd4abe9e0172715086/zappa/core.py#L2645)

Issue #1087, opened 2017-09-07
Use Extends on a Core Env with Global


mcrowson:
## Context

 

Here: https://github.com/Miserlou/Zappa/blob/442484751d332d5b9cf01ef2f5de674a1cc3fa1b/zappa/cli.py#L1576 

The global option gives a little redundancy for each environment. For the most part I imagine a global deployment would use similar settings. I suggest creating a "core" stage that the other extend from. The only real reason to do this is to lower the barrier to entry for new teams/devs so that it doesn't feel like copy/pasting for them when they are just starting out and might not know about the "extends" key.

PR #1095, opened 2017-09-08
Global Stages Extend in zappa init

mcrowson:
…init, and other global stages extend the given stage.

## Description
 
When global deployments happen we still make the stage that they gave earlier in the init process and then the other global stages extend that one.

## GitHub Issues

#1087


mcrowson:
It looks currently like each global stage gets its own s3_bucket. Why is that? Is it necessary?

Miserlou:
No, I was just thinking about this. In fact, it should all use the same bucket using extends, which you've done here! Fantastico.

Miserlou:
Ah, I'm very sorry but this PR got stepped on! Would you mind resolving the conflict so that I can merge it in cleanly again? I'm not sure how this happened, sorry for taking so long!

PR #1103, opened 2017-09-13
update botocore3 to have awscli==1.11.151

richiverse:
This includes the ability to include athena with zappa deployments.

Also, any chance we can bump up the zappa pip version on pypi?


Miserlou:
Will publish a new release today hopefully.

Issue #1113, opened 2017-09-19
'LambdaAsyncResponse' object has no attribute 'response_id'


jwkvam:
I have a task that collects a bunch of subtasks to do and uses the async task functionality of zappa to run them.

## Context
I'm using a python 3.6 from a linux docker environment, it's based off this one: https://github.com/danielwhatmuff/zappa

## Expected Behavior
No Error.

## Actual Behavior
I get this traceback:
```
AttributeError: 'LambdaAsyncResponse' object has no attribute 'response_id'
  File "task/handler.py", line 242, in lambda_handler
    return handler.handler(event, context)
  File "task/handler.py", line 347, in handler
    result = self.run_function(app_function, event, context)
  File "task/handler.py", line 276, in run_function
    result = app_function(event, context)
  File "mymodule/tasks.py", line 103, in gather
    detect(v)
  File "zappa/async.py", line 403, in _run_async
    capture_response=capture_response).send(task_path, args, kwargs)
  File "zappa/async.py", line 164, in send
    'response_id': self.response_id,
```

These are some of the arguments to `_run_async` since I captured it in sentry.io

```
aws_region | 'us-west-2'
capture_response | False
func | <function detect from mymodule.tasks at 0x7eff0a860510>
kwargs | {}
service | 'lambda'
```

## Possible Fix
Not sure.

## Steps to Reproduce
It happens consistently every time I invoke the function remotely. Sorry I don't have a minimal shareable example.

## Your Environment
* Zappa version used: 0.44.1
* Operating System and Python version: Python 3.6.1
* The output of `pip freeze`:
```
argcomplete==1.9.2
base58==0.2.4
boto3==1.4.7
botocore==1.7.12
certifi==2017.7.27.1
chardet==3.0.4
click==6.7
contextlib2==0.5.5
cssselect==1.0.1
cssutils==1.0.2
cycler==0.10.0
docutils==0.14
durationpy==0.5
falcon==1.2.0
future==0.16.0
futures==3.1.1
hjson==3.0.0
hug==2.3.1
idna==2.6
Jinja2==2.9.6
jmespath==0.9.3
kappa==0.6.0
lambda-packages==0.16.1
lxml==4.0.0
MarkupSafe==1.0
matplotlib==2.0.2
numpy==1.13.1
pandas==0.20.3
peewee==2.10.1
pendulum==1.2.5
placebo==0.8.1
premailer==3.1.1
psycopg2==2.7.3
pyparsing==2.2.0
[...]

PR #1114, opened 2017-09-20
fix async tasks with `capture_response=False`

jwkvam:
## Description
Fix async tasks that don't capture response

## GitHub Issues
Fixes: #1113

PS: Would appreciate a minor version bump since this fixes a regression.


Miserlou:
Thanks. Fixed in 0.44.2.

PR #1117, opened 2017-09-20
using the same naming convention for update as for deploy

mcrowson:
Came up on slack, #1094 caused this issue. Needed to add the stage name to the current project variable in the update() as well as deploy()


mcrowson:
`zappa update` no longer works when slim_handler is set to true until this gets patched

Issue #700, opened 2017-03-06
Allow "tagging" a function version with a semantic version


aehlke:
## Context
When continuously deploying from CI, it's useful to be able to check which version of the source code is currently deployed to a stage or to production. It's important that if CI somehow accidentally runs out-of-order, it doesn't deploy an old version over a newer version of the code. (Services like Heroku enforce this out-of-the-box.)

Lambda doesn't seem to have any great way to check which source version is currently deployed, without storing that info out-of-band, or other awkward solutions. Is there a best practice here for Lambda?

My idea is to use the function version's description as the semantic version number (either the git hash, or the tagged release version, or etc. - something that can be compared with the deployment candidate version), since each version can have its own unique description. Currently Zappa uses a static description for all versions, rather than having some way to set it per deploy (without editing the source-controlled configuration file every deploy).

My workaround is to dynamically generate the Zappa YAML/JSON config every deploy, using the source-controlled copy as the starting point. This is pretty awkward.

Miserlou:
We currently use "Version" as an integer for rollback purposes, but there's no reason we have to do it that way. We could either have it be a timestamp, or we could create and Alias for each function which corresponds to some generator function in your settings if you need to tie it to a git tag or something like that. Happy to accept a PR which aligns with your use case here, seems like a very valid one.

Miserlou:
Yeah, I think creating an alias based on a generator function on create and update is probably the right way to do this.

aehlke:
Ah ok - I couldn't find the relevant docs and assumed the version numbers were auto-incrementing, not something we could set.

Miserlou:
Yeah, they are (AFAIK), but aliases are distinct from version numbers.

sam-10e:
I propose we add an `--alias` arg to the deploy/update command? That way CI tools could alias versions as they are deployed, which make them much easier to test.

Currently, we implement this by fetching the list of all lambda versions and find the highest version number, then add an alias to that. That can cause problems if two versions (form different PRs) are deployed for testing at the same time as it leads to a race condition.

PR #1133, opened 2017-09-27
Add cli option to alias lambda on deploy/update

sam-10e:
## Description
Add optional `--alias` parameter to CLI so that versions can have an alias associated with them as they are created.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/700


sam-10e:
@Miserlou - I'd love your input on this PR. Does this look like an approach that would satisfy #700?

Issue #1139, opened 2017-09-28
'zappa certify' unauthorized failure


ScottSturdivant:
## Context
Using python 3.6 and zappa in a virtualenv, I followed the instructions for getting up and running with a custom domain name and a Let's Encrypt certificate.  I was met with 403 error:

> Error signing certificate: 403 b'{\n  "type": "urn:acme:error:unauthorized",\n  "detail": "Error creating new cert :: authorizations for these names not found or expired: dev.example.com",\n  "status": 403\n}'

## Possible Fix
It looks like the the parse_csr() function is returning an empty set for the domains.  This is because with my version of OpenSSL (`OpenSSL 1.1.0f  25 May 2017`) it generates the domain.csr file with "CN = domain.name" while the domain parsing regex is not expecting whitespace.

ScottSturdivant:
Thank you!

PR #1140, opened 2017-09-28
Update CN regex to handle whitespace.

ScottSturdivant:
This is a proposed fix for https://github.com/Miserlou/Zappa/issues/1139

### DIFF
--- a/zappa/letsencrypt.py
+++ b/zappa/letsencrypt.py
@@ -198,7 +198,7 @@ def parse_csr():
     if proc.returncode != 0: # pragma: no cover
         raise IOError("Error loading {0}: {1}".format(csr_filename, err))
     domains = set([])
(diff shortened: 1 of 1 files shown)


Miserlou:
Does this work for both whitespace and non-whitespace?

ScottSturdivant:
It does, yes:

    >>> re.search(r"Subject:.*? CN\s?=\s?([^\s,;/]+)", "Subject: CN=foo.example.com")
    <_sre.SRE_Match object; span=(0, 27), match='Subject: CN=foo.example.com'>
    >>> re.search(r"Subject:.*? CN\s?=\s?([^\s,;/]+)", "Subject: CN = foo.example.com")
    <_sre.SRE_Match object; span=(0, 29), match='Subject: CN = foo.example.com'>

I couldn't find an existing test case to modify, but would be happy to do so if you can point me in the correct direction.

PR #1142, opened 2017-09-28
API Gateway only supports ACM certs from one region.

ScottSturdivant:
To help people following the guide, we should clarify this so they do
not create a cert in their default region, only to find that it's
unusable.

[AWS docs](http://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html) state:  

> To use an ACM Certificate with API Gateway, you must request or import the certificate in the US East (N. Virginia) (us-east-1) Region.

If you follow the guide as written, you may encounter an error such as this when running `zappa certify`:

    botocore.errorfactory.BadRequestException: An error occurred (BadRequestException) when calling the CreateDomainName operation: Invalid certificate ARN: redacted. Certificate must be in 'us-east-1'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -907,7 +907,7 @@ However, it's now far easier to use Route 53-based DNS authentication, which wil
 ##### Deploying to a Domain With AWS Certificate Manager
 
 1. Verify your domain in the AWS Ceriticate Manager console.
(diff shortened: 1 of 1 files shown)


Miserlou:
Excellent catch, thanks Scott!

Issue #1146, opened 2017-09-30
Add a global --color=auto/never/always flag


jleclanche:
At the moment, it is only `zappa tail` that supports a `--no-color` and a `--force-color` argument.

I'd like to suggest instead making a *global* `--color=[auto|never|always]` flag.

That particular flag syntax is very standard across all unix tools (grep, ls, etc)

Miserlou:
Into it.

brad-alexander:
This is a tiny bit hacky and I haven't implemented the "always" option as it seems it might need a customized solution for each subcommand like [colorize_log_entry](https://github.com/Miserlou/Zappa/blob/9c54b88a63e9778249dd6eedffbddeb00f4b1148/zappa/cli.py#L2345) for `zappa tail`

PR #1168, opened 2017-10-12
Add global option to disable color

brad-alexander:
## Description
- Add a global --color=auto/never/always flag
Implemented the "never" option by pushing a Click global context onto the stack.  This is normally managed by Click as part of its command nesting and context passing, but since we are currently only using click.echo and click.style, this should be safe for the meantime. If we start implementing Click more generally, this should be provided as options in a more standard way.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1146

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -14,9 +14,10 @@ import shutil
 import sys
 import tempfile
 
+from click.globals import resolve_color_default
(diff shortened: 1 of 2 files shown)


Miserlou:
Good solution, thanks!

PR #1169, opened 2017-10-12
Use `is` operator for comparing with `None` (Pep8)

smrutiranjanp:


Miserlou:
Travis doesn't like this for some reason.

topiaruss:
self.assertIsNotNone(event)

PR #1183, opened 2017-10-17
Add prefix and sufix filter to S3 events

ebertti:
and better formatting to not miss docs link


Miserlou:
Can you do this as a second rather than primary example? It makes the trivial use case more complicated than it needs to be.

PR #1202, opened 2017-10-27
Native support for block DoS Attacks via browser check cookie

the4rcane:


Miserlou:
Hey - this seems like a cool idea, can you maybe either start a ticket for discussion or rebase this into a more reasonable commit to review? 

Thanks!

Issue #1165, opened 2017-10-11
Check Config Files for Allowed Values


mcrowson:
## Context

 
Was not seeing my remote environment variables because my config value was "remote_evn". Mistake on my part, but it would be nice if zappa checked config files to ensure that values being set were even values used by zappa. 

## Expected Behavior

When a user gives a config value not used by Zappa, it tells the user and suggests that it might be a typo. Just lowers the learning curve slightly and improves QOL.

## Actual Behavior

Users are allowed to wallow in their mistakes

## Possible Fix

Keep a white list of config keys and test against those.

Miserlou:
Related/dupe: https://github.com/Miserlou/Zappa/issues/78

samuelrey:
I'll look into this!

PR #1215, opened 2017-10-31
Check config

daphee:
## Description
This adds a new function `verify_settings` to `cli.py` that is called when settings are loaded.
It goes through the settings dict and compares all used keys with a whitelist in `valid_settings.py`. For anything that isn't included in the whitelist a warning is printed including the valid setting with the closest [Levenshtein-Distance](https://en.wikipedia.org/wiki/Levenshtein_distance).

I got the list of config options from [there](https://github.com/daphee/Zappa/tree/check-config#advanced-settings). I also looked through `load_settings` and checked that at least all options that were loaded there were included in the whitelist.

I am not sure how I'd write a test for that. While I guess this isn't very critical or complicated code it did get a little bit more messier than I initially thought.

## GitHub Issues
Suggested in issue #1165.


Miserlou:
Hey Felix, thanks for doing this! I haven't forgotten about it, I just need to think about the approach a bit more. The explicit approach here feels like a pretty big maintenance burden - rather than enumerating all the settings and making sure that they all match up, I think the could would be a lot more maintainable and easy for people to add new settings if the check only enumerated ones that didn't use the default option in the dict, does that make sense?

daphee:
Do you mean like a function like `get_from_settings(settings_dict, settings_name)` that checks whether `settings_name` is in the dict and if not look for similar entries in `settings_dict`?

Either way I think this would definitely a better approach than the current one. 

This would also mean that if I miss a setting it would still work but just don't print a warning when this setting I missed is misspelled. My current approach on the other hand would print a warning when a setting I missed is used.

daphee:
I think I'll need a little help:

- I am still on the fence about where to add the check: Add a class that subclasses dict or monkey patch the `get` method of the `settings` dict in `stage_config`. I think this will depend on the answer to the next question.
- The new `get` method would just wrap the original one but also try to check whether a missing key was mistyped. To do that it would need the most similar key, *that is not a whole different setting*. I'm not sure how to do the latter. 
    - Maybe only check for mistyped keys after all settings have been read in `load_settings`
      We would need to keep a list of missing keys and a list of successfully used keys and then do the checks at the end of `load_settings`. Do you think that would work?

Issue #1218, opened 2017-11-02
Support user auth with GetFederationToken


cyounkins:
## Context

Some corporate environments use AWS Security Token Service to deal out temporary credentials to developers. Comparison of STS methods: http://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html#stsapi_comparison

Note that if you give out temporary credentials using GetFederationToken, the assumed role cannot call any IAM methods. IAM:GetRole is required by Zappa even with "manage_roles": false, and this causes it to fail.

## Expected Behavior

I expect `update` to work in my federated environment.

## Actual Behavior

Zappa tries to call GetRole which fails.

```
$ zappa update production
Calling update for stage production..
Downloading and installing dependencies..
Packaging project as zip.
Uploading image-viewer-production-1509656439.zip (12.9MiB)..
...
Updating Lambda function code..
Updating Lambda function configuration..
Oh no! An error occurred! :(

==============

Traceback (most recent call last):
  File "/Users/cyounkins/code/av/source/offboard/labeler/tools/elastic_labeling/image_viewer/env/lib/python2.7/site-packages/zappa/cli.py", line 2525, in handle
    sys.exit(cli.handle())
  File "/Users/cyounkins/code/av/source/offboard/labeler/tools/elastic_labeling/image_viewer/env/lib/python2.7/site-packages/zappa/cli.py", line 484, in handle
    self.dispatch_command(self.command, stage)
  File "/Users/cyounkins/code/av/source/offboard/labeler/tools/elastic_labeling/image_viewer/env/lib/python2.7/site-packages/zappa/cli.py", line 528, in dispatch_command
    self.update()
  File "/Users/cyounkins/code/av/source/offboard/labeler/tools/elastic_labeling/image_viewer/env/lib/python2.7/site-packages/zappa/cli.py", line 887, in update
    aws_kms_key_arn=self.aws_kms_key_arn
  File "/Users/cyounkins/code/av/source/offboard/labeler/tools/elastic_labeling/image_viewer/env/lib/python2.7/site-packages/zappa/core.py", line 1019, in update_lambda_configuration
    self.get_credentials_arn()
  File "/Users/cyounkins/code/av/source/offboard/labeler/tools/elastic_labeling/image_viewer/env/lib/python2.7/site-packages/zappa/core.py", line 2009, in get_credentials_arn
    self.credentials_arn = role.arn
[...]

PR #1219, opened 2017-11-04
Add role_arn settings variable

cyounkins:
## Description
 
Adds the role_arn settings variable to avoid any use of the IAM API, which is normally used to resolve a role name to an ARN.

## GitHub Issues

#1218 - Support user auth with GetFederationToken


Miserlou:
Looks good!!

cyounkins:
I tested it in my environment, yes.

PR #1221, opened 2017-11-07
Fix reversed args in README example for invoke

daneah:
## Description

Fixes reversed args in example snippet in README for invoking remote commands.

This is my first contribution here and it seemed like it would be more helpful to open this PR than create an issue first; if the latter is still preferable I'm happy to do so. Let me know!


Miserlou:
Nice catch, Dan!

daneah:
Thanks for the quick turnaround @Miserlou 👍

Issue #813, opened 2017-04-27
`tox.ini` to automate tests in different python versions and building docs locally


vlcinsky:
Currently testing locally or building documentation requires some manual preparation and the tests will not run on Windows (as there is `test.sh` script).

Currently building the docs requires manual installation of sphinx and sphinx_rtd_theme packages and running `$ make html` in docs directory.

Proposed solution
=================

Create `tox.ini` file, which would:
- run test in every zappa supported python version by `$ tox` for all versions or `$ tox -e py27`
  for specific version.
- allow building Sphinx docs via `$ tox -e docs`

Out of scope
============
This task would exclude (apart from things breaking the build):
- edits to the docs 
- modifying test suite

What we get from it
===================
Testing locally under different python interpreters would be much simpler (just run `$ tox`).

Building sphinx documentation locally would be simpler. As the docs is not published to readthedocs site yet, it would effectively make the documentation available at least locally.

Developer on Windows would have higher chance to run the tests and build the docs.

If this sounds acceptable, I am volunteering to do this modification (without testing under Windows).

vlcinsky:
Relates to issue #243 as it makes building (and maintaining) the docs a bit simpler.

PR #1229, opened 2017-11-11
Add tox and simplify Travis CI builds

bxm156:
## Description

1.  Migrated the project to use tox
3. Updated the README
4. Slight refactoring

## GitHub Issues
Relates to #813

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,10 @@ sudo: false
 
 language: python
 
+python:
+  - "2.7"
+  - "3.6"
+
 cache:
   - pip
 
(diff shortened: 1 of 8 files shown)


bxm156:
@Miserlou  Perhaps you can help me, I am running into an issue with a test that has a known problem. It's always failing on my local branch, even without my changes. I suspect its due to caching between builds, and that something was cached on Jenkins for a while, which with my change somehow changes that.

Here is the line, which has a comment about how its broken. https://github.com/Miserlou/Zappa/blob/master/tests/tests_placebo.py#L186

bxm156:
Update: Looks like test pollution. If I run the tests for just tests/tests_placebo.py they pass.

PR #1230, opened 2017-11-12
Migrate bash script to tox

### LINKED ISSUE #813: `tox.ini` to automate tests in different python versions and building docs locally

bxm156:
## Description
1. Migrated ./test.sh to tox
2. Fixed a flaky test
3. Updated Readme

## GitHub Issues
Relates/Fixes #813

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,32 @@
+sudo: false
+
 language: python
+
 python:
   - "2.7"
   - "3.6"
-# command to install dependencies
+
 cache:
   - pip
+
+before_install:
+  - pip install coveralls>=1.1
+
 install:
-  - "pip install setuptools --upgrade; pip install -r test_requirements.txt; pip install -e git+https://github.com/django/django-contrib-comments.git#egg=django-contrib-comments; python setup.py install" 
-# command to run tests
+  - pip install tox-travis
+
 env:
   - TESTCASE=tests/test_handler.py
   - TESTCASE=tests/tests_middleware.py
   - TESTCASE=tests/tests_placebo.py
   - TESTCASE=tests/tests_async.py
   - TESTCASE=tests/tests.py
-script: 
-  - nosetests $TESTCASE --with-coverage --cover-package=zappa --with-timer
-    #  - coverage combine --append
+
+script:
+  - tox -- $TESTCASE
+
 after_success:
   coveralls
+
 notifications:
   slack: zappateam:TTJ0mfHunDK0IBweKkEXjGpR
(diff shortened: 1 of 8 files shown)


Miserlou:
Yep, this is very close, but I have two needs for this:

* `test.sh` needs to work, even if its just calling something else inside the script
* I need a documented (and ideally scriptable) way to run a specific `test`, not a specific test file. This is documented in the current `test.sh`.

Looks fine otherwise, thanks!

bxm156:
Sure, out of curiosity, why test.sh? Is there a specific reason? Maybe a Makefile would be better?

bxm156:
I'm trying to figure out why coveralls reporting on this so many extra times. In the past, coveralls actually reported twice, though it edited its existing comment and so it showed as once. I think it was reporting twice because it saw the py27 and p36 builds. Now its reporting 10 times, and not getting hidden in the UI as easily.

Miserlou:
Are we good to merge now?

bxm156:
@Miserlou Yeah. The multiple coveralls thing is weird, but I verified that part of the flow hasn't changed, and it would seem I'm not the only one seeing it in their github projects.

bxm156:
@Miserlou @scoates  It looks like the last remaining issue with coveralls and multiple parallel builds (travis ci matrix) has been sorted out:

http://docs.coveralls.io/parallel-build-webhook

I can add the necessary changes, but it looks to me like I will need coveralls repo token, which is a secret access key that only @Miserlou can provide, and it would probably need to be added to the .travis.yml file encrypted.

So I can see a couple options:
1. Don't worry about any of it, an live with coveralls pinging an issue 5 times.
2. Remove the build matrix. All the test cases will still run, but travis ci won't seperate them into 5 lines.
3. Implement the hook, and inserting the secret repo token encrypted into the .travis.yaml file.

bxm156:
This should be ready to be pulled.

PR #1242, opened 2017-11-15
Package custom modules

tista3:
## Description
new setting: list of module names that will be packaged into zip

## Motivation
I have many flask apps in my repository that shares common module. This module is available in the PYTHONPATH env variable which points into directory where my common module directory is. I had to copy the common module before every `zappa update` into the flask app directory to be packed into distribution zip. But with this feature I just list the common modules and they are packaged into zip automatically if they can be imported in my environment.

Now I use my custom pypi package `timo-zappa`, but I will be happy If it could be in official zappa package. I tested it on Linux, py2.7, py3.6 and as 3.6 lambda.

This is my first pull request, be gentle :)


jneves:
Mental note: need to check this versus the proposed implementations for requirements.

Issue #1252, opened 2017-11-23
Zappa Undeploy Fails when you have event source as Kinesis


mthipparthi:
When you have Kinesis as event source  to trigger AWS lambda, while performing "undeploy" , we are having "Key error"  and undpeloy is failing.
We are getting below error while undeploying.
`
  File "/var/task/venv/local/lib/python2.7/site-packages/zappa/cli.py", line 1087, in undeploy
    self.unschedule()  # removes event triggers, including warm up event.
  File "/var/task/venv/local/lib/python2.7/site-packages/zappa/cli.py", line 1193, in unschedule
    events=events,
  File "/var/task/venv/local/lib/python2.7/site-packages/zappa/core.py", line 2460, in unschedule_events
    print("Removed event " + name + " (" + str(event_source['events']) + ").")
KeyError: 'events
`
## Expected Behavior
zappa undeploy should be success

## Actual Behavior
zappa undeploy Failing

### Permanent Solution :
This portion of the code is causing the issue. we should check 'events' before we print this below line.

print("Removed event " + name + " (" + str(event_source['events']) + ").")

https://github.com/Miserlou/Zappa/blob/7012e25e2bcfa8bd0c458622f66ee9682d78af3b/zappa/core.py#L2464

        "events": [
           {
               "function": "functions.lambda_handler",
               "event_source": {
                    "arn":  "arn:aws:kinesis:*:*:stream/stream_name",
                    "starting_position": "LATEST",
                    "batch_size": 1,
                    "enabled": true,
               },
               "name": "name_lambda"
           }
        ],

### Temporary fix.
change the configuration
        "events": [
           {
               "function": "functions.lambda_handler",
               "event_source": {
                    "arn":  "arn:aws:kinesis:*:*:stream/stream_name",
                    "starting_position": "LATEST",
                    "batch_size": 1,
                    "enabled": true,
                    _**"events"= []**_
               },
               "name": "name_lambda"
           }
        ],

## Steps to Reproduce

1. Have events in zappa settings like below.
        "events": [
           {
               "function": "functions.lambda_handler",
               "event_source": {
                    "arn":  "arn:aws:kinesis:*:*:stream/stream_name",
                    "starting_position": "LATEST",
                    "batch_size": 1,
                    "enabled": true
               },
               "name": "name_lambda"
           }
        ],
2. zappa undeploy

## Your Environment

* Zappa version used: zappa==0.44.3
* Operating System and Python version: lambci/lambda:build-python2.7
[...]

PR #1253, opened 2017-11-23
[#1252] - Zappa Undeploy Fails when you have event source as Kinesis

mthipparthi:
### Issue 
When you have Kinesis as event source to trigger AWS lambda, while performing "undeploy" , we are having "Key error" and undpeloy is failing.

 
Not every event_resource in the zappa_settings will have "events" in it. For instance Kinesis will have no entry for "events" in the configuration. So in this case, `zappa undeploy` is failing as result of KeyError - 'events'. Please refer to issue [ #1252] for detailed explanation.

## GitHub Issues

[#1252]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ __pycache__/
 # Distribution / packaging
 .Python
 env/
+venv/
 build/
 develop-eggs/
 dist/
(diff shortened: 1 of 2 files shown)


jneves:
Fixed on #1617

PR #1254, opened 2017-11-23
Exit with non 0 error code on failed zappa invoke

inytar:
## Description
 
We use `zappa invoke` in our CI. Right now if the invoked command fails the error is printed, but no exit code is given. This PR adds a CLI argument (`--exit-code`) which will result in the invoke command failing if the invoked function failed. This was added as an extra CLI argument to keep the current behavior for people depending on it.

This PR still misses a test, I'd love some help guiding me in the right direction on how I would write one for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -367,6 +367,10 @@ You can also invoke interpretable Python 2.7 or Python 3.6 strings directly by u
 
     $ zappa invoke production "print 1 + 2 + 3" --raw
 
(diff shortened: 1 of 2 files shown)


Miserlou:
Thanks for this, @inytar! I've merged this in manually but removed the CLI argument since I think this should be the proper behavior.

PR #1257, opened 2017-11-24
Add support for AWS Cognito Triggers

millarm:
## Description
Add support for configuring AWS Cognito triggers.
You can now map any of the triggerSources documented
http://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-identity-pools-working-with-aws-lambda-triggers.html to any function in your Zappa deployment.

This allows for extensive, simple customisation of the AWS Cognito authentication flow, so you can automatically approve users from defined domains.   If you need to check those domains against blacklists in your DB you have all the same access to DB records from the handler.


millarm:
Have I mentioned how much I love unit tests + travis?  :-)

Issue #695, opened 2017-02-27
On updating "Handler" is reset to default "handler.lambda_handler"


senseysensor:
## Context

I suspect this may happen because `apigateway_enabled` is set to `false` in my config. (This function should be invoked by dynamodb stream). `app_function` in config is set to my custom value but each time after updating I'm forced to go to lambda console and change value back from default one.
 

## Expected Behavior

On project update handler should be correctly set according to the value from config.

## Actual Behavior

Handler is reset to default `handler.lambda_handler`.

## Your Environment

* Zappa version used: 0.37.1
* Operating System and Python version: os x, python 2.7.13
* Your `zappa_settings.py`:

dabrazhe:
Is there a workaround? This behaviour still persists.

ctcampbell:
I've also got this problem on 0.45.1

LaundroMat:
I'm experiencing this as well (0.51.0).

PR #1264, opened 2017-11-27
#695 in case apigateway_enabled==false the app_function is called dir…

jakubzeman-acc:
## Description
When the API gateway is disabled the default web app flow is still called.

This pull request solves it by calling the the app_function with parameters event and context.

## GitHub Issues
[#](https://github.com/Miserlou/Zappa/issues/695)


wobeng:
@Miserlou  any update on this pull request?

I want to use zappa for lambda function only

jgissend10:
Interested!

wobeng:
@Miserlou any update on this pull request?

jneves:
Sorry, must be being a bit thick regarding this, but can you explain me what's the difference between using this or just using lambda_handler directly as a use-case? TIA.

wobeng:
@jneves all in one, one stop solution. I currently maintain my own zappa because of this...Once this is merge, I can delete the one I maintain

jneves:
@wobeng can you answer my question above though?

jneves:
@wobeng I don't like this because it breaks a few use cases, particularly support for manually set up API Gateways. For the case you are describing you already have lambda_handler. I can only see that not working in the case you need some zappa handler support and a specific function to be invoked. Currently the option people have is creating an handle that inherits from ZappaHandler for that. I'd welcome a kind of default_handler or something like this for these cases, but app_function takes a WSGI function as its argument and I can easily see it creating confusion to have either a WSGI function or a lambda handler as an argument to the same key.

wobeng:
@jneves what is the purpose of apigateway_enabled flag? One will think if apigateway_enabled is false then the function is invoked directly

Also, Zappa does more than just API gateway, it has side features like create events, roles and etc

Last, when you say " can you explain to me what's the difference between using this or just using lambda_handler directly as a use-case" can you elaborate

jneves:
No, it just doesn't create the api gateway (so you can connect an already existing one, for instance). It doesn't change any of the other behaviour. If you want it to just run a function you can define one with the lambda_handler option and short circuit all the extra processing zappa's handler does for you.

wobeng:
wait so you are saying if I set lambda_handler to custom function and set apigateway_enabled to false I will get the same behaviour as this pull request?

jneves:
Close to it, hence my question on what problem does this solve?

wobeng:
if your theory is true then I don't think we need this...I will test it out

wobeng:
@jneves I'm testing this out now and I have a question, I don't need to set the app `app_function` right? only ` "lambda_handler": "main.handler",` and `    "apigateway_enabled": false,` right?

jneves:
Correct.

PR #1268, opened 2017-11-29
Add documentation to the README for Cognito triggers

millarm:
## Description
Detailed documentation for cognito triggers

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,7 +577,7 @@ def pre_signup(event, context):
     return event
 ```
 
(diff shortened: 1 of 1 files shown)


millarm:
The interesting part of this PR is this change:

https://github.com/Miserlou/Zappa/pull/1268/commits/b8e4fcdf7c1a4d15c15223f1c276155c49e2849f

As I've hit multiple times the mistake "runs fine locally, then fails when deployed" because I've forgotten to make sure the object is serialisable.    This makes this reproducible locally, by... trying to serialise and deserialise the object.  :-)

millarm:
Apologies for the delay in making the PR!

Issue #1270, opened 2017-11-29
async task feature: access event and context for integration with raven-python-lambda


jwkvam:
I'm experimenting with this [library](https://github.com/Netflix-Skunkworks/raven-python-lambda) to get additional information into sentry. It works fine with "normal" functions since they take the event and context arguments and this library expects that.

The problem I'm having is that for async zappa tasks, they don't take the event and context arguments. This means I can't get the additional reporting this package provides from my async tasks.

## Context
I setup my async task which I first decorate to get sentry reporting goodies.
```
@task
@RavenLambdaWrapper()
def bar(event, context):
    logging.info(f'{event}, {context}')
```
Then trying to call `bar` yields:
```
decorated() missing 1 required positional argument: 'context': TypeError
Traceback (most recent call last):
File "/var/task/handler.py", line 511, in lambda_handler
return LambdaHandler.lambda_handler(event, context)
File "/var/task/handler.py", line 242, in lambda_handler
return handler.handler(event, context)
File "/var/task/handler.py", line 347, in handler
result = self.run_function(app_function, event, context)
File "/var/task/handler.py", line 276, in run_function
result = app_function(event, context)
File "/var/task/zappa/async.py", line 253, in route_lambda_task
return run_message(message)
File "/var/task/zappa/async.py", line 289, in run_message
**message['kwargs']
TypeError: decorated() missing 1 required positional argument: 'context'
```

## Expected Behavior
It's a little more work for the end user to extract the payload. But I think it's easier than using the boto api.

## Actual Behavior
Zappa and this library don't work together for async tasks.

## Possible Fix
Add an additional argument to `task` such as `deserialize` which could default to `True` and maintain current behavior. If `deserialize` is `False` then `event` and `context` would get passed on to the user's function.

For now I think I'm going to workaround this by using the boto api to invoke the function.

## Steps to Reproduce
Here's a self contained example.
```
from raven_python_lambda import RavenLambdaWrapper
from zappa.async import task

@task
@RavenLambdaWrapper()
def bar(event, context):
    logging.info(f'bar called {arg}')
    raise Exception(f'bar {event} {context}')

@RavenLambdaWrapper()
def foo(event, context):
    bar(5)
```

## Your Environment
* Zappa version used: 0.44.3

zaraken:
I worked around it by updating async.py with the following
`
@@ -214,7 +214,7 @@ def route_lambda_task(event, context):
     imports the function, calls the function with args
     """
     message = event
[...]

PR #1281, opened 2017-12-05
No deserialize

jwkvam:
Before you submit this PR, please make sure that you meet these criteria

* If you made a new function, did you **write a good docstring** for it?
I attempted so, I didn't add anything to the README.

* Did you write a test for your new code?
I have not, I could use some pointers.

* Did you **make sure this code actually works on Lambda**, as well as locally?
Yes, I tested this on Lambda it works as expected.

* Did you test this code with both **Python 2.7** and **Python 3.6**? 
Only 3.6, I can test it out on 2.7.

## Description
This adds another kwarg to async.task `deserialize` which controls what arguments are passed to the user defined function. It follows the design of `capture_response` very closely.

This was simple to implement and try out that I went ahead and did it. Otherwise I would have waited to see how others felt about it in my original ticket.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1270


rgov:
Python 2.7 tests are failing until #1298 is applied. This change passes Python 3.x tests.

jwkvam:
@rgov thanks for the heads up, I had no idea frankly why it was failing.

Issue #1295, opened 2017-12-16
PermissionError trying to make first deployment


rgov:
## Context
I'm trying to do my first Zappa deployment and hit an exception:

    Calling deploy for stage dev..
    Creating project-dev-ZappaLambdaExecutionRole IAM Role..
    Creating zappa-permissions policy on project-dev-ZappaLambdaExecutionRole IAM Role.
    Oh no! An error occurred! :(
    
    ==============
    
    Traceback (most recent call last):
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/cli.py", line 2610, in handle
        sys.exit(cli.handle())
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/cli.py", line 505, in handle
        self.dispatch_command(self.command, stage)
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/cli.py", line 539, in dispatch_command
        self.deploy(self.vargs['zip'])
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/cli.py", line 709, in deploy
        self.create_package()
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/cli.py", line 2171, in create_package
        disable_progress=self.disable_progress
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/core.py", line 504, in create_lambda_zip
        copytree(cwd, temp_project_path, symlinks=False, ignore=shutil.ignore_patterns(*excludes))
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/utilities.py", line 54, in copytree
        copytree(s, d, symlinks, ignore)
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/utilities.py", line 54, in copytree
        copytree(s, d, symlinks, ignore)
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/utilities.py", line 54, in copytree
        copytree(s, d, symlinks, ignore)
      File "/.../.direnv/python-3.6.3/lib/python3.6/site-packages/zappa/utilities.py", line 32, in copytree
        shutil.copystat(src, dst)
      File "/.../.direnv/python-3.6.3/lib/python3.6/shutil.py", line 218, in copystat
        lookup("chflags")(dst, st.st_flags, follow_symlinks=follow)
[...]

PR #1297, opened 2017-12-16
Skip file metadata when copying files for packaging

rgov:
## Description
This adds a flag to the `copytree` implementation that skips copying metadata such as user permissions. These permissions are ignored when creating the package zip anyway. This works around an issue on macOS which I've filed to the Python bug tracker: https://bugs.python.org/issue32347

## GitHub Issues
#1295


rgov:
Python 2.7 tests are failing until #1298 is applied. This change passes Python 3.x tests.

PR #1298, opened 2017-12-16
Require Django <2.0 for Python 2.7 compatibility

rgov:
`test_requirements.txt` just specifies a minimum version of Django, not a maximum. However, the latest version of Django, 2.0, [does not support Python 2.7](https://docs.djangoproject.com/en/2.0/releases/2.0/#python-compatibility). Thus the Python 2.7 tests always fail.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,6 +1,6 @@
 coveralls>=1.1
 coverage>=4.3.1
-Django>=1.10.5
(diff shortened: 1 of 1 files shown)


Miserlou:
Hey @rgov , thanks for the PR but it actually got fixed in another commit from elsewhere. Thanks!

Issue #1296, opened 2017-12-16
Unsafe uses of temporary directory


rgov:
In several location Zappa creates temporary directories by constructing paths rooted at `tempfile.gettempdir()`. The paths are likely predictable on some platforms due to subdirectory names that use to fixed strings or easily predictable values (such as the local time) This is risky behavior and is likely vulnerable to [TOCTTOU race conditions](https://en.wikipedia.org/wiki/Time_of_check_to_time_of_use).

From a quick audit, it looks like there are a few places where `tempfile` is used unsafely.

- In `zappa/letsencrypt.py`, encryption keys are written directly into the root temporary directory which seems very risky to me. These should definitely be changed to use a top-level temporary directory generated with `tempfile.mkdtemp()` which will have the correct permissions set.

- In `zappa/core.py`, it is used unsafely to populate `temp_project_path` and `temp_package_path`. These uses can be changed to simply use `tempfile.mkdtemp()` and the subsequent directory creation can be removed.

- In `zappa/core.py`, it is used to generate cache directories. Since it may be undesirable to re-populate the cache on every execution, maybe using `tempfile.mkdtemp()` is not the correct solution.

- In `zappa/cli.py`, it is used unsafely to download or copy an `account.key` file related to Let's Encrypt support.

scoates:
Looks like this was fixed in #1299. Please reopen if not.

PR #1299, opened 2017-12-16
Securely create temporary directory for Let's Encrypt processing

rgov:
Note: I have not tested this change beyond running tests on Travis CI, as I do not yet know how to exercise the code. However, due to the severity of handling cryptographic keys in insecure temporary directories, I am sharing my progress towards a resolution of #1296.

- Lazily creates a temporary directory with `tempfile.mkdtemp()` that is used for all processing in `zappa/letsencrypt.py`.

- Deletes the Let's Encrypt temporary directory when `cleanup()` is called, or when Python exits. The `no_cleanup` option is removed for now.

- Creates the temporary directories for `temp_project_path` and `temp_package_path` using `temple.mkdtemp()`.

- Replaces the custom `Popen` call with `subprocess.check_output` and family, and uses built-in file I/O functions when possible. (A little out-of-scope, sorry.)

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1209,16 +1209,23 @@ class TestZappa(unittest.TestCase):
 
         # We need a fake account key and crt
         import subprocess
-        proc = subprocess.Popen(["openssl genrsa 2048 > /tmp/account.key"],
-            stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
-        out, err = proc.communicate()
-        if proc.returncode != 0:
-            raise IOError("OpenSSL Error: {0}".format(err))
-        proc = subprocess.Popen(["openssl req -x509 -newkey rsa:2048 -subj '/C=US/ST=Denial/L=Springfield/O=Dis/CN=www.example.com' -passout pass:foo -keyout /tmp/key.key -out test_signed.crt -days 1 > /tmp/signed.crt"],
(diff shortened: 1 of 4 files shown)


tests/tests.py
```diff
@@ -1204,13 +1203,11 @@ def test_lets_encrypt_sanity(self):
             '-newkey', 'rsa:2048',
             '-subj', '/C=US/ST=Denial/L=Springfield/O=Dis/CN=www.example.com',
             '-passout', 'pass:foo',
-            '-keyout', '/tmp/key.key',
-            '-out', 'test_signed.crt',
+            '-keyout', os.path.join(gettempdir(), 'key.key'),
+            '-out', os.path.join(gettempdir(), 'test_signed.crt'),
```
rgov: I think perhaps it's intended to write out to `signed.crt`, but the original code did not do that.

rgov:
It looks like there might be a little more console spew where `subprocess.check_call` or `subprocess.check_output` is no longer consuming stdout or stderr. The solution to that is to open `os.devnull` and pass it. I'll make that change.

rgov:
Let's Encrypt sanity check is now passing. More changes incoming. Notably, I removed the `--no-cleanup` option from the command line interface; now cryptographic material is always deleted, even if Python hits an exception.

rgov:
CI tests are now passing. However I recommend giving it one real-life try before merging. (I am not yet using Zappa in my development, so I rely on the unit tests.)

PR #1313, opened 2017-12-21
Add SNS message filters

tripliks:
## Description
AWS recently introduced SNS message filtering using message attributes. This is very useful if you use single topic for multiple consumers (SQS, Lambda, etc) and you don't want every message to trigger function or just don't want to filter messages inside Zappa. 

By simply setting the following in your config:
```javascript
        "events": [
            {
                "function": "your_module.your_function",
                "event_source": {
                    "arn":  "arn:aws:sns:::your-event-topic-arn",
                    "filters": {
                        "interests": ["python", "aws", "zappa"],
                        "version": ["1.0"]
                    },
                    "events": [
                        "sns:Publish"
                    ]
                }
            }
        ]
```
a Zappa function will be executed for SNS messages with attributes where at least one value matches `interests` list.

Note that each filter must be list of strings.

For detailed documentation how it works: http://docs.aws.amazon.com/sns/latest/dg/message-filtering.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -516,6 +516,27 @@ Similarly, for a [Simple Notification Service](https://aws.amazon.com/sns/) even
         ]
 ```
 
+Optionally you could add [SNS message filters](http://docs.aws.amazon.com/sns/latest/dg/message-filtering.html):
+
+```javascript
+        "events": [
+            {
+                "function": "your_module.your_function",
+                "event_source": {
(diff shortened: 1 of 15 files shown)


tripliks:
Hey @Miserlou any thoughts or comments on this?

tripliks:
This is also possible using post callback:  https://gist.github.com/tripliks/948739af2140af7875c2f6d8ce454e28

Issue #1323, opened 2018-01-01
Parameter validation failed when using @task_sns.


Debetux:
## Context
I want to invoke a task which use @task_sns.

E.g. code:
```
@task_sns
def my_async_task(data):
    a = data.get('a')
    b = data.get('b')
    return a + b


def my_endpoint():
    my_async_task({
        'a': 4,
        'b': 10
    })

    return jsonify({
        'success': True
    })
```

This code produces this error:
```
Parameter validation failed:
Invalid type for parameter Message, value: b'{"task_path": "my_async_task", "capture_response": false, "response_id": null, "args": [{"a": 4, "b": 10}], "kwargs": {}, "command": "zappa.async.route_sns_task"}', type: <class 'bytes'>, valid types: <class 'str'>: ParamValidationError
Traceback (most recent call last):
  File "/var/task/handler.py", line 509, in lambda_handler
    return LambdaHandler.lambda_handler(event, context)
  File "/var/task/handler.py", line 240, in lambda_handler
    return handler.handler(event, context)
  File "/var/task/handler.py", line 345, in handler
    result = self.run_function(app_function, event, context)
  File "/var/task/handler.py", line 270, in run_function
    result = app_function(event, context) if varargs else app_function()
  File "/var/task/main.py", line 81, in update_results
    my_async_task(search_serialized)
  File "/var/task/zappa/async.py", line 421, in _run_async
    capture_response=capture_response).send(task_path, args, kwargs)
  File "/var/task/zappa/async.py", line 170, in send
    self._send(message)
  File "/var/task/zappa/async.py", line 249, in _send
    Message=payload
  File "/var/runtime/botocore/client.py", line 314, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/var/runtime/botocore/client.py", line 586, in _make_api_call
    api_params, operation_model, context=request_context)
  File "/var/runtime/botocore/client.py", line 641, in _convert_to_request_dict
    api_params, operation_model)
  File "/var/runtime/botocore/validate.py", line 291, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
```

## Expected Behavior
Invoking the task should not fail. Message argument sent to botocore must be a string, not bytes.

## Actual Behavior
We can't invoke the task using a SNS channel, because botocore expect a string, not bytes.

## Steps to Reproduce
1. use the code pasted above
2. `zappa invoke dev main.my_endpoint`
3. See the result.

## Your Environment

[...]

PR #1324, opened 2018-01-01
fixing botocore.exceptions.ParamValidationError

Debetux:
A fix for issue #1323.


Debetux:
Commit #1331 is a better fix for this issue.

Issue #1332, opened 2018-01-04
Enhancement request: async execution for a non-defined function


michelorengo:
## Context
My use case is to be able to execute a function ("task") using the async execution in a different lambda. That lambda has a different code base than the calling lambda. In other words, the function ("task") to be executed is not defined in the calling lambda.
The async execution lets you specify a remote lambda and remote region but the function (to be executed) has to be defined in the code. The request is to be able to simply provide a function name as a string in the form of <module_name>.<function_name>.
This obviously does not work for the decorator. It works only using "zappa.async.run".

## Expected Behavior
The below should work:
`
from zappa.async import run

run(func="my_module.my_function", remote_aws_lambda_function_name="my_remote_lambda", remote_aws_region='us-east-1', kwargs=kwargs)
`

## Actual Behavior
The function/task path is retrieved via inspection (hence requires a function type) by "get_func_task_path"

## Possible Fix
This is a bit hackish but is the least intrusive. I'll make a PR but I'm thinking of:
`
def get_func_task_path(func):

    """
    Format the modular task path for a function via inspection if param is
    a function. If the param is of type string, it will simply return it. 
    """

    if isinstance(func , (str, unicode)):
        return func

    module_path = inspect.getmodule(func).__name__

    task_path = '{module_path}.{func_name}'.format(
                                        module_path=module_path,
                                        func_name=func.__name__
                                    )

    return task_path
`

klaith25:
I  have exactly the same use-case. +1

klaith25:
Tried working around this issue by calling LambdaAsyncResponse directly but found that the send method also tried to import the module:

  File "/var/task/zappa/async.py", line 301, in run_message
  func = import_and_get_task(message['task_path'])
  File "/var/task/zappa/async.py", line 456, in import_and_get_task
  app_module = importlib.import_module(module)

so the proposed solution probably won't work.

michelorengo:
I did a PR and it worked for me. The PR [https://github.com/Miserlou/Zappa/pull/1333] was done a while ago and I did not follow the code base since so you may need to adjust the code edits.

PR #1333, opened 2018-01-04
Be able to provide the path of the task as a string

michelorengo:
## Description
This is to be able to provide the function/task path as a string (and not use the inspection)

## GitHub Issues
#1332 1332


klaith25:
Any idea on when this will be merged?

Issue #1350, opened 2018-01-13
Running zappa with no parameters causes exception


OverlordQ:
Running `zappa` command with no other arguments causes exception

## Context
Running inside venv with zappa installed

## Expected Behavior
Help message appears

## Actual Behavior
```
$ zappa
Oh no! An error occurred! :(

==============

Traceback (most recent call last):
  File "/home/ovrlrdq/.pyenv/versions/3.6.4/envs/venv364zappa/lib/python3.6/site-packages/zappa/cli.py", line 2610, in handle
    sys.exit(cli.handle())
  File "/home/ovrlrdq/.pyenv/versions/3.6.4/envs/venv364zappa/lib/python3.6/site-packages/zappa/cli.py", line 505, in handle
    self.dispatch_command(self.command, stage)
  File "/home/ovrlrdq/.pyenv/versions/3.6.4/envs/venv364zappa/lib/python3.6/site-packages/zappa/cli.py", line 520, in dispatch_command
    if not self.vargs['json']:
KeyError: 'json'

==============

Need help? Found a bug? Let us know! :D
File bug reports on GitHub here: https://github.com/Miserlou/Zappa
And join our Slack channel here: https://slack.zappa.io
Love!,
 ~ Team Zappa!
```

## Your Environment

* Zappa version used: 0.45.1
* Operating System and Python version: Debian, Python 3.6.4
* The output of `pip freeze`: https://hastebin.com/igepoveked.makefile

dfurtado:
Added a PR for this https://github.com/Miserlou/Zappa/pull/1368

scoates:
Python 2 vs 3 problem (doesn't happen in 2.7). @dfurtado's PR fixes it.

Miserlou:
Fixed via PR.

PR #1368, opened 2018-01-27
Fixed when running zappa with no parameters

dfurtado:
Fixed issue when running Zappa with no arguments. First, there was no checking to verify if any command was passed as an argument in the command-line and later on the dispatch_command function we were trying to get the value of 'json' in self.vargs causing an unhandled KeyError
exception. Now, if no argument is specified, the help will be displayed to the users.

See reported issue here: https://github.com/Miserlou/Zappa/issues/1350


scoates:
Looks good to me (tested). Python 2 vs 3 problem, as well (doesn't happen in 2.7).

abargnesi:
Simple fix and improves usability. LGTM.

PR #1381, opened 2018-02-07
feat: adding settings for mapping aws lex bot events

### LINKED ISSUE #1377: support AWS Lex Bot events
## Context
Currently zappa supports excuting django [AWS events](https://github.com/Miserlou/Zappa#executing-in-response-to-aws-events). It doesn't recognise the Lex bot event. It explicitly checks for `'Records'` inside the request. I want to organise all Lex bot functions inside the django app itself. So that I will have access to database and others. [Here](https://docs.aws.amazon.com/lambda/latest/dg/eventsources.html#eventsources-lex) is a sample Lex event source format

## Possible Fix
add new section to `events ` settings where we can configure the `intent's hook`.
```js
"events": [
      {
        // The function to execute
        "function": "mailer.tasks.send_emails",
        // When to execute it (in cron or rate format)
        "expression": "rate(5 minutes)"
      },
      {
        "function": "lexbot.handlers.book_appointment.handler",
        "event_source": {
          // intent's ARN : arn:aws:lex:region:accountId:
          "arn": "arn:aws:lex:<region>:<account-id>:intent:<intent-name>:$LATEST",
          // a list of configured invocations. possible values are from [FulfillmentCodeHook , DialogCodeHook]
          "events": [
            "DialogCodeHook"
          ]
        }
      }
    ],
  
```
 I will create an MR with support for this. It would not require much changes since there are similar invocation code inside handler. But we should arrive at a definitive way to get the function configuration from zapp_settings.json

## Your Environment
* Zappa version used: 0.45.1
* Operating System and Python version: Python 3.6

jnoortheen:
feat: handle incoming event from bot intent

resolves #1377

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,6 +532,21 @@ Similarly, for a [Simple Notification Service](https://aws.amazon.com/sns/) even
        ]
 ```
 
+For configuring Lex Bot's Intent triggered events
+```javascript
+	"bot_events": [
+        {
+            "function": "lexbot.handlers.book_appointment.handler",
+            "event_source": {
+                "arn": "arn:aws:lex:us-east-1:029992932068:intent:TmpTestLexEventNames:$LATEST", // optional. In future it will be used to configure the intent
+            	"intent":"intentName", // name of the bot event configured
+            	"invocation_source":"DialogCodeHook", // either FulfillmentCodeHook or DialogCodeHook
+            }
+        }
+	]
+ 
+```
+
 Events can also take keyword arguments.
 ```javascript
        "events": [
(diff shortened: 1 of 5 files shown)


Miserlou:
This actually looks okay, could do with a test.. can somebody confirm working?

jnoortheen:
I will add tests to it. I've tested the configuration on actual AWS environment and able to get results from lambda. Only thing is, I need to configure the initialisation hook the to the lambda function created by zappa. I've added that as a todo thing in future in docs.

jnoortheen:
@Miserlou I've added a test the PR.

PR #1382, opened 2018-02-07
Update regions in which API Gateway and Lambda are available

jneves:
Honestly, I was just annoyed about the warning messages when deploying to sa-east-1.


mcrowson:
Thanks!

PR #1390, opened 2018-02-12
Move task decorator async decision from import time to call time.

paulclarkaranz:
Move the `@task` decorator's decision about whether to run the wrapped
function immediately or dispatch it asynchronously from import time to call
time. Previously the `lambda_function_name` was being fetched from the
environment when the to-be-async function was being imported, making testing
the async nature of the function more difficult in a test setup where the
module was already imported.

During testing it can be useful to ensure that the application under test
performs when tasks are executed either blocking or asynchronously. Although
it could be argued that if we are testing that a third party library does the
right thing we are testing the wrong thing, or that we should do integration
testing in a docker-like environment. But I thought this change couldn't hurt
anyway.

Edit: amended to remove python3-isms.

### DIFF
--- a/tests/tests_async.py
+++ b/tests/tests_async.py
@@ -60,3 +60,34 @@ class TestZappa(unittest.TestCase):
     def test_sync_call(self):
         funk = import_and_get_task("tests.test_app.async_me")
         self.assertEqual(funk.sync('123'), "run async when on lambda 123")
+
+    def test_async_call_with_defaults(self):
+        """Change a task's asynchronousity at runtime."""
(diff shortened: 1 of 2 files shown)


paulclarkaranz:
`tests.tests_placebo.TestZappa.test_handler` is failing on my local machine with "KeyError: hello" but it is failing without this change too.

PS: Normally if I split production requirements and test requirements I have `-r requirements.txt` in `test_requirements.txt`. I don't know if you didn't do that on purpose or don't care but I installed the test requirements then tried to run the tests and it complained about not being able to import boto.

PR #1404, opened 2018-02-21
Support option for providing log config file

soloman1124:
Hi, it is essential in our use case to have consistent logging format. We'd like to this by suppling our own log config file if possible.


mcrowson:
Do you mind making an issue for this so we can discuss the implementation there?

soloman1124:
@mcrowson I have created a issue for it: https://github.com/Miserlou/Zappa/issues/1405

Issue #825, opened 2017-05-07
UnicodeDecodeError at zappa init


peteskii:
## Context

 
I read that Python 3 is now supported by AWS Lambda and Zappa, but when I try zappa init with default settings I get the below error:

It looks like this is a Flask application.
Oh no! An error occurred! :(

==============

Traceback (most recent call last):
  File "c:\users\pharm\documents\alexa\repo_stats\venv\lib\site-packages\zappa\cli.py", line 2352, in handle
    sys.exit(cli.handle())
  File "c:\users\pharm\documents\alexa\repo_stats\venv\lib\site-packages\zappa\cli.py", line 427, in handle
    self.init()
  File "c:\users\pharm\documents\alexa\repo_stats\venv\lib\site-packages\zappa\cli.py", line 1383, in init
    matches = detect_flask_apps()
  File "c:\users\pharm\documents\alexa\repo_stats\venv\lib\site-packages\zappa\utilities.py", line 143, in detect_flask_apps
    lines = f.readlines()
  File "c:\users\pharm\documents\alexa\repo_stats\venv\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 308: character maps to <undefined>

## Steps to Reproduce

Following this tutorial, but with Python 3: https://developer.amazon.com/blogs/post/8e8ad73a-99e9-4c0f-a7b3-60f92287b0bf/new-alexa-tutorial-deploy-flask-ask-skills-to-aws-lambda-with-zappa

Error occurs at zappa init with defaults selected
## Your Environment

Windows 10
Python 3.6

Miserlou:
This is a Windows-specific bug. 0.41.2 may have it fixed?

fferroni:
Hi,

I get the same error using

Windows 7
Python 3.5.2
zappa 0.41.3

The app I use is saved in UTF-8 encoding,

```
from flask import Flask
app = Flask(__name__)


@app.route('/')
def hello():
    return 'Hello, from Zappa!\n'

if __name__ == '__main__':
    app.run()
```

So, doesn't look like it was fixed in 0.41.2 :-(

Miserlou:
Can't repro and I don't have a windows machine to test on. Could be fixed by adding the coding header to files maybe?

mcdallas:
I am getting exactly the same error.

Windows 10
Python 3.6
[...]

PR #1406, opened 2018-02-21
Adding Flask app detection encoding

Noah-Wagner:
## Description
Added explicit UTF-8 encoding to the detection of Flask apps.
 

## GitHub Issues
[Issue 825](https://github.com/Miserlou/Zappa/issues/825)

### DIFF
--- a/zappa/utilities.py
+++ b/zappa/utilities.py
@@ -143,7 +143,7 @@ def detect_flask_apps():
 
             full = os.path.join(root, filename)
 
-            with open(full, 'r') as f:
+            with open(full, 'r', encoding='utf8') as f:
                 lines = f.readlines()
                 for line in lines:
                     app = None


Noah-Wagner:
Python 2 doesn't have an encoding parameter for open() and this broke that build.

mcrowson:
I think 'utf-8' (not 'utf8') is the default encoding anyhow, so I'm not sure this would fix anything.

Noah-Wagner:
I believe the default encoding on Unix platforms is UTF-8 while a quick test on my Windows machine gives me a default encoding of "cp1252" [Wiki link](https://en.wikipedia.org/wiki/Windows-1252).

So I think we need this to explicitly be a UTF-8 file open/read, but Python 2 doesn't support explicit encodings in this method.

Noah-Wagner:
Also [this](https://docs.python.org/3/library/codecs.html#standard-encodings) table is leading me to believe that 'utf8' is an okay alias for 'utf-8', but you're probably right that the correct encoding name is best practice.

Setting this encoding explicitly on my Windows/Python3 dev machine fixed the issue I was having with 
`zappa init`

PR #1411, opened 2018-02-22
Allow non-route53 domains to re-certify

paulclarkaranz:
When certifying an existing custom domain name don't check to see if it has a route53 record if route53 is disabled.

`get_domain_name()` was returning None because it couldn't find the domain in the route53 zone and so `create_domain_name()` was being called instead of `update_domain_name()` and I was getting an exception with message

> An error occurred (BadRequestException) when calling the CreateDomainName
> operation: The domain name you provided already exists.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1410,6 +1410,49 @@ class TestZappa(unittest.TestCase):
         finally:
             sys.stdout = old_stdout
 
+    @mock.patch('troposphere.Template')
+    @mock.patch('botocore.client')
+    def test_get_domain_respects_route53_setting(self, client, template):
+        zappa_core = Zappa(
+            boto_session=mock.Mock(),
+            profile_name="test",
+            aws_region="test",
+            load_credentials=False
+        )
+        zappa_core.apigateway_client = mock.Mock()
+        zappa_core.route53 = mock.Mock()
+
+        # Check it returns valid and exits early
+        record = zappa_core.get_domain_name('test_domain', route53=False)
+        self.assertIsNotNone(record)
+        zappa_core.apigateway_client.get_domain_name.assert_called_once()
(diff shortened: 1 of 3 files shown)


paulclarkaranz:
I added a test for `Zappa.get_domain_name()` because coveralls was complaining about it.

Miserlou:
> I added a test for Zappa.get_domain_name() because coveralls was complaining about it.

My man!

tamiel:
Looks like it introduced a regression when creating the domain with zappa certify

```python
==============

  File "/virtualenv/lib/python3.6/site-packages/zappa-0.45.1-py3.6.egg/zappa/cli.py", line 2679, in handle
    sys.exit(cli.handle())
  File "/virtualenv//lib/python3.6/site-packages/zappa-0.45.1-py3.6.egg/zappa/cli.py", line 504, in handle
    self.dispatch_command(self.command, stage)
  File "/virtualenv/lib/python3.6/site-packages/zappa-0.45.1-py3.6.egg/zappa/cli.py", line 611, in dispatch_command
    manual=self.vargs['manual']
  File "/virtualenv/lib/python3.6/site-packages/zappa-0.45.1-py3.6.egg/zappa/cli.py", line 1838, in certify
    route53=route53
TypeError: create_domain_name() got an unexpected keyword argument 'route53'

==============
```

I reverted to the commit before this patch ( commit 8dc76dde215bdf7a382675cf633006237b16f66c ) and it worked without error.

paulclarkaranz:
Yeah... that's a typo, sorry. The kwarg is being passed to the wrong function. And the mock used in the integration test doesn't care what args are passed. I've got a fix now I'm just trying to work with `mock.create_autospec` to make the test actually fail for this kind of thing.

tamiel:
@paulclarkaranz thanks for the quick fix!

Issue #1360, opened 2018-01-23
Enhancement request: Introduce Payload Compression options (AWS API Gateway)


ignacioHermosilla:
## Context
Enhancement request: Introduce new available feature for AWS API Gateway:  [Payload Compression for an API ](https://aws.amazon.com/about-aws/whats-new/2017/12/amazon-api-gateway-supports-content-encoding-for-api-responses/?sc_channel=em&sc_campaign=Launch_RN20180110&sc_publisher=aws&sc_medium=em_&sc_content=launch_la_nontier1&sc_country=&sc_geo=&sc_category=mult&sc_outcome=launch&mkt_tok=eyJpIjoiTlRGa01tSTJPV1prTlRSaCIsInQiOiI4ZXVJMDk3ZmxaMm5oVGlDY21QUXZ3MnVyQ0tkZlFNM2UyanFmVUk4cjhLYUxSQklkNXhnaTZVVmZLNlZLQkowYUZ6VWs2c2RNaFpZWTl6NFo4dTZQWHk5R1FobXhkaENOMDVSTGJoWDh1SzJcLzJFOEU3RzhoSkJDUlhxSm5NTGEifQ%3D%3D)

## Expected Behavior
Be able to enable/disable the Payload Compression and set the [minimumCompressionsSize](https://docs.aws.amazon.com/apigateway/api-reference/resource/rest-api/#minimumCompressionSize) property related to that option.

## Actual Behavior
Unsupported

## Possible Fix
I can create a pull request with the implementation but I would like to discuss the names first.
My proposal are:

- payload_compression: True/False (default False)

- payload_minimum_compression_size: In bytes (0 to 10485760)

rgov:
I think the documentation should reference security trade-offs with regard to the [CRIME attack](https://en.wikipedia.org/wiki/CRIME).

PR #1420, opened 2018-02-26
API Payload Compression

bcongdon:
## Description
 
* Addresses #1360 -- adds ability to enable payload compression on Lambda Rest APIs
* New setting `payload_compression` for enabling/disabling payload compression
* New setting `payload_minimum_compression_size` to set the threshold for compressing the payload

Payload compression is a property of the REST api -- like binary support -- so it has to be enabled/disabled independently from API stages

## GitHub Issues

#1360

### DIFF
--- a/README.md
+++ b/README.md
@@ -830,7 +830,7 @@ to change Zappa's behavior. Use these at your own risk!
         "log_level": "DEBUG", // Set the Zappa log level. Can be one of CRITICAL, ERROR, WARNING, INFO and DEBUG. Default: DEBUG
         "manage_roles": true, // Have Zappa automatically create and define IAM execution roles and policies. Default true. If false, you must define your own IAM Role and role_name setting.
         "memory_size": 512, // Lambda function memory in MB. Default 512.
-        "payload_compression": False, // Whether or not to enable API gateway payload compression (default: False)
+        "payload_compression": true, // Whether or not to enable API gateway payload compression (default: true)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -830,6 +830,8 @@ to change Zappa's behavior. Use these at your own risk!
         "log_level": "DEBUG", // Set the Zappa log level. Can be one of CRITICAL, ERROR, WARNING, INFO and DEBUG. Default: DEBUG
         "manage_roles": true, // Have Zappa automatically create and define IAM execution roles and policies. Default true. If false, you must define your own IAM Role and role_name setting.
         "memory_size": 512, // Lambda function memory in MB. Default 512.
+        "payload_compression": False, // Whether or not to enable API gateway payload compression (default: False)
```
Miserlou: This should be `false`, not `False`.

Miserlou:
Nice! Small doc change please. 

Is there any reason not to enable this by default?

bcongdon:
Hmm... I guess it does seem sensible to enable by default.

It [looks like](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-receive-response-with-compressed-payload.html) the compression is only done if an appropriate `Accept-Encoding` header is set, so it should be safe.

rgov:
See my comment in #1360:

> I think the documentation should reference security trade-offs with regard to the [CRIME attack](https://en.wikipedia.org/wiki/CRIME).

I am not sure it should be turned on by default in light of this attack.

PR #1429, opened 2018-03-02
Add IOpipe to "Related Projects" section

kolanos:
## Description

We're big fans of Zappa at @iopipe. Our agent supports Zappa out-of-the-box:
https://github.com/iopipe/iopipe-python#zappa


scoates:
This seems fine to me. We already link Sentry. Feel free to revert if you think otherwise, @Miserlou.

PR #1434, opened 2018-03-07
Fix certifying a domain for the first time.

paulclarkaranz:
## Description
In #1411 I added the route53 kwarg to `core.Zappa.get_domain_name` but when I
actually went to use the arg in `cli.ZappaCLI.certify` I added it to the next
line down... This was because I had originally made the change in my project's
venv, just to see if it was as simple as it looked, and when it worked I
cloned the Zappa repo and reimplemented it. Obviously I messed that up.

As reported by @tamiel on #1411 the call to `core.Zappa.create_domain_name` was
throwing a TypeError because of the unrecognized keyword argument. This was
not getting caught in the integration tests around `certify` because the
mocked functions were taking variable arguments.

This commit fixes the original mistake and changes the mock in
`tests.TestZappa.test_certify_sanity_checks` to use `mock.create_autospec`
instead of a custom class so that it actually has the same function signatures
as a real instance would.

## GitHub Issues
No issue raised yet but the problem was reported here: https://github.com/Miserlou/Zappa/pull/1411#issuecomment-370967045

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -1285,28 +1285,19 @@ class TestZappa(unittest.TestCase):
                 # fails when it tries to inspect what Zappa has deployed.
                 pass
 
-            class ZappaMock(object):
-                def __init__(self):
-                    self.function_versions = []
-                    self.domain_names = {}
-                    self.calls = []
+            # Set up a core.Zappa mock and let us save some state about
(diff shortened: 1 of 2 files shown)


jxltom:
Thanks, it works with existing domain now

PR #1453, opened 2018-03-22
Fix run_function when function has type annotations (issue #1452)

### LINKED ISSUE #1452: Invoking function failed when function has type annotation
## Context / Env
- Python 3.6, in a virtualenv.
- Django 1.11 project
- Running zappa in production for a year

We started adding type annotation to our code base.

*task.py*
```
def run() -> None:
    print("hello")
```

When running `zappa invoke task.run`, it raises a `ValueError` in the lambda

*Traceback*
```
Function has keyword-only parameters or annotations, use getfullargspec() API which can support them: ValueError
  File "/var/task/handler.py", line 509, in lambda_handler
  File "/var/task/handler.py", line 240, in lambda_handler
    return handler.handler(event, context)
  File "/var/task/handler.py", line 345, in handler
    result = self.run_function(app_function, event, context)
  File "/var/task/handler.py", line 267, in run_function
    args, varargs, keywords, defaults = inspect.getargspec(app_function)
  File "/var/lang/lib/python3.6/inspect.py", line 1045, in getargspec
    raise ValueError("Function has keyword-only parameters or annotations"
ValueError: Function has keyword-only parameters or annotations, use getfullargspec() API which can support them
```

Potential fix is to use `getfullargspec` rather than `getargspec` in `handler.py` I will test it and submit a PR later today

rcatajar:
Resolves #1452 .

I've also fixed the travis builds

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,7 +11,7 @@ jmespath==0.9.3
 kappa==0.6.0
 lambda-packages==0.19.0
 pip==9.0.1
-python-dateutil>=2.6.1
+python-dateutil>=2.6.1, <2.7.0
 python-slugify==1.2.4
 PyYAML==3.12
 requests>=2.10.0
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -1,5 +1,6 @@
-import unittest
 from mock import Mock
+import sys
+import unittest
 from zappa.handler import LambdaHandler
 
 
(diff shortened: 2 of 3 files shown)


Miserlou:
Ah, this is an excellent idea.

rcatajar:
@Miserlou I'm done here, should be good to go, let me know.

rcatajar:
tests should be good now.

Issue #1188, opened 2017-10-22
Zappa "sometimes" not passing content type header


Speedrockracer:
I am building an api for and iOS app using Flask and zappa. Somehow flask won't parse my post request when I use the iOS app to post something. I have it deployed to aws and it is working fine for GET requests. When calling the api with postman everything also works fine. Running Flask locally also works (the headers are passed).

## Context
Here are the logs from zappa tail for the failing request made with iOS / Alamofire
```
[...]

PR #1468, opened 2018-04-05
Fix canonical headers

ajw0100:
## Description
This PR fixes a bug in `zappa.wsgi.create_wsgi_request` function that non-deterministically results in some HTTP headers not being titlecased and consequently not being associated with the correct key in the `environ` dict that is returned. For example, the `content-type` header may end up in  `environ['HTTP_CONTENT_TYPE']` instead of `environ['CONTENT_TYPE']` where a wsgi app expects it to be. The reason this happens is because the `headers` dict is modified while it is being iterated. This PR refactors the title casing logic into a utility function which creates and returns a new dict instead of modifying the existing dict in place. The utility function has the added benefit of being reusable and more easily tested.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1188


Miserlou:
tHis Is A gOoD pR

Issue #1402, opened 2018-02-20
Retain X number of lambda functions on update call


mcrowson:
## Context

AWS Lambda retains old copies of code. This makes it easier to rollback, but its also possible to reach the 75G cap of stored lambda functions if a CI pipeline is constantly updating lambda code.

We should add a key to the zappa_settings.json for `retained_versions` or something similarly named that takes an integer and when `zappa update` is called, we prune the older versions greater than the `retained_versions` value.

## Expected Behavior

Zappa cleans up old versions of lambda functions when the update command is called.

## Actual Behavior

No pruning of old versions happens.

harryd31:
I completely agree, actually ran into that issue last week. Had to manually delete previous versions.

drewstreib:
I do this in a very mundane way today, and manually prune wide ranges:

#!/bin/bash
for i in {201..300}; do
  aws lambda delete-function --function-name $1 --qualifier $i 
  echo "lambda $1 version $i was deleted"
done

party98:
Hey, I'd like to take this up. Please help me as to where should i start.

mcrowson:
I see it as a few changes:
1. Add an argument to the zappa_settings.json file (update readme to show that it is there, and its default)
2. Add an optional parameter to the [update_lambda_function](https://github.com/Miserlou/Zappa/blob/master/zappa/core.py#L1067) that takes the number of revisions to keep.
3. Get the revisions_to_keep setting from the stage config and pass it to all calls of the update_lambda_function from the [cli's update function](https://github.com/Miserlou/Zappa/blob/master/zappa/cli.py#L836)
4. After the [function gets updated](https://github.com/Miserlou/Zappa/blob/master/zappa/core.py#L1083), get the value from the settings then make a call to get the[ versions currently in lambda](http://boto3.readthedocs.io/en/latest/reference/services/lambda.html#Lambda.Client.list_versions_by_function) for this function.
5. Compare the number of functions desired and the number of functions in lambda and then make a call to [lambda to delete](http://boto3.readthedocs.io/en/latest/reference/services/lambda.html#Lambda.Client.delete_function) the oldest versions in excess of the desired amount.

That's how I'd approach it anyhow. I might be missing something, and if it possible to write a test for this please do that too.

party98:
Thanks. On it

party98:
@mcrowson Can you elaborate on what is stage config.

dmateos:
[...]

PR #1470, opened 2018-04-07
Add support to delete old versions of lambda functions.

party98:
## Description
Added settings flag to limit the number of versions for a function in lambda to avoid crossing the limit on max storage for lambda.
 

## GitHub Issues
Resolves https://github.com/Miserlou/Zappa/issues/1402

### DIFF
--- a/README.md
+++ b/README.md
@@ -847,6 +847,7 @@ to change Zappa's behavior. Use these at your own risk!
         "log_level": "DEBUG", // Set the Zappa log level. Can be one of CRITICAL, ERROR, WARNING, INFO and DEBUG. Default: DEBUG
         "manage_roles": true, // Have Zappa automatically create and define IAM execution roles and policies. Default true. If false, you must define your own IAM Role and role_name setting.
         "memory_size": 512, // Lambda function memory in MB. Default 512.
+        "num_retained_versions":10, // Indicates the number of old versions to retain for the lambda. If absent, keeps all the versions of the function.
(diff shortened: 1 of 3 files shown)


mcrowson:
Awesome! A few things I'd tweak if you don't mind.

1. We like to keep the values in the readme example with what the defaults are. So change the 10 to a null. 

2. Instead of protecting against bad input values in the core.py file, do so in the cli.py and tell users if they give you something unexpected. So after you get the value from self.stage_config, do the check for valid types there and exit with a helpful message if users have a bad value. Something like:
```python
if invalid_value_for_num_revisions:
    raise ClickException("Please supply either an integer or null for num_retained_versions in the zappa_settings.json.")
```
Then when you get to the core.py you can just check to see if a value exists for it as you already know you've got a valid integer. 

Great work!

party98:
Sure, I'll update the PR.

mcrowson:
Looks like it is breaking some tests: https://travis-ci.org/Miserlou/Zappa/jobs/363488651
```python
# Check for valid values of num_retained_versions
if type(self.num_retained_versions) not in [int,None]:
     raise ClickException("Please supply either an integer or null for num_retained_versions in the zappa_settings.json.")
    elif self.num_retained_versions<1):
        raise ClickException("The value for num_retained_versions in the zappa_settings.json should be greater than 0.")
```

mcrowson:
Awesome thanks!

dmateos:
Any idea when this will make it into PIP so i can use it on deployments?

Issue #1478, opened 2018-04-14
Pip 10 compatibility issues


jleclanche:
Pip 10 was just released and I'm now getting this with it:

```pytb

12:15:05 Traceback (most recent call last):
12:15:05   File "/var/lib/jenkins/workspace/Deploy-Twitch-EBS-Lambda/.tox/jenkins_build/venv-deploy/lib/python3.6/site-packages/zappa/cli.py", line 2610, in handle
12:15:05     sys.exit(cli.handle())
12:15:05   File "/var/lib/jenkins/workspace/Deploy-Twitch-EBS-Lambda/.tox/jenkins_build/venv-deploy/lib/python3.6/site-packages/zappa/cli.py", line 505, in handle
12:15:05     self.dispatch_command(self.command, stage)
12:15:05   File "/var/lib/jenkins/workspace/Deploy-Twitch-EBS-Lambda/.tox/jenkins_build/venv-deploy/lib/python3.6/site-packages/zappa/cli.py", line 549, in dispatch_command
12:15:05     self.update(self.vargs['zip'], self.vargs['no_upload'])
12:15:05   File "/var/lib/jenkins/workspace/Deploy-Twitch-EBS-Lambda/.tox/jenkins_build/venv-deploy/lib/python3.6/site-packages/zappa/cli.py", line 877, in update
12:15:05     self.create_package()
12:15:05   File "/var/lib/jenkins/workspace/Deploy-Twitch-EBS-Lambda/.tox/jenkins_build/venv-deploy/lib/python3.6/site-packages/zappa/cli.py", line 2171, in create_package
12:15:05     disable_progress=self.disable_progress
12:15:05   File "/var/lib/jenkins/workspace/Deploy-Twitch-EBS-Lambda/.tox/jenkins_build/venv-deploy/lib/python3.6/site-packages/zappa/core.py", line 595, in create_lambda_zip
12:15:05     installed_packages = self.get_installed_packages(site_packages, site_packages_64)
12:15:05   File "/var/lib/jenkins/workspace/Deploy-Twitch-EBS-Lambda/.tox/jenkins_build/venv-deploy/lib/python3.6/site-packages/zappa/core.py", line 751, in get_installed_packages
12:15:05     pip.get_installed_distributions()
12:15:05 AttributeError: module 'pip' has no attribute 'get_installed_distributions'
12:15:05 Build step 'Execute shell' marked build as failure
12:15:05 Finished: FAILURE
```

[...]

PR #1481, opened 2018-04-18
Pip10 compatibility

iamjohnnym:
## Description

Zappa fails due to changes in pip==10.0.0.  `pip` changed how they deal with `get_installed_distributions`.  Unfortunately, the changes require usage of `pip._internal`, which goes against standard but it definitely works.

We can leverage the new method while also ensuring previous versions work.
```python
try: # for pip >= 10
    from pip._internal.utils.misc import get_installed_distributions
except ImportError: # for pip <= 9.0.3
    from pip import get_installed_distributions
```

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/1478

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ hjson==3.0.1
 jmespath==0.9.3
 kappa==0.6.0
 lambda-packages==0.19.0
-pip==9.0.1
+pip>=9.0.1, <=10.0.0
 python-dateutil>=2.6.1, <2.7.0
 python-slugify==1.2.4
 PyYAML==3.12
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -165,13 +165,9 @@ class TestZappa(unittest.TestCase):
 
         with mock.patch('os.path.isdir', return_value=True):
             with mock.patch('os.listdir', return_value=['super_package']):
-                import pip  # this gets called in non-test Zappa mode
-                try:
-                    with mock.patch('pip._internal.utils.misc.get_installed_distributions', return_value=mock_pip_installed_packages):
-                        self.assertDictEqual(z.get_installed_packages('',''), {'super_package' : '0.1'})
-                except ImportError:
-                    with mock.patch('pip.get_installed_distributions', return_value=mock_pip_installed_packages):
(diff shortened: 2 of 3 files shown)


wobeng:
@iamjohnnym I think you are missed [module 'pip' has no attribute 'main'](https://github.com/Miserlou/Zappa/blob/a93a24e39ceb6b583886735c0409a0d3d6d74824/zappa/core.py#L406)

they made a refactoring, We can support old and new with

```
try:
    from pip import main as pipmain
except:
    from pip._internal import main as pipmain
```

cc @Miserlou

scoates:
I realize this is going for a quick workaround, but it's going to be super fragile. The pip folks [broke this in 9.0.2](https://github.com/pypa/pip/issues/5081), and they're likely to do it again, without notice, in the 10.x tree.

From [the docs](https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program):

> pip is a command line program. While it is implemented in Python, and so is available from your Python code via import pip, you must not use pip's internal APIs in this way.

The correct approach here is to use `subprocess` and shell out to the command line. Unfortunately.

iamjohnnym:
Did they?  I was testing it on 9.0.1 - 10.0.0 and all seemed good and well.  

I figured it would fragile but I was leaning on pinning the version range in the requirements, we'd be able to keep it functional until a better way is created.

I believe we can do this with `pkg_resources` instead of `subprocess`.

I have an idea and when I get a bit of time later today, I'll see about changing that PR to be a bit more elegant.

@wobeng I did miss that, I appreciate you pointing it out.  I'll look into that too.

iamjohnnym:
@wobeng I believe I have solved for the `pip.main()` issue.

@scoates I implemented `pkg_resources.WorkingSet()` in lieu of `pip`.  This should solve for current and future supported versions.

iamjohnnym:
@scoates any thoughts on this or do we need to do something else?

scoates:
I haven't tried it and won't be able to until at least mid-May. We're all kind of busy. /-:

iamjohnnym:
Unfortunate but understandable.  Let me know if/when you need me more about this issue.

Keep on kicking ass fellas.

PR #1492, opened 2018-04-27
Allow newest Werkzeug.

kimvais:
## Description
Flask 1.0 requires Werkzeug>=0.14 which is incompatible with Zappa requirement of Werkzeug==0.12

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,6 +19,6 @@ six>=1.11.0
 toml>=0.9.4
 tqdm==4.19.1
 troposphere>=1.9.0
-Werkzeug==0.13
+Werkzeug>=0.13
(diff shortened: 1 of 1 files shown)


jakul:
@Miserlou Can this get merged?

shreyashag:
Could someone please merge this?

It throws off installs using pipenv and flask.
I use zappa with that environment and i have to skip-lock currently, which fails the point of using pipenv.

gs-jackal:
@kimvais  resolve the conflict?

paulclarkaranz:
The requirements have already been changed to `Werkzeug==0.14` on master but there is no new release with that change on pypi yet, so this is still a problem.

PR #1538, opened 2018-06-19
Allow providing a JSONEncoder class

charlax:
## Description

This is allowing the user to provide a `JSONEncoder` class to customize how message are serialized. Can we useful if you're using `uuid` a lot for instance.

## GitHub Issues

This is a fairly trivial change, it's faster to just show the code.

I haven't written any test yet, opening this PR to get some feedback on the idea and general implementation direction.


charlax:
What do you think about this?

pztrick:
+1. I typically use a `uuid` and would like to avoid `TypeError: Object of type UUID is not JSON serializable`.

Issue #1504, opened 2018-05-11
Feature request: Include custom domain base path mapping on zappa settings


alexandrefvb:
## Context
When deploying an application using zappa there is no option to specify the base path mapping on API gateway custom domain configuration. With this limitation it is not possible to use the same domain to deploy different APIs changing only the  base path.

## Expected Behavior

If a user want to deploy an API using an existing domain in a different base path he specifies the base_path option in zappa settings stage configuration and zappa create/update base path mappings according to this configuration. 

## Actual Behavior

There is no option to make this configuration.

## Possible Fix

Include an option on stages configuration to specify the base path to use on custom domain mapping.
Change certify command to do the base path mapping configuration on AWS API Gateway.

## Steps to Reproduce

N/A

## Your Environment

N/A

cl1ent:
i think this issue can be closed. the feature has been merged in (see above).
just use the `base_path` option in your apps config. its also documented in the readme:


`"base_path": "your-base-path", // Optional base path for API gateway custom domain base path mapping. Default None.`

alexandrefvb:
It was resolved in PR #1505

PR #1543, opened 2018-06-21
Add base_path stripping and fix unsetting of base_path

madeddie:
Minor fixes to existing code.

- When removing/unsetting a base_path after previously having it
deployed, it would fail because the base_path would be None, instead of
''.
- When a base_path was set, the path wasn't stripped from the request
path, making (at least) Flask, non-functional without changing the Flask
code.
(borrowed from
https://github.com/logandk/serverless-wsgi/blob/master/wsgi.py#L77)

If a base_path is updated, both a zappa update and a zappa certify action have to
be taken, I'm not sure if there's already procedure/code for forcing one
or the other.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -2000,6 +2000,7 @@ class ZappaCLI(object):
         self.profile_name = self.stage_config.get('profile_name', None)
         self.log_level = self.stage_config.get('log_level', "DEBUG")
         self.domain = self.stage_config.get('domain', None)
+        self.base_path = self.stage_config.get('base_path', None)
         self.timeout_seconds = self.stage_config.get('timeout_seconds', 30)
         dead_letter_arn = self.stage_config.get('dead_letter_arn', '')
         self.dead_letter_config = {'TargetArn': dead_letter_arn} if dead_letter_arn else {}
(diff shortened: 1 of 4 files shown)


madeddie:
Fixes a tiny bug in #1504 (which should be closed)

madeddie:
Not quite sure how to fix:
```
module 'tests.test_wsgi_script_name_settings' has no attribute 'BASE_PATH'
```

Either I just add `BASE_PATH = None` to `tests/test_wsgi_script_name_settings.py` or I build sanity checking into `zappa/handler.py`, although that doesn't seem to happen for any other setting.

So, for now I'm choosing to edit the test data, although that's mostly against my principles.

madeddie:
Ow, I seem to be mistaken, some sanity checking is done, I'll try that first.

madeddie:
https://github.com/awslabs/aws-serverless-java-container/issues/34 actually makes a good case for having a stripBasepath flag or something, in case people _want_ the basepath to be included in the path.

madeddie:
todo:

- [ ] flag to turn using base_path on and off (make sure to also fix output of cli commands)
- [ ] tests

ssilverm:
This pull request works for me.  This allows me to do multiple "subfolder" apps on one custom domain.

I have tested this with Django2.0/Python3, Flask/Python3, and Flask/Python2 and was able to have 
`
mydomain.com/app1/
mydomain.com/app2/
mydomain.com/app3/
`
and they all worked.

My zappa settings required me to certify the domain for each app (this may not have been needed), and added these to my json:
`
        "base_path": "app1",
        "certificate_arn": "arn:aws:xxxxxxxxxxxxxxxxxxxx",
        "domain": "mydomain.com",
        "route53_enabled": false, # I'm running my own DNS
`

ssilverm:
@Miserlou any chance this can be merged in?

mcrowson:
Like it. How hard would tests be for this?

Issue #1547, opened 2018-06-25
zappa package with slim_handler=true doesn't include Werkzeug package (introduced v0.46.0)


sheats:
## Context
I thought I was running into issue #64 but turns out I found a regression introduced in v0.46 related to how the handler_venv is packaged and uploaded when using slim_handler.  

I created a small repo to show the problem:  https://github.com/sheats/zappa_issue
If you keep `zappa==0.46.0` (or 0.46.1) in `requirements.txt` and deploy the function the only thing that prints in the logs is:
```
Unable to import module 'handler': No module named 'werkzeug'
```
If you change requirements.txt to `zappa==0.45.1` and update the function the problem goes away.

When I unzipped the handler venv zip after running `zappa package` I only see the following:
```
drwx------   9 user  staff    288 Jun 25 06:27 .
drwxr-xr-x  16 user  staff    512 Jun 25 06:27 ..
-rwxr-xr-x   1 user  staff      0 Jun 25  2018 __init__.py
-rw-------@  1 user  staff  85824 Dec 31  1979 _sqlite3.so
-rw-r--r--   1 user  staff    605 Sep 13  2016 django_zappa_app.py
-rw-------   1 user  staff  23461 Dec 31  1979 handler.py
-rw-------   1 user  staff    143 Dec 31  1979 package_info.json
drwxr-xr-x  14 user  staff    448 Jun 25 06:27 zappa
-rw-r--r--   1 user  staff    467 Jun 25  2018 zappa_settings.py
```
## Expected Behavior
I expect the logs to show other errors or normal successful logs.

## Actual Behavior
handler.py tries to [import werkzeug on line 17](https://github.com/Miserlou/Zappa/blob/master/zappa/handler.py#L17) but fails because Werkzeug wasn't included in the handler_venv zipped package.

## Possible Fix
Don't know the code well enough to come up with a fix strategy yet.

## Steps to Reproduce
```
git clone git@github.com:sheats/zappa_issue.git
make reproduce && make tail
```
## Your Environment
Everything is in the repo I referenced.

kuait:
Same here! I just downgraded  to `0.45.1`

giovannicimolin:
Same here, downgrading to 0.45.1 solves the problem.

5bodnar:
Had the same issue and can confirm that downgrading to 0.45.1 solves it.

jneves:
@sheats @giovannicimolin @5bodnar can any of you try with the current zappa master branch?

[...]

PR #1548, opened 2018-06-26
Wait for pip's execution and actually check the return code

jneves:
## Description

When building the pip environment for the zip when using slim_handler, we're using `.poll()` which doesn't wait for the process to finish. `.wait()` would be the easier replacement, but it carries the risk of deadlocks, hence using `.communicate()`.

## GitHub Issues
I suspect this contributes to at least: https://github.com/Miserlou/Zappa/issues/1547


dev-brian:
+1, threw a bunch of warnings about the target directory already existing, but it deployed successfully. Thank you!

Issue #1583, opened 2018-08-08
Add tagging of lambda functions to cli update and deploy methods


kukui:
Provide a mechanism for easily tagging lambda functions when deploying or updating and display tags
in the status command. This is a generally useful feature we'd like to have at work so that we can quickly identify the deployed code using the hash which we add as a tag

Issue #860, opened 2017-05-17
Ability to provide additional tags (key-value pairs) (to lambda functions)


TammoHeeren:
Maybe this is already supported, but I was unable to find it.

Background:
I am trying to better understand the cost breakdown of my deployment. The Amazon billing console support grouping/filtering by service type (lambda, etc). This is not sufficiently detailed for me to understand what lambda function is creating the highest cost. They additionally support grouping/filtering by Tag. This would allow me to filter more specifically. 

Feature request:
Support ability to provide additional Tags (key-value pairs) upon function creation.

Tammo

Miserlou:
This is a good one, and I'd like COST MANAGEMENT in general to be a feature of Zappa.

PR #1584, opened 2018-08-08
Lambdat tagging

kukui:
This PR adds tagging of lambda functions when using zappa deploy and update
zappa deploy --lambda-tags '{"alpha":"beta"}'
zappa update --lambda-tags '{"hash":"commithash"}'

Relevant tickets
https://github.com/Miserlou/Zappa/issues/1583
https://github.com/Miserlou/Zappa/issues/860
https://github.com/Miserlou/Zappa/issues/700

### DIFF
--- a/tests/placebo/TestZappa.test_create_lambda_function_local/lambda.ListTags_1.json
+++ b/tests/placebo/TestZappa.test_create_lambda_function_local/lambda.ListTags_1.json
@@ -1,0 +1,10 @@
+{
+    "status_code": 200, 
+    "data": {
+      "Tags": {
+        "alpha": "beta",
+        "gamma": "delta",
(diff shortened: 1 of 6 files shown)


scoates:
Is there a reason this makes more sense as a command line argument instead of living in the `zappa_settings`? Unless I'm missing something, it sounds like it should be the latter…

kukui:
perhaps my lack of familiarity with the zappa way of doing things? I was looking for a way to tag functions and other aws assets when deploying or updating. I could preprocess a zappa_settings template add my tags and then then run zappa update or zappa deploy but to me it felt more sensible to tag them from the prompt. I don't really like the idea of littering my repository with settings files for each update.

Issue #1356, opened 2018-01-18
Detect Python Version of .pyc Files and Ignore if Incompatible


mcrowson:
## Context

Zappa will package up .pyc files instead of the .py files if .pyc files are present. This can cause a ImportError: bad magic number if the .pyc files were compiled using a different Python version than is running on Lambda.

## Expected Behavior

Zappa only packages up .pyc files if they are compatible with the python version specified in the 
"runtime" setting. If the .pyc file is incompatible with the desired runtime, package the .py file instead.

## Actual Behavior

Zappa packages any .pyc file that exists instead of the .py file.

## Possible Fix

Check the .pyc magic number against known values for the runtimes.

Getting the magic numberscd  of a pyc file.
```python
with open('somefile.pyc') as f:
    magic = f.read(2)
magic_int = struct.unpack("<H", magic)[0]
```

Known Magic Numbers:
https://github.com/google/pytype/blob/master/pytype/pyc/magic.py

## Steps to Reproduce

1. `zappa update` with incompatible .pyc files
2. get frustrated

Miserlou:
TIL Python magic number!

mcrowson:
Yea i had to learn the hard way looking at logs hah

jakul:
This has caused some confusing bugs for me in the past when I have editably linked various local codebases together, but one of them was last executed by a Python 2 interpreter and the rest by a Python 3. It also happens ~twice a month for other developers in my organisation

rgov:
Would it be preferable to not package up any existing .pyc files, but recreate them at packaging time with the correct version of the interpreter? You also don't know if any of the .pyc files are outdated, right?

scoates:
@rgov Aren't some packages distributed only as `.pyc`? Maybe not in practice, but I think that's one its selling features (can distribute opcode-compiled Python instead of the source).

rgov:
I don't know of any packages distributed as .pyc, and if they were, it would be a poor choice: For one, as mentioned, they would only work with one specific version of the CPython interpreter (forget PyPy, etc.). Secondly, .pyc can easily be [decompiled back into code](https://pypi.python.org/pypi/uncompyle6), so it's not an effective obfuscation method.

But, if this is an actual case that people encounter, Zappa should probably do something, like warn that the .pyc could not be rebuilt because the file was not found.

scoates:
I agree that a warning is in order. We outright disallow .pyc in our app, altogether:

```python
[...]

PR #1586, opened 2018-08-11
don't package .pyc files with the wrong magic number ; 

scoates:
closes https://github.com/Miserlou/Zappa/issues/1356

All of the necessary information is there. This patch prevents invalid-magic-number `.pyc` files from being packaged.

Testing is painful (commented in the issue).

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -26,6 +26,7 @@ import zipfile
 from builtins import bytes, int
 from distutils.dir_util import copy_tree
 from io import open
+import struct
 
 import requests
 from setuptools import find_packages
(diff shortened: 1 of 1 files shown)


mcrowson:
Just found these. Looks like Python has some built in methods for validating magic numbers. If we used this rather than the approach above it would check against the current venv rather than the known values for the specified runtime. So this PR is probably the right approach for us, but might be something worth using in the cpython source.

Py3 has a function to check it: 
https://github.com/python/cpython/blob/3.6/Lib/importlib/_bootstrap_external.py#L430

Py2 just had a helper script
https://github.com/python/cpython/blob/2.7/Tools/scripts/checkpyc.py

Currently have bad internet and can't really dig too much else. My only question is would it be better to try and use any native python magic number checking vs keeping a static list? I'm probably over thinking it.

PR #1587, opened 2018-08-15
Specify unit divisor for tqdm

gaul:
Previously, Zappa formatted in both SI and binary:

```
Uploading githubwatch-dev-1534354789.zip (14.8MiB)..
100%|###########################| 15.5M/15.5M [00:03<00:00, 4.54MiB/s]
```

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -937,7 +937,7 @@ class Zappa(object):
         try:
             source_size = os.stat(source_path).st_size
             print("Uploading {0} ({1})..".format(dest_path, human_size(source_size)))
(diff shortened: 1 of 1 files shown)


gaul:
This scales the size consistently from 15.5 to 14.8 but needs tqdm/tqdm#595 to change the suffix from M to MiB

gaul:
Abandoning this PR; while it would be nice to show progress as 7MiB/42MiB, the existing 7M/42M works fine.

gaul:
Obsolete.

PR #1594, opened 2018-08-22
Editing requirements.txt to remove upper version limit of 10.1.0.

### LINKED ISSUE #1588: Upgrade or lift pip max version requirement
## Context
Zappa cannot currently be installed with the most modern released version of pip, which right now it 18.0. Zappa has the restriction `pip<=10.1.0,>=9.0.1`. I do not know if zappa is actually incompatible with pip 18.0, but presumably a change in the tests would determine this pretty easily.

## Expected Behavior
You'd be able to install with modern pip.

## Actual Behavior

```
zappa 0.46.2 has requirement pip<=10.1.0,>=9.0.1, but you'll have pip 18.0 which is incompatible.
```

## Possible Fix
Changing https://github.com/Miserlou/Zappa/blob/master/requirements.txt#L13 to either `pip>=9.0.1, <=18.0`, or `pip>=9.0.1`

## Steps to Reproduce
if pip and zappa are installed:
`pip install -U pip`

or if zappa isn't installed yet:
`pip install -U pip; pip install zappa`

nixjdm:
## Description
The end result of that ticket (#1588) is either this or raising the upper version limit of pip to current pip, which is v18. (they changed their versioning scheme)

I preferred this approach for a handful of minor reasons, but they are minor IMO, so the other solution (including another upper limit) is fine with me if the reviewer prefers it, as long #1588 actually is fixed. Using pip 18 caused no changes in my tests locally on Python 2.7 and 3.6, and works for me.

## GitHub Issues
Fixes #1588

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,7 +10,7 @@ hjson==3.0.1
 jmespath==0.9.3
 kappa==0.6.0
 lambda-packages==0.20.0
-pip>=9.0.1, <=10.1.0
+pip>=9.0.1
 python-dateutil>=2.6.1, <2.7.0
 python-slugify==1.2.4
 PyYAML==3.12


CurroRodriguez:
👍 I've confirmed that this fixes the issue and causes no other problems on my system.

CurroRodriguez:
This is passing, why hasn't been merged and released? The problem is becoming a pain for every CI/CD pipeline that creates a new virtual environment and causes failures installing requirements.

swedishmike:
I've tested this on both Python 2.7 and Python 3.6 on a couple of different systems and I cannot find any issues with it either.

PR #1600, opened 2018-08-30
Update/response

kiarashm:
## Description
 
I was using the asynchronous functionality of zappa and wanted to be able to update the status of my executing 'task' in the dynamodb table created corresponding to the capture_response flag being set to true in the '@task' decorator. Currently, the status is just set to 'in progress', and is set to 'completed' when the task finished executing. I had a long-running task, and wanted to be able to get more accurate feedback in the progress of my task (what stage was it on). In order to do so, I needed to be able to retrieve the response_id of the lambda function in which the task was being executed in so I could use it inside the function to update its progress.

To solve my issue, I have passed in 'response_id' as a parameter to **kwargs when I call the lambda function corresponding to my asynchronous task. This value requires the developer to add a **kwargs argument in the async task being defined AND add in a flag which enables this feature (allow_update_status) when declaring the task decorator:
ex)
from zappa.async import task
@task(capture_response=True, allow_update_status=True)
def _asynch_task(foo, bar, **kwargs):
     print("TRYING TO GET THE INSTANCE ID")
     if 'response_id' in kwargs:
         response_id = kwargs['response_id']
         print(response_id)

For convenience I also added in a method into async.py which allows the developer to update the dynamodb table used to track async task status. The method simply takes in the response_id and the string to update the 'async_status' value to.
ex)
from zappa.async import task, update_async_response
@task(capture_response=True, allow_update_status=True)
def _asynch_task(foo, bar, **kwargs):
     print("TRYING TO GET THE INSTANCE ID")
     if 'response_id' in kwargs:
         response_id = kwargs['response_id']
         update_async_response(response_id, "UPDATED STATUS")

Note(*) I currently have print statements in the update_async_response method for logging purposes. These can be removed or made optional if i include a flag in the method.

Would love your feedback on this! Have tested for my own use cases and seems to do the trick! Thanks!

### DIFF
--- a/zappa/async.py
+++ b/zappa/async.py
@@ -129,7 +129,7 @@ class LambdaAsyncResponse(object):
     Base Response Dispatcher class
     Can be used directly or subclassed if the method to send the message is changed.
     """
-    def __init__(self, lambda_function_name=None, aws_region=None, capture_response=False, **kwargs):
+    def __init__(self, lambda_function_name=None, aws_region=None, capture_response=False, allow_update_status=False, **kwargs):
         """ """
         if kwargs.get('boto_session'):
             self.client = kwargs.get('boto_session').client('lambda')
@@ -154,6 +154,7 @@ class LambdaAsyncResponse(object):
             self.response_id = None
 
         self.capture_response = capture_response
+        self.allow_update_status = allow_update_status
 
 
     def send(self, task_path, args, kwargs):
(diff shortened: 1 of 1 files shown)


bmi2013:
Having the option in the asynchronous task function body to update the status in DynamoDB is necessary to support a number of use-cases. Good option to include in Zappa.

PR #1612, opened 2018-09-18
Use new method to avoid deprecation warning of pypandoc

felixonmars:
`DeprecationWarning: Due to possible ambiguity, 'convert()' is deprecated. Use 'convert_file()'  or 'convert_text()'.`


jneves:
Closing as we're not using pypandoc anymore.

Issue #1613, opened 2018-09-19
Support SQS events to trigger lambda


jneves:
## Context
Amazon launched SQS events as triggers for lambda.

https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

## Expected Behavior
If I add an event source like the one below it should work:

```
        "events": [{
            "function": "app.process",
            "event_source": {
                  "arn": "arn:aws:sqs:<region>:<account_id>:zappa-test",
               	  "batch_size": 10, // Max: 10
                  "enabled": true // Default is false
               }
	}]
```

## Actual Behavior
I get an error.

## Possible Fix
See PR following this issue.

PR #1617, opened 2018-09-20
Adds SQS events to zappa

jneves:
Fixes: #1613

## Description
Adds support to SQS events.

This version uses an event source based on kappa. It should be removed if  https://github.com/garnaat/kappa/pull/133 is merged in kappa.

## GitHub Issues
#1613

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,7 +458,7 @@ However, it's now far easier to use Route 53-based DNS authentication, which wil
 
 ## Executing in Response to AWS Events
 
-Similarly, you can have your functions execute in response to events that happen in the AWS ecosystem, such as S3 uploads, DynamoDB entries, Kinesis streams, and SNS messages.
+Similarly, you can have your functions execute in response to events that happen in the AWS ecosystem, such as S3 uploads, DynamoDB entries, Kinesis streams, SNS messages, and SQS queues.
 
 In your *zappa_settings.json* file, define your [event sources](http://docs.aws.amazon.com/lambda/latest/dg/invoking-lambda-function.html) and the function you wish to execute. For instance, this will execute `your_module.process_upload_function` in response to new objects in your `my-bucket` S3 bucket. Note that `process_upload_function` must accept `event` and `context` parameters.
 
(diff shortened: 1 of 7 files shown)


jneves:
Tested on 2.7 and 3.6. Relevant configuration: ```        "events": [{
            "function": "app.process_event",
            "event_source": {
                  "arn": "arn:aws:sqs:eu-west-1:539577909625:zappa-test",
               	  "batch_size": 10, // Max: 10
                  "enabled": true // Default is false
               }
	}]```

MetricMike:
Running great on several of my 3.6 functions that use a 2-stage zappa function for webhooks with bursty traffic (one public-facing to persist JSON to an SQS queue with a variable message timer to smooth out the traffic, one private-facing to process them into another API or database). This PR eliminates most of our empty receives and SQS polling code :champagne:

mcrowson:
Any way you can squash these commits?

README.md
```diff
@@ -556,6 +556,21 @@ Optionally you can add [SNS message filters](http://docs.aws.amazon.com/sns/late
...
+[SQS](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html) is also pulling messages from a stream.  At this time, [only "Standard" queues can trigger lambda events, not "FIFO" queues](https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html).  Read the AWS Documentation carefully since Lambda calls the SQS DeleteMessage API on your behalf once your function completes successfully.  
+
+```javascript
+       "events": [
+           {
+               "function": "your_module.process_messages",
+               "event_source": {
+                    "arn":  "arn:aws:sqs:us-east-1:12341234:your-queue-name-arn",
+                    "batch_size": 10, // Max: 10. Use 1 to trigger immediate processing
+                    "enabled": true // Default is false
```
mcrowson: Why is this needed? We don't have an enabled field for other event sources. If they are present, they are enabled.

README.md
```diff
@@ -458,7 +458,7 @@ However, it's now far easier to use Route 53-based DNS authentication, which wil
 
 ## Executing in Response to AWS Events
 
-Similarly, you can have your functions execute in response to events that happen in the AWS ecosystem, such as S3 uploads, DynamoDB entries, Kinesis streams, and SNS messages.
+Similarly, you can have your functions execute in response to events that happen in the AWS ecosystem, such as S3 uploads, DynamoDB entries, Kinesis streams, SNS messages, and SQS queues.
```
mcrowson: Do we need to update the default IAM stuff to include SQS perms?

jneves:
The easiest way would be doing it on merge.On 1 Oct 2018 15:42, MCrowson <notifications@github.com> wrote:Any way you can squash these commits?

README.md
jneves: No, it's a push event.

README.md
jneves: Probably isn't going to test without it and clean up if it works.

edgarroman: Strangely, none of the event sources have the 'enabled' knob built into the AWS configuration.  See 'Enabled' parameter at end of documentation: https://docs.aws.amazon.com/lambda/latest/dg/with-sqs.html

I suspect it's probably some sort of throttling mechanism that typical SQS event sinks have control over.  Agree that it's a little wonky, but I would recommend to leave it.

jneves: @mcrowson enabled is false by default on push events - we also have the same behaviour for the other push events (dynamodb and kinesis). Without enabled you need to turn it on in the console, which makes some sense to me if you want to control/test when a consumer is "consuming" information from a queue.

PR #1624, opened 2018-09-27
feat: adding support for AWS Connect as one of the event source for Z…

### LINKED ISSUE #1473: AWS Connect as an event source
## Context
As of now Zappa doesn't support AWS Connect as. I am creating a PR that will add this ability. 

## Possible Fix
There should be settings available for configuring AWS Connect as trigger

## Your Environment
* Zappa version used: 0.45.1
* Operating System and Python version: Manjaro, Python 3.6

jnoortheen:
fixes #1473

## Description
 - Adds ability to set AWS connect as an event trigger.
- add docs about this support

## GitHub Issues

#1473

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,8 +594,8 @@ For configuring Lex Bot's intent triggered events:
     ]
 
 ```
+Events can also take keyword arguments.
 
-Events can also take keyword arguments:
 ```javascript
        "events": [
             {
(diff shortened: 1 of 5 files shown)


jneves:
@jnoortheen when I asked you to eliminate the spacing changes in zappa/cli.py it was so it doesn't take hours to review over 500 lines of unneeded changes in that file and to avoid this becoming a discussion on spacing instead of about the feature added. The current PR is, as it stands, is just as unreviewable as the previous one :( - I was hoping I could do a review of this as I've been working on the SQS event support, so I know my way around this particular part of the code.

PR #1650, opened 2018-10-12
Introduce ability to specify a CloudFormation role (resolves #1649)

### LINKED ISSUE #1649: Zappa Should Accommodate Passing a RoleARN to CloudFormation
## Context
For security purposes, there are certain permissions contexts where CloudFormation needs to be invoked by Role A, but Role A does not have the permissions required to fully spin up the stack, or to assume Role B, which does. In this case, AWS lets Role A specify that CloudFormation itself should assume Role B when provisioning these resources.

## Expected Behavior
I should be able to specify a `stack_role_arn` value in the `zappa_settings.json` that triggers behavior to specify an alternate RoleARN in when working with the CloudFormation API in Boto. 

## Actual Behavior
This functionality isn't currently available in Zappa.

## Possible Fix
Add support for the field, and optionally pass the RoleARN value when it is set.

relwell:
When users provide the `stack_role_arn`, CloudFormation will assume that
role when provisioning, modify, or deleting the stack specified by Zappa.

From the [Boto3 CloudFormation docs](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html):

```
The Amazon Resource Name (ARN) of an AWS Identity and Access Management (IAM)
role that AWS CloudFormation assumes to roll back the stack. AWS CloudFormation
uses the role's credentials to make calls on your behalf. AWS CloudFormation always
uses this role for all future operations on the stack. As long as users have permission
to operate on the stack, AWS CloudFormation uses this role even if the users don't
have permission to pass it. Ensure that the role grants least privilege.

If you don't specify a value, AWS CloudFormation uses the role that was previously
associated with the stack. If no role is available, AWS CloudFormation uses a
temporary session that is generated from your user credentials.
```

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -116,6 +116,7 @@ class ZappaCLI(object):
     aws_kms_key_arn = ''
     context_header_mappings = None
     tags = []
+    stack_role_arn = None
 
     stage_name_env_pattern = re.compile('^[a-zA-Z0-9_]+$')
 
@@ -1462,11 +1463,8 @@ class ZappaCLI(object):
             # There literally isn't a better way to do this.
             # AWS provides no way to tie a APIGW domain name to its Lambda function.
             domain_url = self.stage_config.get('domain', None)
-            base_path = self.stage_config.get('base_path', None)
             if domain_url:
                 status_dict["Domain URL"] = 'https://' + domain_url
-                if base_path:
-                    status_dict["Domain URL"] += '/' + base_path
             else:
                 status_dict["Domain URL"] = "None Supplied"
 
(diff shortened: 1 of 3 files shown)


relwell:
I was looking to add some tests to this PR, but I'm having a little difficulty figuring out where I would be able to add them, given the current test structure. I'd definitely appreciate some guidance!

relwell:
Can someone help me get this change merged? Thanks

zappa/cli.py
```diff
@@ -1464,11 +1465,8 @@ def tabular_print(title, value):
             # There literally isn't a better way to do this.
             # AWS provides no way to tie a APIGW domain name to its Lambda function.
             domain_url = self.stage_config.get('domain', None)
-            base_path = self.stage_config.get('base_path', None)
```
jneves: Any idea why the base_path is being removed from here?

zappa/cli.py
```diff
@@ -1464,11 +1465,8 @@ def tabular_print(title, value):
...
             if domain_url:
                 status_dict["Domain URL"] = 'https://' + domain_url
-                if base_path:
```
jneves: And here?

zappa/cli.py
relwell: My guess is a bad merge, seeing as I had to remove `base_path` before to merge master into my branch. It looks like you might be taking this PR over -- in which case you're welcome to make that change yourself, or I can.

relwell:
@jneves let me know if these updates work for you -- thanks!

jneves:
Looks good to me. Just needs some user testing. Hoping to have some time later, but anyone else can take it for a run and report it back here.

Issue #1508, opened 2018-05-14
Multiple query fields with same name only returning last value


seandavi:
## Context

I have a Flask/Flask-RESTPlus (very simple) that has a query parameter, `fields`, that is described as:

```python
parser = api.parser()
parser.add_argument('q', type=str,
                    help='A [lucene-syntax](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html) query',
                    default='*')
parser.add_argument('from', type=int, help='The starting record to return (for pagination), zero-based', default=0)
parser.add_argument('size', type=int, help='The number of records to return (for pagination)', default=10)
parser.add_argument('fields', help="fields to include in output. See 'mapping' endpoints for list of available fields", action='append')
```

## Expected Behavior

When running locally and querying with:

```
http://localhost:5000/....fields=a&fields=b...
```

My code sees the fields as an array of fields. 

## Actual Behavior

When running on the API endpoint after deploying (and updating), I see only the last value of the `fields` parameter. 

## Steps to Reproduce

The issue only manifests itself once deployed, not in the "local" environment.

## Your Environment

* Zappa version used: 0.45.1
* Operating System and Python version: Python 3.6.4 :: Anaconda, Inc., MacOS
* The output of `pip freeze`:
```
$ pip freeze
aniso8601==3.0.0
argcomplete==1.9.2
base58==0.2.4
boto3==1.7.6
botocore==1.10.6
certifi==2018.4.16
cfn-flip==1.0.3
chardet==3.0.4
click==6.7
docutils==0.14
durationpy==0.5
elasticsearch==6.2.0
Flask==0.12.2
flask-restplus==0.10.1
future==0.16.0
hjson==3.0.1
idna==2.6
itsdangerous==0.24
Jinja2==2.10
jmespath==0.9.3
jsonschema==2.6.0
kappa==0.6.0
lambda-packages==0.19.0
MarkupSafe==1.0
placebo==0.8.1
python-dateutil==2.6.1
python-slugify==1.2.4
pytz==2018.4
PyYAML==3.12
requests==2.18.4
s3transfer==0.1.13
six==1.11.0
toml==0.9.4
tqdm==4.19.1
troposphere==2.2.1
Unidecode==1.0.22
urllib3==1.22
Werkzeug==0.14.1
wsgi-request-logger==0.4.6
zappa==0.45.1
```
* Link to your project (optional):
* Your `zappa_settings.py`: 
```
{
    "dev": {
[...]

PR #1659, opened 2018-10-17
Support for multivalue querystring parameters

antwan:
## Description
AWS released support for multi-value (aka array) querystring parameters in API Gateway earlier this year. This means one can pass querystring containing arguments with same name (treated as an array). e.g: `param=value&param=othervalue`

This PR adds support for multi-value querystring parameters into Zappa so they are transferred correctly to the WSGI handler.

Also adds tests for multivalue headers (see #1755)

PR #1674, opened 2018-10-22
Fix for #1508 Multiple query fields with same name only returning last value

jgissend10:
## Description
This commit checks the event for a `multiValueQueryStringParameters` dict and if it finds it expands out the query params. APGW now support it but keep the values in a separate dictionary.
https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#apigateway-multivalue-headers-and-parameters

The helper in the first commit is created so that multivalueheaders can also be implemented in another PR.

## GitHub Issues

#1508

### DIFF
--- a/zappa/utilities.py
+++ b/zappa/utilities.py
@@ -1,6 +1,8 @@
 import botocore
 import calendar
 import datetime
+import itertools
+
 import durationpy
 import fnmatch
 import io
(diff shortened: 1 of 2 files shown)


antwan:
Note `multiValueQuerystringParameters` is now always returned there is no need to check if it's present or not.

You should add these to your tests.
Also `urlencode(query)` can handle a dict of arrays if you pass `True` as second arg, which removes the need to pre-transform the dict.

See #1659 which is duplicate of this PR.

PR #1681, opened 2018-10-25
fix: exclude setting to be considered for editable packages as well

### LINKED ISSUE #1680: `exclude` setting doesn't work with packages installed as editable (from github, etc.,)
## Context
I have installed `lambda-packages` using its `git` repo URL. But this makes it to end up in the final zip package. As I have inspected the code, the editable packages are copied without considering the exclude setting. 

## Expected Behavior
`lambda_packages` folder should completely be not present in the final zip package.

## Actual Behavior
It ends up in the zip and making it more than 100MB of size.

## Possible Fix
I am creating a PR to consider the exclude setting for the editable packages copying as well.

## Your Environment

* Zappa version used: 0.47
* Operating System and Python version: Manjaro & 3.7

jnoortheen:
## Description
fixes #1680


jneves:
@jnoortheen Can you check the test failures, please?

jnoortheen:
merged latest changes.

zappa/core.py
```diff
@@ -624,7 +624,7 @@ def splitpath(path):
                 copytree(site_packages_64, temp_package_path, metadata = False, symlinks=False)
 
         if egg_links:
-            self.copy_editable_packages(egg_links, temp_package_path)
+            self.copy_editable_packages(egg_links, temp_package_path, ignore=shutil.ignore_patterns(*excludes))
```
ryancausey: Looks like the remaining CI build failures are because `excludes` isn't necessarily defined by this point because it is set within the `if minify:` blocks above. The quick fix would be to set `excludes = ZIP_EXCLUDES + exclude` before this line, but I don't know if it would be better to move setting `excludes` up and out of the nested conditionals into one spot.

jnoortheen:
I have moved the assignment to the outer block.

ryancausey:
@jneves Looks like all the tests are passing now.

ryancausey:
Is there anything preventing this from being merged at this point?

ryancausey:
@jneves is there something preventing this from being merged?

Issue #1540, opened 2018-06-20
Create Private Deployments


mcrowson:
## Context

Support the new AWS Feature for [private API Gateways](: https://aws.amazon.com/about-aws/whats-new/2018/06/api-gateway-supports-private-apis/)

## Expected Behavior

Private deployments are only accessible within the VPC

## Actual Behavior

Currently all deployments are public

## Possible Fix

- Add `private` key to the zappa_settings.json in readme with default of False
- Change the [deploy function](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-resource-policies-create-attach.html) to use the new private flag if supplied in the config.
- Create or attach a [resource policy](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-resource-policies-create-attach.html) for the private API to have access within vpc. This will likely require some other setting in the config file specifying the VPC it is allowed in. Would likely require an additional config value in zappa settings where the specified VPC is allowed. Not sure how complicated this part could get.

I might be missing something as I've just read through the docs/announcement. So jump in if there is some requirement or solution piece I've missed.

brylie:
From what I can tell, this feature would create a resource policy rule to [Allow private API traffic based on source VPC or VPC endpoint](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-resource-policies-examples.html#apigateway-resource-policies-source-vpc-example):

```
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": "*",
            "Action": "execute-api:Invoke",
            "Resource": [
                "arn:aws:execute-api:region:account-id:api-id/*"
            ]
        },
        {
            "Effect": "Deny",
            "Principal": "*",
            "Action": "execute-api:Invoke",
            "Resource": [
                "arn:aws:execute-api:region:account-id:api-id/*"
            ],
            "Condition" : {
                "StringNotEquals": {
                   "aws:SourceVpc": "vpc-2f09a348"
                }
            }
        }
    ]
}
```

jgissend10:
This is merged https://github.com/Miserlou/Zappa/pull/1820

PR #1719, opened 2018-12-03
Enable Private API GW via EndpointConfiguration

IlyaSukhanov:
## Description
This change makes EndpointConfiguration configurable. This, primarily,
enables for creation of Private API Gateway Endpoints, ones only
accessible from within a VPC. Documentation is updated with example
configuration of the API Gateway and Resource policy.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1540

## See also
https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-apis.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,8 @@
   - [Using Zappa With Docker](#using-zappa-with-docker)
   - [Dead Letter Queues](#dead-letter-queues)
   - [Unique Package ID](#unique-package-id)
+  - [Endpoint Configuration](#endpoint-configuration)
+    - [Example Private API Gateway configuration](#example-private-api-gateway-configuration)
 - [Zappa Guides](#zappa-guides)
 - [Zappa in the Press](#zappa-in-the-press)
 - [Sites Using Zappa](#sites-using-zappa)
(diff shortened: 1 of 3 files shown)


glassresistor:
Any thoughts on this?  I would love to have it merged in soon.

monkut:
Interested in this.. but still trying to figure out how to package our app with surrounding VPC setup/configuration..

jneves:
Looks good to me. Ideally needs someone to test it out for different python versions and report back.

IlyaSukhanov:
FWIW; tested this with 3.6, can test with 2.7 also

gpkc:
I was able to successfully test this on both Python 2.7 and 3.6.

However, I am getting a `ConnectionError` on every `zappa deploy` and `zappa update`. This happens because zappa does a `GET` to the endpoint after the deployment (see the `touch_endpoint` method).

This would not happen, I believe, if two conditions are met:
* You are connected to the VPC, therefore being able to access the private endpoint
* You enable the `Enable Private DNS Name` option when creating the endpoint, which allows you to call the API Gateway URL directly, instead of having to use the endpoint URL (see https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-private-apis.html#apigateway-private-api-test-invoke-url)

jneves:
Merged in #1820 - thanks a lot!

Issue #837, opened 2017-05-10
Allow passing environment variables from shell to lambda through Zappa


majackson:
Currently, the two supported methods of supplying environment variables to a lambda function through Zappa is explicitly naming the environment variables in the`environment_variables` settings in `zappa_settings.json`, or placing a JSON environment variables file in an S3 bucket that the AWS credentials used with Zappa have access to.

A third way which would be useful is if variables present in the shell running Zappa could be named in `zappa_settings.json`, and the values of these could be passed through to lambda. This would allow environment variables to be stored only in the environment Zappa is executed in, and negate the need for an additional S3 bucket solely for storing sensitive values.

An example of how this might appear in the `zappa_settings.json` file could be

```
{
    "production": {
        "project_name": "my-project"
        "django_settings": "my_project.settings",
        "environment_pass": ["SECRET_THING"]
        "environment_variables": {
            "PRODUCTION": "TRUE",
        }
}
```
In this case, when Zappa is executed in a shell with SECRET_THING set to "123" the Lambda function would be deployed configured with PRODUCTION set to "TRUE", and SECRET_THING set to "123".

This would mirror some parts of docker tooling (i.e. docker-compose) which allow for environment variables to either be explicitly set, or passed through from the parent shell.

fishi0x01:
I think in general it would be nice to allow env vars in multiple parts of the configuration 
```
{
    "production": {
        "project_name": "my-project"
        "django_settings": "${DJANGO_SETTINGS}",
        "environment_pass": ["${SECRET}"]
        "environment_variables": {
            "PRODUCTION": "${PRODUCTION}",
        }
}
```

That way the `zappa_settings.json` acts like a template and can be reused in multiple environments. It could also be placed under source control and shared without worries about secrets or too static configuration.

samlobel:
I think it would be helpful to be able to use a local file instead of a remote_env file, in exactly the same way. That way you could edit it easily, and just not check the sensitive parts into source control. Referenced by `local_env : "./relative/path/to/.env"`

Thoughts?

arkaitzmugica:
I think the best way to make this would be the one majackson says... doing it "the docker way".

kontrollanten:
This would be really great! I've done a workaround by creating a zappa_settings.json.tpl with env variables names like
```
   "environment_variables": {
     "DATABASE_PASSWORD": "$DATABASE_PASSWORD"
   }
```

[...]

PR #1731, opened 2018-12-15
Add a command for setting environment variables.

youcandanch:
## TODO
- [ ] Add tests around new command.
- [ ] Solicit feedback on naming.

## Description
This is a little bit of a divergence from what was brought up in #837, but I think it's another avenue for how to solve the problem as described in that issue. This PR introduces a new command: `zappa set-environment-variable`. I'm a little concerned that the name is over-verbose, but figured I'd err on the side of being as explicit as possible.

For our particular use case, this would be valuable in CI for updating variables we expect in each environment (things like the SHA associated with the release, the time of deployment, etc.), but I can imagine it'd be helpful for a few different reasons.

The command takes three arguments:
`--key`: the key of the environment variable to set
`--value`: the value to set the environment variable to
`--overwrite`: a boolean indicating that, if an environment variable _already exists_ with the provided key, it should be overwritten. In the absence of this flag, a `KeyError` will be raised. (I'm not that thrilled with using `KeyError`, but couldn't think of anything better.

Tested as working in 2.7 and 3.6 (and 3.7, with a quick `__init__.py` tweak, since 3.7 support is coming down the pipeline).

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/837

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -268,6 +268,26 @@ class ZappaCLI(object):
         )
 
         ##
+        # Set Environment Variable
+        ##
+        set_envvar_parser = subparsers.add_parser(
+            'set-environment-variable',
+            parents=[env_parser],
+            help='Set an environment variable for a running stage.'
+        )
+        set_envvar_parser.add_argument(
+            '-k', '--key', help='The name of the variable.'
+        )
+        set_envvar_parser.add_argument(
+            '-v', '--value', help='The value to set for the env var.'
+        )
+        set_envvar_parser.add_argument(
+            '-o', '--overwrite',
+            help='If provided, will overwrite an existing env var with the same key.',
+            action='store_true'
+        )
+
+        ##
         # Template
         ##
         template_parser = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


zappa/cli.py
```diff
@@ -1622,24 +1648,24 @@ def init(self, settings_file="zappa_settings.json"):
...
+
             click.echo(click.style("Invalid bucket name!", bold=True))
             click.echo("S3 buckets must be named according to the following rules:")
             click.echo("""* Bucket names must be unique across all existing bucket names in Amazon S3.
 * Bucket names must comply with DNS naming conventions.
 * Bucket names must be at least 3 and no more than 63 characters long.
 * Bucket names must not contain uppercase characters or underscores.
 * Bucket names must start with a lowercase letter or number.
-* Bucket names must be a series of one or more labels. Adjacent labels are separated 
+* Bucket names must be a series of one or more labels. Adjacent labels are separated
```
youcandanch: Gah, sorry for the auto linting. I'll yank this in another commit.

zappa/cli.py
```diff
@@ -618,6 +638,12 @@ def dispatch_command(self, command, stage):
             )
         elif command == 'shell': # pragma: no cover
             self.shell()
+        elif command == 'set-environment-variable': # pragma: no cover
```
MihaZelnik: name is really a little over-verbose imo. What if we do same as empire has it
```
- set                set env var
- unset              unset env var
- env                list env vars
```

Issue #754, opened 2017-03-29
Zipfile doesn't preserve symlinks


mbox:
## Context

 
If the source directories contain symlinks, then in the generated zipfile the symlinks are replaced by copies of the original file.

This is particularly bad when dealing with binary libraries which are typically large and have multiple symlinks pointing at them.

## Expected Behavior

Symlinks should be preserved e.g. zipping:

```
lrwxr-xr-x    17B 28 Mar 16:19 libgdal.so@ -> libgdal.so.20.1.3
lrwxr-xr-x    17B 28 Mar 16:19 libgdal.so.20@ -> libgdal.so.20.1.3
-rwxr-xr-x   69M 28 Mar 16:19 libgdal.so.20.1.3*
```

Should result in an archive that when unzipped generates the same files (and takes 69MB)
## Actual Behavior

Each symlink is replaced by a copy of the file, so the files above end up as 3 x 69MB files
## Possible Fix

The bug looks like it's in zappa.py where the zipfile is manually constructed but without handling symlinks specially. Zipfiles can contain symlinks - see https://gist.github.com/kgn/610907 for an example of how to include them
## Steps to Reproduce

1. In your project, create a file
2. Create multiple symlinks to this file
3. run `zappa package`
4. Unzip the generated package
5. Observe that the original symlinks have been replaced by copies

## Your Environment

* Zappa version used:
* Operating System and Python version:
* Link to your project (optional):
* Your `zappa_settings.py`: 

zappa 0.39.1, MacOS X 10.11, Python 2.7.13

Miserlou:
Good bug, please feel free to send a PR and a test.

scoates:
If this gets worked on, I request that we have a configuration option (or even command line argument/flag) to turn this on (or if breaking BC, to keep the existing behaviour). We currently rely on it.

/cc @jnhmcknight

timmartin:
I've got a pull request for this. It doesn't yet have a configuration option; is the requirement for this still current? I can add a configuration option, but it wasn't immediately obvious to me why it would be useful / how it would be used, and I don't want to complicate the interface unnecessarily.

If the need for symlinks is limited to some libraries that have special requirements, it might be OK to just have it as an on / off switch. But if we rely on not preserving symlinks in mainline cases, then we'll need to do something more sophisticated (e.g. specifying subtrees in which to preserve symlinks) since it will not able to be turned on for the entire package.

jnhmcknight:
[...]

PR #1732, opened 2018-12-16
Preserve symlinks when packaging files into the zip file

timmartin:
## Description
Symlinks in the source directories are created as symlinks in the zip file, rather than deploying multiple copies of the same file.

## GitHub Issues

- #754 Zipfile doesn't preserve symlinks

### DIFF
--- a/README.md
+++ b/README.md
@@ -909,6 +909,7 @@ to change Zappa's behavior. Use these at your own risk!
         "payload_compression": true, // Whether or not to enable API gateway payload compression (default: true)
         "payload_minimum_compression_size": 0, // The threshold size (in bytes) below which payload compression will not be applied (default: 0)
         "prebuild_script": "your_module.your_function", // Function to execute before uploading code
+        "preserve_symlinks": false, // Symlinks to files within the source directory are preserved in the deployment package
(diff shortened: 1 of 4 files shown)


jneves:
@timmartin if the symlink is pointing to a file that is not being copied into the lambda by other code, does this work? If so, can you add a specific test for that case?

timmartin:
@jneves Good point, it doesn't cope with that. I patched it, but then I realised that this still doesn't cope with the case of symlinks to absolute paths that are outside the directory being packaged (which won't be broken when the link is tested, but will be broken later on when the zip is extracted).

I'll take another run at this later.

timmartin:
This should now deal with the case of symlinks to files outside the directory. I still need to implement the configuration switch before this can be merged, though.

jneves:
@timmartin any news?

timmartin:
Sorry, I should have commented on my last commit. The most recent commit adds the missing configuration switch, so I think this is ready to merge (or at least, ready to review) now. Feedback welcome, of course.

timmartin:
Is there any chance of getting this merged? I believe it's been ready to merge for 3 months now.

timmartin:
OK, nobody seems to be interested in this, I'll close it.

PR #1746, opened 2019-01-03
fix: Decoding issue for windows in zappa init

penguindustin:
## Description

Some windows machines can't decode the title text from `zappa init` causing it to fail. This change should allow for a safe alternative text to be displayed in those cases.

## GitHub Issues
[#825](https://github.com/Miserlou/Zappa/issues/825)

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -1551,12 +1551,25 @@ class ZappaCLI(object):
             raise ClickException("This project already has a " + click.style("{0!s} file".format(settings_file), fg="red", bold=True) + "!")
 
         # Explain system.
-        click.echo(click.style(u"""\n███████╗ █████╗ ██████╗ ██████╗  █████╗
+        # On windows, the following text fails to decode causing zappa init to fail.
+        # If this fails, fall back on a "safe" but still fancy title text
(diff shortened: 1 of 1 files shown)


zappa/cli.py
```diff
@@ -1551,12 +1551,25 @@ def init(self, settings_file="zappa_settings.json"):
...
+        # If this fails, fall back on a "safe" but still fancy title text
+        # Related: https://github.com/Miserlou/Zappa/issues/825        
+        try:
+            click.echo(click.style(u"""\n███████╗ █████╗ ██████╗ ██████╗  █████╗
 ╚══███╔╝██╔══██╗██╔══██╗██╔══██╗██╔══██╗
   ███╔╝ ███████║██████╔╝██████╔╝███████║
  ███╔╝  ██╔══██║██╔═══╝ ██╔═══╝ ██╔══██║
 ███████╗██║  ██║██║     ██║     ██║  ██║
 ╚══════╝╚═╝  ╚═╝╚═╝     ╚═╝     ╚═╝  ╚═╝\n""", fg='green', bold=True))
+        except :
```
jneves: Can you make the except more explicit?

penguindustin: Added specifying the decode error in the new commit

Issue #1716, opened 2018-11-30
Support for Lambda layers


jgissend10:
Lambda now supports layers and custom run times: https://aws.amazon.com/blogs/aws/new-for-aws-lambda-use-any-programming-language-and-share-common-components/

With this the lambda libs could be a layer and no longer a dep that has to be deployed.

jhorman:
Has anyone attempted to manually create a base layer, and then `exclude` as much as possible when deploying?

deejvince:
I used zap package to prepare a clean Zappa archive, I had to move the content of that archive inside a /python sub directory to support where lambda layers look for the layer content - and the. Uploaded the new zip as a layer manually.
I then created a new function referencing the new layer and was able to run a flask app in the lambda environment while kicking it off from the Zappa handler in the layer.
Wish there was a setting to make package or even deploy directly to lambda layer

farridav:
Implementing the ability to include layers (#1752 ) is a great start, though beyond that, perhaps automatically packaging up all project dependencies into a separate layer, and hashing a Profile/requirements file to see if the layer needs republishing, would ensure that most deployment artifacts are lightweight..

aralmpa:
Up untill today I have been using a fork of Zappa ([bahoo's](https://github.com/bahoo/Zappa/)) to apply layers to my Zappa deployed Lambda and it was working. I thought that the release 0.50 - [merged pull request #2027](https://github.com/Miserlou/Zappa/commit/379c0bf822eef91faccaa57fcf0111be21430cb1)  - provided Lambda layers support but a deployment I made just now does **not** add **any** layers to the function. Is this a bug?

jneves:
Probably. Can you share a config I can test with, please?

aralmpa:
Yes of course - this is the zappa_settings.json I use:
```
{
  "dev": {
    "aws_region": "eu-west-1",
    "django_settings": "persefoni.settings",
    "project_name": "persefoni",
    "runtime": "python3.6",
    "memory_size": 1024,
    "s3_bucket": "persefoni-zappa-deployment",
    "keep_warm": false,
    "endpoint_configuration":["REGIONAL"],
    "slim_handler": false,
    "touch": false,
    "layers": ["arn:aws:lambda:eu-west-1:472486572044:layer:geodjango-spatial-libraries-aralmpa:2"],
    "exclude": [
      "boto3*",
      "botocore*"
    ],
    "environment_variables": {
[...]

PR #1752, opened 2019-01-11
Add lambda layers support

piotrdrazikowski:
## Description
Lambda layers support
Added `layers` keyword in zappa_settings.json, put a list of ARN's there and zappa will add them to your function.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1716

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -117,6 +117,7 @@ class ZappaCLI(object):
     aws_kms_key_arn = ''
     context_header_mappings = None
     tags = []
+    layers = None
 
     stage_name_env_pattern = re.compile('^[a-zA-Z0-9_]+$')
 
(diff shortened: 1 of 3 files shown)


zappa/cli.py
```diff
@@ -757,7 +758,8 @@ def deploy(self, source_zip=None):
                 memory_size=self.memory_size,
                 runtime=self.runtime,
                 aws_environment_variables=self.aws_environment_variables,
-                aws_kms_key_arn=self.aws_kms_key_arn
+                aws_kms_key_arn=self.aws_kms_key_arn,
+                layers=self.layers
```
jgissend10: You pass this into the create, but you also need to pass it into the self.zappa.update_lambda_configuration function call in this file

jedestep:
@jgissend10 @piotrdrazikowski I made the change to `zappa update` and added some tests in #1842 . I'm planning on using layer support in our dev environment soon, so if there are other issues I'll report back.

farridav:
Keen to see this in, a good first iteration of layer support within the framework, needs another rebase though

CurroRodriguez:
When is this planned to be released? Any particular things that are awaiting testing?

Issue #1755, opened 2019-01-15
Error in handler when using multi-value headers


jedestep:
## Context

If an API gateway or ALB is configured to use multi-value headers, the handler module is unable to find the request headers and breaks in two places:
1. when setting script name based on the Host header (this is all that's seen in production because the lambda crashes here)
2. when forwarding the wsgi request via werkzeug, because no dict called 'headers' is included.

## Expected Behavior
Headers should all be combined into a single dict called 'headers'.

## Actual Behavior
Lambda crashes ignominiously with this error: `'NoneType' object has no attribute 'get'`.

## Possible Fix
Simple, just combine `multiValueHeaders` and `headers` into a single dict. I have a PR for this I have been working on.

## Steps to Reproduce

1.
2.
3.

## Your Environment

* Zappa version used:
* Operating System and Python version:
* The output of `pip freeze`:
* Link to your project (optional):
* Your `zappa_settings.py`:

paulclarkaranz:
There is actually two other PRs for this already #1659 #1675. One of them already has tests.

jedestep:
Correct me if I'm wrong but #1659 doesn't appear to support multi-valued headers in handler? #1756 does also include tests.

jedestep:
Any other thoughts on this?

jedestep:
Two PRs relating to this issue have been merged, so closing.

PR #1756, opened 2019-01-15
#1755 support for multi-value headers

jedestep:
## Description
This PR adds support for API gateways/ALBs that have multi-value headers enabled.

## GitHub Issues
#1755

### DIFF
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -86,7 +86,8 @@ class TestZappa(unittest.TestCase):
         lh = LambdaHandler('tests.test_wsgi_script_name_settings')
         event = {
             'headers': {
-                'a': 'b'
+                'a': 'b',
+                'z': 'q'
             },
             'multiValueHeaders': {
                 'a': ['c'],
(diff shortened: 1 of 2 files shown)


grjones:
There seem to be multiple long standing issues and PRs for this.  Is this the prevailing one?  And if so, what needs to happen to get this merged in?  I'm happy to help any way I can.

jedestep:
@grjones there was mention of the other PRs on [the issue](https://github.com/Miserlou/Zappa/issues/1755), but this PR seems to be the most complete (resolves the problem in the handler directly and has tests). No word yet from a contributor on where things are headed.

jneves:
I like it, I'm reviewing all the PRs now, but I like how this one looks. Another user trying it out and reporting on it would help in getting merged.

youcandanch:
@jneves I'll take it for a spin a bit later. We'll likely need to adjust a bit when #1730 lands, as MVH support for ALB is _slightly_ different, but this'll give us a foundation to go off.

youcandanch:
So this supports MVH but not multi-value querystrings, which is an important distinction to make. I think if we're going to try to support one, we should support both; it seems feasible that maybe we could combine this with https://github.com/Miserlou/Zappa/pull/1659?

youcandanch:
@jedestep https://github.com/youcandanch/Zappa/commit/3ae50c8e75f7e7ea589f0639087d262d1b0a3e95 adds multi value query string support and adjusts headers on the WSGI response to ensure that ALB works when we add it. Gonna add tests tonight, then we should be clear to merge this. It's been tested against both APIGW and ALB.

jedestep:
@youcandanch I agree that #1659 is a separate and also useful addition and having both merged would be ideal. I appreciate you taking the time to help out with this!

PR #1757, opened 2019-01-15
allow pyyaml>=3.13 in order to avoid vulnerabilities in pyyaml 3.13

dtenenba:
## Description
 

Change dependency on pyyaml from `==3.13` to `>=3.13`.

I have a couple of projects in GitHub that depend on Zappa and use `pipenv` to manage dependencies. GitHub has started sending me security alerts saying that there is a [vulnerability](https://nvd.nist.gov/vuln/detail/CVE-2017-18342) in that version of PyYAML and to upgrade to >= 4.02b1. 

This is impossible to do because of Zappa's explicit dependency on `pyyaml==3.13`. So I just changed the `==` to `>=`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,7 @@ lambda-packages==0.20.0
 pip>=9.0.1
 python-dateutil>=2.6.1, <2.7.0
 python-slugify==1.2.4
-PyYAML==3.13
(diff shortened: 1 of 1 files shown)


jneves:
Thanks, merged as part of #1786 (just merged).

Issue #1710, opened 2018-11-20
Lambda now supports python 3.7


jgissend10:
https://aws.amazon.com/blogs/compute/python-3-7-runtime-now-available-in-aws-lambda/

adulion:
could we get an ETA on zappa supporting 3.7?

chris-erickson:
See #1581 for one attempt.

purificant:
Proposed patch: https://github.com/Miserlou/Zappa/pull/1762

jgissend10:
https://github.com/Miserlou/Zappa/pull/1762 is merged!

Issue #1666, opened 2018-10-19
async is a reserved word in Python 3.7


kappa90:
## Context
In Python 3.7 async is a reserved word. This means that the zappa.async module cannot be imported anywhere, as it triggers a syntax error. I thought it was a pylint error at the beginning, but it's actually a py3.7 one, so the code won't run at all.

## Expected Behavior
I should be able to import the zappa.async module

## Actual Behavior
async module cannot be imported

## Possible Fix
Renaming the module, although I understand it would be a breaking change

## Steps to Reproduce
1. Install Python 3.7
2. Try to import zappa.async

## Your Environment
* Zappa version used: 0.47
* Operating System and Python version: Windows 10, Python 3.7
* The output of `pip freeze`:

aniso8601==3.0.2
argcomplete==1.9.3
astroid==2.0.4
boto3==1.9.27
botocore==1.12.25
certifi==2018.10.15
cfn-flip==1.0.3
chardet==3.0.4
Click==7.0
colorama==0.4.0
decorator==4.3.0
docutils==0.14
durationpy==0.5
Flask==1.0.2
Flask-GraphQL==2.0.0
future==0.16.0
graphene==2.1.3
graphene-pynamodb==2.1.0
graphql-core==2.1
graphql-relay==0.4.5
graphql-server-core==1.1.1
hjson==3.0.1
idna==2.7
isort==4.3.4
itsdangerous==0.24
Jinja2==2.10
jmespath==0.9.3
kappa==0.6.0
lambda-packages==0.20.0
lazy-object-proxy==1.3.1
MarkupSafe==1.0
mccabe==0.6.1
placebo==0.8.2
promise==2.1
pylint==2.1.1
pynamodb==3.3.1
python-dateutil==2.6.1
python-slugify==1.2.4
pytz==2018.5
PyYAML==3.13
requests==2.19.1
Rx==1.6.1
s3transfer==0.1.13
singledispatch==3.4.0.3
six==1.11.0
toml==0.10.0
tqdm==4.19.1
troposphere==2.3.3
typing==3.6.6
Unidecode==1.0.22
urllib3==1.23
validators==0.12.2
Werkzeug==0.14.1
wrapt==1.10.11
wsgi-request-logger==0.4.6
zappa==0.47.0

mberledgylabs:
Why would you need zappa in python3.7 for the current version? It's specified in readme that zappa works only with 2.7 or 3.6.

jgissend10:
[...]

Issue #936, opened 2017-06-19
async module should probably be renamed 


jleclanche:
In Python 3.7, `async` is a keyword and you will not be able to use it as a module name (like in an import).

Try out the deprecation warnings:

```
% PYTHONWARNINGS=all python3.6 -c 'async = 1'
<string>:1: DeprecationWarning: 'async' and 'await' will become reserved keywords in Python 3.7
```

bcongdon:
Maybe it could be changed to some variation of `asynctask` / `async_task`?

Miserlou:
Goddammit.

Maybe just `task` but I don't really like that either.

charlax:
`async_` ?

purificant:
Py37 is now officially supported on AWS Lambda, because of this there is now a compelling reason to also support 3.7 in Zappa. Proposed patch: https://github.com/Miserlou/Zappa/pull/1762

her0e1c1:
I think this can be close. Use `from zappa.asynchronous import task` instead :)

PR #1762, opened 2019-01-19
python3.7 support

purificant:
## Description

Since Nov 19, 2018 AWS Lambda Supports Python 3.7: https://aws.amazon.com/about-aws/whats-new/2018/11/aws-lambda-supports-python-37/

This PR enables support for Python 3.7 in Zappa.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1710
https://github.com/Miserlou/Zappa/issues/1666
https://github.com/Miserlou/Zappa/issues/936

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,8 @@ language: python
 python:
   - "2.7"
   - "3.6"
+  - "3.7"
+dist: xenial
 # command to install dependencies
 cache:
   - pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ dist: xenial
 cache:
   - pip
 install:
-  - "pip install --upgrade pip; pip install --upgrade setuptools; pip install -r test_requirements.txt; pip install -e git+https://github.com/django/django-contrib-comments.git#egg=django-contrib-comments; python setup.py install" 
+  - "pip install --upgrade pip; pip install --upgrade setuptools; pip install -r test_requirements.txt; pip install -e git+https://github.com/django/django-contrib-comments.git#egg=django-contrib-comments; python setup.py install"
 # command to run tests
 env:
   - TESTCASE=tests/tests_docs.py
@@ -17,7 +17,12 @@ env:
   - TESTCASE=tests/tests_placebo.py
   - TESTCASE=tests/tests_async.py
   - TESTCASE=tests/tests.py
-script: 
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+script:
   - nosetests $TESTCASE --with-coverage --cover-package=zappa --with-timer
     #  - coverage combine --append
 after_success:
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -781,6 +781,16 @@ class ZappaCLI(object):
 
         endpoint_url = ''
         deployment_string = click.style("Deployment complete", fg="green", bold=True) + "!"
+
+        if self.use_alb:
+            kwargs = dict(
+                lambda_arn=self.lambda_arn,
+                lambda_name=self.lambda_name,
+                alb_vpc_config=self.alb_vpc_config,
+                timeout=self.timeout_seconds
+            )
+            self.zappa.deploy_lambda_alb(**kwargs)
+
         if self.use_apigateway:
 
             # Create and configure the API Gateway
@@ -917,31 +927,27 @@ class ZappaCLI(object):
 
         # Register the Lambda function with that zip as the source
         # You'll also need to define the path to your lambda_handler code.
+        kwargs = dict(
+            bucket=self.s3_bucket_name,
+            function_name=self.lambda_name,
+            num_revisions=self.num_retained_versions
+        )
         if source_zip and source_zip.startswith('s3://'):
             bucket, key_name = parse_s3_url(source_zip)
-            self.lambda_arn = self.zappa.update_lambda_function(
-                                            bucket,
-                                            self.lambda_name,
-                                            key_name,
-                                            num_revisions=self.num_retained_versions
-                                        )
+            kwargs = kwargs.update(dict(
+                bucket=bucket,
+                s3_key=key_name
+            ))
+            self.lambda_arn = self.zappa.update_lambda_function(**kwargs)
         elif source_zip and not source_zip.startswith('s3://'):
             with open(source_zip, mode='rb') as fh:
                 byte_stream = fh.read()
-            self.lambda_arn = self.zappa.update_lambda_function(
-                                            self.s3_bucket_name,
-                                            self.lambda_name,
-                                            local_zip=byte_stream,
-                                            num_revisions=self.num_retained_versions
-                                        )
+            kwargs['local_zip'] = byte_stream
+            self.lambda_arn = self.zappa.update_lambda_function(**kwargs)
         else:
             if not no_upload:
-                self.lambda_arn = self.zappa.update_lambda_function(
-                                                self.s3_bucket_name,
-                                                self.lambda_name,
-                                                handler_file,
-                                                num_revisions=self.num_retained_versions
-                                            )
+                kwargs['s3_key'] = handler_file
+                self.lambda_arn = self.zappa.update_lambda_function(**kwargs)
 
         # Remove the uploaded zip from S3, because it is now registered..
         if not source_zip and not no_upload:
@@ -1098,6 +1104,9 @@ class ZappaCLI(object):
             confirm = input("Are you sure you want to undeploy? [y/n] ")
             if confirm != 'y':
                 return
+
+        if self.use_alb:
+            self.zappa.undeploy_lambda_alb(self.lambda_name)
 
         if self.use_apigateway:
             if remove_logs:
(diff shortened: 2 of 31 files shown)


zappa/core.py
```diff
@@ -648,8 +648,8 @@ def splitpath(path):
                 # This is a special case!
                 # SQLite3 is part of the _system_ Python, not a package. Still, it lives in `lambda-packages`.
                 # Everybody on Python3 gets it!
-                if self.runtime == "python3.6":
-                    print(" - sqlite==python36: Using precompiled lambda package")
+                if self.runtime == "python3.6" or self.runtime == "python3.7":
```
cclauss: if self.runtime in (“python3.6”, “python3.7”):

.travis.yml
```diff
@@ -2,6 +2,8 @@ language: python
 python:
   - "2.7"
   - "3.6"
+  - "3.7"
+dist: xenial
```
cclauss: This changes all test runs to xenial which is generally available on Travis but is still not the default distro.  A more conservative alternative would be to only run Python 3.7 on xenial:
```
matrix:
  include:
    - python: “3.7”
       dist: xenial
```

purificant: :+1: This would indeed be a more cautious approach.

I've opted for not using the build matrix because it requires specifying all combinations of `env` in the matrix, just adding a single include will only run the first `env`. This creates a lot of duplication in the yaml file and makes it significantly bigger. In addition, trusty is end of life April 2019, which may or may not affect it's use by Travis as the default VM.

Because of the above two reasons I've opted to switch to xenial by default. It gives us py37 and keeps the current yaml file mostly intact.

That being said, I'm happy to amend this approach and run py37 on xenial and continue running py 3.6 and 2.7 on trusty if necessary.

purificant:
Hi @cclauss, thank for you looking at this PR and your review. I've done the necessary amends.

willgdjones:
Looking forward to seeing this merged! Thanks for all the work.

.travis.yml
cl1ent: @purificant cant you use [global variables](https://docs.travis-ci.com/user/environment-variables/#global-variables) here?

purificant: @cl1ent It's the same issue when using matrix inside the env, only the first `env` is run for py37 and not all permutations. My sample code was:
```
language: python
matrix:
  include:
    - python: "2.7"
    - python: "3.6"
    - python: "3.7"
      dist: xenial
# command to install dependencies
cache:
  - pip
install:
  - "pip install --upgrade pip; pip install --upgrade setuptools; pip install -r test_requirements.txt; pip install -e git+https://github.com/django/django-contrib-comments.git#egg=django-contrib-comments; python setup.py install" 
# command to run tests
env:
  matrix:
    - TESTCASE=tests/tests_docs.py
    - TESTCASE=tests/test_handler.py
    - TESTCASE=tests/tests_middleware.py
    - TESTCASE=tests/tests_placebo.py
    - TESTCASE=tests/tests_async.py
    - TESTCASE=tests/tests.py
script: 
  - nosetests $TESTCASE --with-coverage --cover-package=zappa --with-timer
    #  - coverage combine --append
after_success:
  coveralls
notifications:
  slack: zappateam:TTJ0mfHunDK0IBweKkEXjGpR
```

Lenn4rt:
+1

cclauss:
There are 30 👍 on the last two comments but there are only 2 positive reviews on the top right of this page.  Instead of a 👍, please consider __Review changes__ because that will give project maintainers more confidence that ___you have read through the actual changes___ and believe that those changes improve the state of this project...

_Anyone_ can review any GitHub pull request:
1. Scroll up on this page and click on __Files changed__ just above the initial commit message.
2. Click on the green button __Review changes__
3. Click on __Comment__ or __Approve__ or __Request changes__

Thanks for taking the time to go through these steps which should help maintainers.

willgdjones:
Hi @cclauss, thanks for the suggestion. I have just gone through those steps now and have left my review. The changes (in my opinion) seem lightweight, and directly address the core problem of running Zappa with Python 3.7, namely the use of the `async` decorator which overrides the `async` functionality introduced in Python 3.7.

mcrowson:
Could we do a check for the python version and do an alias for async -> asyncrhonous for <3.7 versions so this wouldn't be breaking?

mcrowson:
Hrm, i'm seeing now that async is a file not a class. Hrm.... I wish there was a way this wouldn't break for <3.7

cclauss:
Python does __not__ allow:
```python
try:
    x
except SyntaxError:
    y
```

mcrowson:
Could we just have an async.py file with just `from zappa.asynchronous import *` and some comments as to why this exists. It will still raise the syntaxerror in 3.7, but will allow <3.7 to keep working without breaking existing code that imports from async.

cl1ent:
@mcrowson why don't you just use a fixed version zappa==0.47.1 and stick with py3.6? if you then decide to upgrade to py3.7 and the newer, you need to check your code anyway.

cclauss:
Some very large projects like [pytorch, cuda](https://github.com/pytorch/pytorch/pull/4999) faced this same issue.  That team merely selected a replacement word (“non_blocking” in their case) and then on a major version release, declared that there was a _breaking change_ required to keep using modern versions of Python.  It seems that this project will need to do something similar.

Thank you to the 12 who have given ✔️ [reviews](https://github.com/Miserlou/Zappa/pull/1762#issuecomment-462682060) of this PR at the upper right of this page.

mcrowson:
Let's say some amazing feature or critical bug fix gets released in 0.49 or some other future release. All projects would have to migrate to >=3.7 in order to get those, and that would be non-trivial in my case and for others as well.

cclauss:
Those projects do not need to move to Python 3.7 to use that new release of Zappa.  They can continue using their current version of Python but they must change their code from __Zappa.async__ —> __Zappa.asyncronous__.

jneves:
@cclauss giving context: it's rare for us to have a breaking change like this - we want to release this ASAP but being a breaking change has me and the rest of the core team (of which @mcrowson is part) discussing whether it should be done separately from the next release or not. It also makes the release harder for us and creates support requests in the near future. Anything that can reduce that and reduces the impact of the change will make its release faster.

ippeiukai:
If adding a small file is all needed for some limited backward compatibility that helps most of affected users, why not give it to them? It’s probably  easier than discussing the implication of breaking changes. (Perhaps with DeprecationWarning? Can `__import__` function avoid SyntaxError?)

Said that, I’m against putting off Python 3.7 support any longer because it needs breaking changes. It’s supported by Lambda for months now and many people here are frustrated having to wait for Zappa to support it.

ippeiukai:
> Could we just have an async.py file with just from zappa.asynchronous import * and some comments as to why this exists.

@mcrowson @purificant https://github.com/purificant/Zappa/pull/1

cclauss:
@jneves #936 was opened two years ago so don’t feel too rushed.

One solution would be to put out your release without fixing Python 3.7 compatibility.  That release shoul include a warning to users of a future breaking change.  Immediately following that release, make a _beta_ release that supports Python 3.7.  That would allow a small, dedicated group of users to _experiment_ with the solution before you open 3.7 compatibility to a broader audience in a follow-on release.

purificant:
> > Could we just have an async.py file with just from zappa.asynchronous import * and some comments as to why this exists.
> 
> @mcrowson @purificant [purificant#1](https://github.com/purificant/Zappa/pull/1)

Thank you @ippeiukai for [purificant#1](https://github.com/purificant/Zappa/pull/1)

If maintainers are happy with the extra code needed to test the imports from the old async file, this approach is an excellent compromise. However, I would recommend setting a schedule to delete `async.py` and the associated test suite along the lines of @cclauss's:

> One solution would be to put out your release without fixing Python 3.7 compatibility. That release shoul include a warning to users of a future breaking change. Immediately following that release, make a _beta_ release that supports Python 3.7. That would allow a small, dedicated group of users to _experiment_ with the solution before you open 3.7 compatibility to a broader audience in a follow-on release.

zappa/cli.py
```diff
@@ -1716,7 +1716,7 @@ def init(self, settings_file="zappa_settings.json"):
             env: {
                 'profile_name': profile_name,
                 's3_bucket': bucket,
-                'runtime': 'python3.6' if sys.version_info[0] == 3 else 'python2.7',
+                'runtime': get_venv_from_python_version(),
```
itamt: I think it's better to use `get_runtime_from_python_version`.
Because:
previous code allows only `'python2.7'` and `'python3.6'`
suggested code allows any `pythonX.X` according to `sys.version`

```suggestion
                'runtime': get_runtime_from_python_version,
```

cl1ent: doesn't `get_runtime_from_python_version()` only return `python2.7` or `python3.6`?
https://github.com/Miserlou/Zappa/blob/6a43ed0b06e5877e71e829686fad93f8e51b78c1/zappa/utilities.py#L173 seems to be the better way as it gets the version from `*sys.version_info`

purificant: @itamt Because of https://github.com/Miserlou/Zappa/blob/6a43ed0b06e5877e71e829686fad93f8e51b78c1/zappa/__init__.py#L3 and the exception that is raised if you run CLI using anything other than one of the supported versions, both functions [get_venv_from_python_version](https://github.com/Miserlou/Zappa/blob/6a43ed0b06e5877e71e829686fad93f8e51b78c1/zappa/utilities.py#L173) and [get_runtime_from_python_version](https://github.com/Miserlou/Zappa/blob/6a43ed0b06e5877e71e829686fad93f8e51b78c1/zappa/utilities.py#L176) will essentially produce identical output.

In my view, relying on `get_venv_from_python_version` is more sustainable in the long run as new python versions are enabled in the future and less code changes are needed to support them. However, as the code is right now, it does not matter which function is used.

As far as this PR is concerned, the proposed change here is to move from hand crafted logic `'python3.6' if sys.version_info[0] == 3 else 'python2.7'` to an existing function that produces similar output. There is further room to cleanup use of both functions and make the code more pythonic ( [There should be one-- and preferably only one --obvious way to do it.](https://www.python.org/dev/peps/pep-0020/) ), however, such changes are outside the scope of this PR.

purificant:
Good spot @kimbriancanavan, thank you.
And thank you @AndrewScibek for a speedy PR.

cclauss:
`self.manylinux_wheel_file_suffix = 'cp{}{}m-manylinux1_x86_64.whl'.format(*sys.version_info)`

jneves:
Thanks a lot to all, particularly @purificant!

purificant:
Thank you everyone for commenting, reviewing, testing, suggesting, adding code and merging this PR!

Issue #538, opened 2016-12-14
Dead link in top level readme + missing instructions on running tests


horenmar:
I am trying to make a PR against zappa and when looking for information about how to run tests, etc, I found this in the readme 
"If you are adding a non-trivial amount of new code, please include a functioning test in your PR. For AWS calls, we use the placebo library, which you can learn to use in the test writing guide."

where "in the test writing guide" is a link to nonexistent page `https://github.com/Miserlou/Zappa/blob/master/docs/README.md`.

Also `docs/` subfolder and top level readme seem to be out of sync.

Miserlou:
Good catch!

Issue #510, opened 2016-12-02
Load Project and Site-Packages from S3


mcrowson:
Currently, the entire application and site packages are uploaded to Lambda. This issue proposes uploading the project contents and deps to S3 instead, and having the zappa handler load the project files.

Proposed Changes:
- Zappa CLI: Zip up minimal handler application with Zappa and dependencies for lambda
- Zappa CLI: Send project zip to S3
- Handler: On handler initialization, load the project zip in S3 into /tmp/{project_name}
- Handler: Add the /tmp/{project_name} to sys path.

The first time the application runs it will likely be slower pulling from S3, but if the application uses the keep_warm appropriately, then the subsequent loads shouldn't have the same slowdown due to lambda keeping /tmp around between function calls.

The benefit of this change would be allowing much larger applications to run on Lambda.

Thoughts? Am I missing something?
Credits to @mathom for suggesting P3 via S3 this way.

mathom:
You probably want to check the file hash somehow (s3 etag) in case you push an update to s3 and need to pull down over an already unzipped project in a warm handler. Not sure if that's really an issue depending on how long your stuff lives but it might be a while with the keep warm setting.

It should be pretty obvious when the application is actually having to download and unzip in the timings, so it will be easy to keep an eye on.

mcrowson:
Thanks @machbio for wanting to help. Here is my fork for getting going. https://github.com/mcrowson/Zappa/tree/project-to-s3

There are a few things I'm working through now. 

-  We will need to keep the project in S3. The current zappa paradigm cleans up after itself in S3, so we might need a new settings for how many of these to retain and where.
-  As this approach is only needed for projects over 50 MB, we might want this feature to only take place if the size of the packaged zip is > 50 MB. This way smaller projects can avoid the additional startup time on cold-starts.

mathom:
A cool side effect of keeping your project in a bucket is you could move all the lambda management stuff over to cloud formation, too. Not sure if that's worth pursuing at the moment but it's def an option to simplify some of the update stuff.

machbio:
Sorry about doubting '/tmp' space guarantee - its pretty clear that '/tmp' space is 512MB as given in the Limit - http://docs.aws.amazon.com/lambda/latest/dg/limits.html#limits-list

> Ephemeral disk capacity ("/tmp" space)	512 MB

and also on the FAQ page for Lambda - https://aws.amazon.com/lambda/faqs/

[...]

PR #1765, opened 2019-01-20
[Docs] Clarify "Large Projects" section

gbrova:
This PR clarifies the documentation about large projects.

If my understanding is correct: 
 - Using `slim_handler` increases disk usage, not memory requirements.
 - We are still limited to a maximum size of 512 MB, because AWS restricts the size of /tmp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -991,7 +991,7 @@ You can also simply handle CORS directly in your application. Your web framework
 
 ### Large Projects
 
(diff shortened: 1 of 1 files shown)


cl1ent:
according to the docs, the limit for deployment package (zip) uploads from s3 is at 250mb (or more specific 262144000 bytes) and not 512mb. its even smaller for direct uploads.

see [here](https://docs.aws.amazon.com/lambda/latest/dg/limits.html) and [here](https://hackernoon.com/exploring-the-aws-lambda-deployment-limits-9a8384b0bec3)

gbrova:
Yes, that is true. 

However, if my understanding is correct, the purpose of the `slim_handler` flag is to overcome that restriction. 

Reading through #538, seems this is implemented by copying the zip to `/tmp`. So, we are still restricted on space, but the restriction is slightly more lax -- according to your first link and also the discussion in [510](https://github.com/Miserlou/Zappa/issues/510#issuecomment-265178651), the limit for `/tmp` seems to be 512MB. 

It would be nice if this limit was documented somewhere, because I initially (incorrectly) assumed that I can deploy arbitrarily large functions using `slim_handler`. 

Please let me know if I'm misunderstanding something.

PR #784, opened 2017-04-08
Fix for Zappa no longer creating Route53 records for AWS  issue #762

bitshark:
Original Author: Aneesh Kumar, anush0247
Applied to latest commit on Zappa master and tested by bitshark

This is a fix for Zappa issue #762.  Zappa is not creating and
persisting the appropriate CNAME records in AWS Route 53, when using
an AWS Hosted Zone with Let's Encrypt certificates (via zappa certify)

Zappa creates the Custom Domain Name properly in API gateway, and it
also properly creates the let's encrypt certification DNS records, but
it does not create and persist the correct Route 53 CNAME records for the app

This fix resolves the issue and has been tested locally to confirm the fix

Full details:
https://github.com/Miserlou/Zappa/issues/762
https://github.com/anush0247/Zappa/commit/8ea4ecf2cc70e3fcbc39824d12eb77fa82c38ff1

## Description
This fixes the Route53 certification issues described in issue #762.  It is a simple change, only two or three lines total across two files.   The original author is Aneesh, who's commit is linked above.  I've just ported it to master and submitted it as a pull request after confirming it resolves my issue as well.

## GitHub Issues
Issue #762

Issue #762, opened 2017-03-31
zappa certify no longer creates route53 records for letsencrypt certificates


codereverser:
## Context
With the recent update for custom SSL certificates support, route53 domains are not created when letsencrypt is used. 

 

## Expected Behavior

route53 records should be created if `route53_enabled` is True and `lets_encrypt_key` is set.
## Actual Behavior

route53 dns records are created only if custom SSL certificates are given. Also the documentation for `route53_enabled` says _"Have Zappa update your Route53 Hosted Zones when certifying with **a custom domain**"_ . So is this the expected behaviour? 
## Possible Fix

This is due to `update_route53_records` being called only when custom SSL certificates are used. Copying the logic for `update_route53_records` in `ZappaCLI.certify`  to `letsencrypt. get_cert_and_update_domain` should fix the issue. 
```python
letsencrypt.py (partial code)
                if not zappa_instance.get_domain_name(domain):
                    dns_name = zappa_instance.create_domain_name(
                        domain_name=domain,
                        certificate_name=domain + "-Zappa-LE-Cert",
                        certificate_body=certificate_body,
                        certificate_private_key=certificate_private_key,
                        certificate_chain=certificate_chain,
                        certificate_arn=None,
                        lambda_name=lambda_name,
                        stage=api_stage
                    )
                    if route53_enabled:
                        zappa_instance.update_route53_records(domain, dns_name)
```
## Steps to Reproduce

1. Set `route53_enabled` to `true`, `lets_encrypt_key` to your account key and remove any settings for using custom SSL certificates.
2. `zappa certify <env>`
3. API is deployed properly, custom domain is created in API gateway, but route53 records are not created. 

## Your Environment

* Zappa version used: **0.40.0**
* Operating System and Python version: MacOS Sierra , python 2.7.11
* Link to your project (optional): N/A
* Your `zappa_settings.py`: 
```json
{
  "dev": {
    "aws_region": "ap-southeast-2",
    "s3_bucket": "REDACTED-app",
    "django_settings": "server.settings",
    "remote_env_file": "dev.json",
    "remote_env_bucket": "REDACTED",
    "domain": "REDACTED",
    "lets_encrypt_key": "s3://REDACTED/account.key",
    "lets_encrypt_expression": "rate(15 days)",
    "timeout_seconds": 120,
    "use_precompiled_packages": true,
  }
}
```

jbcurtin:
[...]

PR #1780, opened 2019-02-15
fix certify route53 record creation with lets encrypt

kontrollanten:
solves #762

## Description
 
Create S3 routes after creating lets encrypt certificates with `zappa certify`. https://github.com/Miserlou/Zappa/pull/784 already has a PR for this, but it's without tests and currently there's conflicts.
I reached lets encrypts rate limit when testing in python 3.6, but it's verified in python 2.7.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/762
https://github.com/Miserlou/Zappa/pull/784


kontrollanten:
Okay. This went bad. I'll go home and do my job. I'll be back.

Issue #1792, opened 2019-02-28
S3 Bucket for Zappa is not encrypted by default.


dci-aloughran:
The S3 bucket used to host the zappa files is not encrypted.  S3 has recently been updated to allow default server-side encryption.
## Context
The ability to set S3 bucket to be encrypted by default adds a subtle but additional layer of security to your S3 buckets.

## Expected Behavior
S3 bucket should be created with default encryption turned on to AES256.

## Actual Behavior
S3 bucket is created with no encryption.

## Possible Fix
I've added the code that sets the default encryption on the bucket to SSE AES256.

## Steps to Reproduce

1. Create a new Zappa app using this code.
2. Create a new test project
3. See that the bucket is now set with "Default Encryption" on, in the Bucket Properties.

elcolie:
I propose input argument like `awscli` like `--sse` to do the job
Because not all people really need default encryption

![image (2)](https://user-images.githubusercontent.com/18206728/77390818-a7d79900-6dc9-11ea-904a-0f44f918e579.png)

cyrillay:
Same issue here. Currently we have to add the bucket encryption manually after if was created by Zappa. We'd like Zappa to create the bucket with encryption, either by default or with a configuration in `zappa_settings.json`.

PR #1793, opened 2019-02-28
Set bucket to have  SSE encryption by default

dci-aloughran:
Potential fix for #1792

## Description
Sets the encryption on the bucket to AES256 by default when it's created.

## GitHub Issues

#1792


youcandanch:
It'd be good to have a way to opt-out of encryption, if at all possible. While I get that having it by default is a great idea, there's probably a use case out there that requires unencrypted S3 assets. Can you add something to the CLI settings parser to allow that?

Issue #1715, opened 2018-11-29
Supprt for ALB as event source?


dmateos:
Just wondering if theres any insight to if this will work and if not if it can be made to in the future? :) 

As per here: 
https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html

This would eliminate some timeout issues we have using API Gateway.

PeteW:
API gateway, for my usage needs, imposes a deal-breaking 30-second hard limit on request execution time. While 99.9% of the methods meet this limit I simply cannot guarantee all times will fall within that time range now and future.

For this reason I moved away from API gateway and back to dedicated EC2 instances. The ALB route not only works around this 30-second issue but it presents a potentially graceful transition to switching to lambda.

I'm familiar with the zappa framework and would be happy to assist with implementation of support for ALB deployment. It is an architectural question for the core developers of this project. Is ALB deployment something they see as "in-scope" for zappa?

PeteW:
I'm new to zappa source code. I was under the notion that API Gateway was inextricably woven into the core. After reading/experimenting today I was pleasantly surprised how well-organized the api-gateway stuff is logically cordoned from lambda/s3/cli. 


The `use_apigateway` as an option set to `False` convinced me that while API Gateway is a major feature, the real power is in lambda management and execution.


In this PR I've delivered functionality where we leverage the already perfect zappa deployment tools and adjust the "handler" logic to detect if the caller is API Gateway or ALB and react accordingly. This code change enables zappa to work flask functions from ALB. API Gateway concepts such as "stages", "domains", "authentication" dont currently carry over. Also I didnt add the support for auto-provisioning ELBs and targeting (currently I'm convinced cloudformation doesnt yet support this sort of configuration yet, I have an open support ticket)


To get it working, you will need to:
* Set `use_apigateway` option to false (unless you want both APIGateway AND ELB which presumably should work but most people wouldnt want that)
* Run zappa deploy
* Manually establish your own ALB. Ensure the subnets align with your lambda function and the security groups are reachable
* Manually configure the targeting to route to the lambda function (I recommend doing this from EC2, not lambda so you can have human-readable names)

PeteW:
Bump: anyone around to review/merge this?

kennovak-zy:
I've been using this in my shop for a few weeks, and it's working well.  Super-useful!

kennovak-zy:
[...]

PR #1807, opened 2019-03-09
Application Load Balancer Support

youcandanch:
### Summary
This pull request introduces application load balancer support (ALB) for Zappa, as an alternative to API Gateway for web applications and APIs. Support has been added alongside the existing support for APIGW, and is handled on an opt-in basis using zappa_settings. This PR includes both the adjustments to `handler.py` to accomodate ALB's different event structure, and the additions to provisioning logic to both create and destroy the ALB infrastructure on `zappa deploy` and `zappa undeploy`.

The work for this began and was discussed in #1730 and #1715. 

### Why?
ALB has upsides that make it a compelling alternative to API Gateway for some use cases:
- There's no hard timeout limit. API Gateway has a 30 second timeout, which means services can't take full advantage of the 15 minute Lambda timeout. ALB timeouts can be set to any amount and match whatever's provided via the `timeout` parameter of zappa_settings.
- API Gateway can be incredibly cost-prohibitive for high throughput services, since billing is calculated by request. For web services receiving millions of events a day, using an ALB as the event source can result in some pretty serious cost savings (so far, about 100 dollars a day for one of our high throughput APIs).
- ALBs can be placed within a VPC and targeted privately; API GW _does_ support this using PrivateLink, but placing all logical resources within one's VPC may be an appealing design.

### Technical Decisions
- Troposphere doesn't support ALBs sending events to Lambdas yet, specifically because Cloudformation _also_ doesn't support that functionality. Since we can't rely on CF support, @PeteW implemented the provisioning and de-provisioning logic with boto3. This feels like a non-contentious decision, as the Lambda provisioning itself is _also_ using boto3, but it does diverge from the pattern of provisioning web event sources with Troposphere.
- We've created an alias on Lambda creation called `current-alb-version`. When using an ALB as an event source for Lambda, unless an alias is used, it loses permissions on each subsequent `zappa update`; even if the version the ALB TargetGroup is associated with is `$LATEST`. Using an alias allows us to have zero downtime 
- There's a new `zappa_settings` parameter for `alb_vpc_config`; it's likely that most developers will want to have separate subnets and security groups for their ALBs, and so using `vpc_config` and having both Lambdas and ALBs share this info seemed a little off. It also makes it very clear and explicit that a Zappa service is using ALB as a target.  
- Since #1802 landed, I've defaulted support for multi-value headers to be _on_ on the ALB, which matches API Gateway behaviour. I can't see a compelling reason to make this configurable, but could be swayed.

### Testing
I've been running this branch or permutations of it in production for a few weeks on a fairly high throughput service with no blips, as have other folks in #1730.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.pythonPath": "env/bin/python3"
-}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "env/bin/python3"
+}
--- a/README.md
+++ b/README.md
@@ -654,10 +654,9 @@ Putting a try..except block on an asynchronous task like this:
 @task
 def make_pie():
     try:
-    ingredients = get_ingredients()
-    pie = bake(ingredients)
-    deliver(pie)
-
+        ingredients = get_ingredients()
+        pie = bake(ingredients)
+        deliver(pie)
     except Fault as error:
         """send an email"""
     ...
@@ -670,8 +669,7 @@ will cause an email to be sent twice for the same error. See [asynchronous retri
 @task
 def make_pie():
     try:
-    """code block"""
-
+        """code block"""
     except Fault as error:
         """send an email"""
     ...
@@ -819,12 +817,12 @@ to change Zappa's behavior. Use these at your own risk!
 ```javascript
  {
     "dev": {
-        "alb_enabled": false // enable provisioning of application load balancing resources
+        "alb_enabled": false, // enable provisioning of application load balancing resources. If set to true, you _must_ fill out the alb_vpc_config option as well.
         "alb_vpc_config": {
-            "CertificateArn": // ACM certificate ARN for ALB
-            "SubnetIds": [] // list of subnets for ALB
+            "CertificateArn": "your_acm_certificate_arn", // ACM certificate ARN for ALB
+            "SubnetIds": [], // list of subnets for ALB
             "SecurityGroupIds": [] // list of security groups for ALB
-        }
+        },
         "api_key_required": false, // enable securing API Gateway endpoints with x-api-key header (default False)
         "api_key": "your_api_key_id", // optional, use an existing API key. The option "api_key_required" must be true to apply
         "apigateway_enabled": true, // Set to false if you don't want to create an API Gateway resource. Default true.
@@ -873,7 +871,7 @@ to change Zappa's behavior. Use these at your own risk!
         "delete_s3_zip": true, // Delete the s3 zip archive. Default true.
         "django_settings": "your_project.production_settings", // The modular path to your Django project's settings. For Django projects only.
         "domain": "yourapp.yourdomain.com", // Required if you're using a domain
-        "base_path": "your-base-path", // Optional base path for API gateway custom domain base path mapping. Default None.
+        "base_path": "your-base-path", // Optional base path for API gateway custom domain base path mapping. Default None. Not supported for use with Application Load Balancer event sources.
         "environment_variables": {"your_key": "your_value"}, // A dictionary of environment variables that will be available to your deployed app. See also "remote_env" and "aws_environment_variables". Default {}.
         "events": [
             {   // Recurring events
@@ -1394,6 +1392,8 @@ Like API Gateway, Zappa can automatically provision ALB resources for you.  You'
 ```
 
 More information on using ALB as an event source for Lambda can be found [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html).
+
+*An important note*: right now, Zappa will provision ONE lambda to ONE load balancer, which means using `base_path` along with ALB configuration is currently unsupported.
 
 ## Zappa Guides
 
--- a/README.md
+++ b/README.md
@@ -819,6 +819,12 @@ to change Zappa's behavior. Use these at your own risk!
 ```javascript
  {
     "dev": {
+        "alb_enabled": false // enable provisioning of application load balancing resources
+        "alb_vpc_config": {
+            "CertificateArn": // ACM certificate ARN for ALB
+            "SubnetIds": [] // list of subnets for ALB
+            "SecurityGroupIds": [] // list of security groups for ALB
+        }
         "api_key_required": false, // enable securing API Gateway endpoints with x-api-key header (default False)
         "api_key": "your_api_key_id", // optional, use an existing API key. The option "api_key_required" must be true to apply
         "apigateway_enabled": true, // Set to false if you don't want to create an API Gateway resource. Default true.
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -1365,6 +1366,22 @@ For monitoring of different deployments, a unique UUID for each package is avail
...
 ```
 
+### Application Load Balancer Event Source
+
+Zappa can be used to handle events triggered by Application Load Balancers (ALB). This can be useful in a few circumstances:
+- Since API Gateway has a hard limit of 30 seconds before timing out, you can use an ALB for longer running requests.
+- API Gateway is billed per-request; therefore, costs can become excessive with high throughput services. ALBs pricing model makes much more sense financially if you're expecting a lot of traffic to your Lambda.
+- ALBs can be placed within a VPC, which may make more sense for private endpoints than using API Gateway's private model (using AWS PrivateLink).
+
+Zappa can't automatically provision an ALB for you, but you can set your Lambda to be an ALB target by doing the following:
```
youcandanch: TODO: Update documentation around this, as we're pulling in the provisioning code as well.

zappa/cli.py
```diff
@@ -2080,6 +2095,11 @@ def load_settings(self, settings_file=None, session=None):
         self.xray_tracing = self.stage_config.get('xray_tracing', False)
         self.desired_role_arn = self.stage_config.get('role_arn')
 
+        # Load ALB-related settings
+        self.use_alb = self.stage_config.get('alb_enabled', False)
+        self.alb_vpc_config = self.stage_config.get('alb_vpc_config', {})
+        self.alb_lambda_alias = 'current-alb-version' if self.use_alb else None
```
youcandanch: TODO: [Change this to a constant.](https://github.com/youcandanch/Zappa/commit/9d13780daa1f5cf22f48ebc4b9e25ae5f79ba83e#r32557898)

zappa/cli.py
```diff
@@ -917,31 +931,29 @@ def update(self, source_zip=None, no_upload=False):
...
-                                            bucket,
-                                            self.lambda_name,
-                                            key_name,
-                                            num_revisions=self.num_retained_versions
-                                        )
+            kwargs = kwargs.update(dict(
+                bucket=bucket,
+                s3_key=key_name
+            ))
+            self.lambda_arn = self.zappa.update_lambda_function(**kwargs)
```
youcandanch: TODO: [Simplify this by always creating an alias, even without ALB support.](https://github.com/youcandanch/Zappa/commit/9d13780daa1f5cf22f48ebc4b9e25ae5f79ba83e#r32557878)

youcandanch:
@PeteW / @jneves we're pretty much at the finish line with this. I'll attempt to take it over it tonight. Remaining:
- Remediate the TODOs that @PeteW noted from code review.
- Tweak the documentation to reflect provisioning.
- Maybe add more tests if the coveralls report shows we've got substantial gaps.

At that point we should be clear to run some final tests and then merge.

youcandanch:
@PeteW this should complete the work for ALB support from tip to tail. Been talking to @jneves about it and he's going to take it for a spin tomorrow. Can you eye it and let me know if there's any concerns?

PeteW:
@youcandanch we originally agreed to make this two separate PRs but I presume you've been discussing this with @jneves in separate threads and reached some consensus to consolidate. I had concerns about that approach but looking at the consolidated files you put together I cant see any issue which hasnt been fully addressed.

youcandanch:
@PeteW it's mostly that there's a good chunk of things that have to adjust when the provisioning part lands; they _can_ be decoupled, but considering we wanted both as part of the next release to PyPI, we'd end up landing a PR that would swiftly have a good chunk of alterations added to it in the next PR. Sorry for any confusion that caused!

mathom:
We tested this today and ran into an issue:
```
Deploying ALB infrastructure...
Oh no! An error occurred! :(

==============

  File "/Users/collinferguson/development/coding/2019_coding/prod_support/prod-support-web-app/support-api/prod_support_web/venv/lib/python3.7/site-packages/zappa/cli.py", line 2759, in handle
    sys.exit(cli.handle())
  File "/Users/collinferguson/development/coding/2019_coding/prod_support/prod-support-web-app/support-api/prod_support_web/venv/lib/python3.7/site-packages/zappa/cli.py", line 509, in handle
    self.dispatch_command(self.command, stage)
  File "/Users/collinferguson/development/coding/2019_coding/prod_support/prod-support-web-app/support-api/prod_support_web/venv/lib/python3.7/site-packages/zappa/cli.py", line 546, in dispatch_command
    self.deploy(self.vargs['zip'])
  File "/Users/collinferguson/development/coding/2019_coding/prod_support/prod-support-web-app/support-api/prod_support_web/venv/lib/python3.7/site-packages/zappa/cli.py", line 792, in deploy
    self.zappa.deploy_lambda_alb(**kwargs)
  File "/Users/collinferguson/development/coding/2019_coding/prod_support/prod-support-web-app/support-api/prod_support_web/venv/lib/python3.7/site-packages/zappa/core.py", line 1387, in deploy_lambda_alb
    response = self.lambda_client.add_permission(**kwargs)
  File "/Users/collinferguson/development/coding/2019_coding/prod_support/prod-support-web-app/support-api/prod_support_web/venv/lib/python3.7/site-packages/botocore/client.py", line 357, in _api_call
  File "/Users/collinferguson/development/coding/2019_coding/prod_support/prod-support-web-app/support-api/prod_support_web/venv/lib/python3.7/site-packages/botocore/client.py", line 661, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the AddPermission operation: Cannot find alias arn: arn:aws:lambda:us-east-1:SNIP:function:support-api-prod:current-alb-version```

Our deployed lambda had no aliases associated so maybe its an issue with the order of operations.

**Comment by youcandanch** [comment]
Interesting. I'll take it for a spin on a clean repo tonight. Thanks for the report!

**Comment by youcandanch** [comment]
@mathom yep, had a bit of vestigal code left over from a refactor. Pushed a commit, tested locally, that's fixed now. Thanks!

File: .vscode/settings.json [hunk 4]
```diff
@@ -0,0 +1,3 @@
+{
```
jneves: Can you remove this file please? I suspect it was added before the .gitignore change.

README.md
```diff
@@ -653,14 +654,14 @@ Putting a try..except block on an asynchronous task like this:
 @task
 def make_pie():
     try:
-	ingredients = get_ingredients()
-	pie = bake(ingredients)
-	deliver(pie)
-	
+    ingredients = get_ingredients()
```
jneves: This change breaks the sample code.

README.md
```diff
@@ -669,12 +670,12 @@ will cause an email to be sent twice for the same error. See [asynchronous retri
 @task
 def make_pie():
     try:
-	"""code block"""
-	
+    """code block"""
```
jneves: Same here.

zappa/__init__.py
```diff
@@ -1,6 +1,6 @@
 import sys
 
-SUPPORTED_VERSIONS = [(2, 7), (3, 6)]
+SUPPORTED_VERSIONS = [(2, 7), (3, 6), (3, 7)]
```
jneves: Sorry, the 3.7 update is in another PR, not merged yet. Can you reset this please.

README.md
```diff
@@ -818,6 +819,12 @@ to change Zappa's behavior. Use these at your own risk!
 ```javascript
  {
     "dev": {
+        "alb_enabled": false // enable provisioning of application load balancing resources
```
jneves: This should also note that the "alb_vpc_config" section is mandatory if alb_enabled is true.

README.md
```diff
@@ -818,6 +819,12 @@ to change Zappa's behavior. Use these at your own risk!
...
+        "alb_vpc_config": {
+            "CertificateArn": // ACM certificate ARN for ALB
```
jneves: We try to keep the reference as legal JSON so it can be used as a template for a configuration.

```suggestion
            "CertificateArn": "arn_of_the_certificate", // ACM certificate ARN for ALB
```

README.md
```diff
@@ -818,6 +819,12 @@ to change Zappa's behavior. Use these at your own risk!
...
+            "SubnetIds": [] // list of subnets for ALB
```
jneves:
```suggestion
            "SubnetIds": [], // list of subnets for ALB
```

youcandanch:
Ah, yep, I'll get to 'em likely tomorrow. Thanks João!

zappa/__init__.py
youcandanch: Ah, yep, was testing locally and, uh... may have been too lazy to explicitly call 3.6 every time. My bad!

jneves:
@youcandanch can you add a note on the reference that base_path and alb_enabled are not compatible for now? (it will be needed to support one ALB using multiple lambdas, but I'm OK with that being added afterwards, just document the incompatibility at the moment.)

jneves:
Thank you very much @youcandanch and @PeteW!

ambientlight:
@PeteW, @jneves:

Guys, quick question related to this:
https://github.com/Miserlou/Zappa/blob/c925275b990527d534a6ac265f5a3472fdcf36fb/zappa/core.py#L2834-L2842

https://github.com/Miserlou/Zappa/blob/c925275b990527d534a6ac265f5a3472fdcf36fb/zappa/core.py#L2559-L2575

It looks like lambda ALB permission (in function's resource-based policy) like:

```
{
    "Sid": "",
    "Effect": "Allow",
    "Principal": {
        "Service": "elasticloadbalancing.amazonaws.com"
    },
    "Action": "lambda:InvokeFunction",
    "Resource": "arn:aws:lambda:",
    "Condition": {
        "ArnLike": {
            "AWS:SourceArn": "arn:aws:elasticloadbalancing:"
        }
    }
}
```

going to be removed if zappa events are used on `update`. I am just wondering whether this bring s a problem where next update will result in ALB not being able to invoke lambda. Have you guys tested this pull request will lambda events? (haven't tested myself yet)

jneves:
@ambientlight from my tests it doesn't - this was my sequence of commands and at all points the ALB access was still working:

```
  527  zappa deploy dev
  528  zappa schedule dev
  529  zappa unschedule dev
  530  zappa schedule dev
  531  zappa update dev
  532  zappa unschedule dev
```

Re-looking at the code, I can see why, the _clear_policy resets who can trigger events (policy), not the permissions (controlled by the execution role).

ambientlight:
@jneves: 
Ability to invoke lambdas by AWS services are controlled by those resource based policies, execution role is other way round, allows lambdas to access AWS services, am I right?

I mean that `_clear_policy` will not only clean policies that allow triggering lambda from events, but also clean the policies with allow calling lambdas by ALB.

Issue #1199, opened 2017-10-27
Zappa with Flask cannot route unicode strings.


DingGGu:
Zappa cannot route Unicode Strings
## Context
I'm using Zappa with Flask in Python3.6

Zappa cannot route string that include unicodes.

```python
@bp.route('/product/<name>')
def teacher(name):
    return flask.jsonify(dict(name=name))
```

Request To: "/product/123"
ResponseBody: {"name": "123"}

Request To: "/product/ㅁㄴㅇㄹ"
ResponseBody: Displaying Error 

Log:

```js
{
'message': 'An uncaught exception happened while servicing this request. You can investigate this with the `zappa tail` command.',

'traceback': [
[...]

PR #1814, opened 2019-03-14
Fix Python 2 Unicode path behavior

littlebtc:
I am investigating our Zappa code problem but found out a regression caused by previous fix.
So I am sending a pull request.

Also I made the commit message wrong. Should be only `decode` to iso-8859-1 instead of `encode`.

## Description

Fix Python 2 unicode path behavior, by only decoding to iso-8859-1 on Python 3

Minimal Flask app to introduce the problem:
```python
import sys
from flask import Flask, request, jsonify

app = Flask(__name__)


@app.route('/a/<name>')
def name(name):
    return jsonify({
        'url': request.url
    })


if __name__ == '__main__':
    app.run(debug=True)
```

Use the following URL:
```https://~~~.amazonaws.com/dev/a/%E6%B8%AC%E8%A9%A6?q=%E6%B8%AC%E8%A9%A6```

The URL in JSON should be ends with `/dev/a/測試?q=測試` in either Python 2 or 3 locally.
But for current Zappa, it ends with `/a/æ¸¬è©¦?q=測試"` in Python 2.

### Reason and fixes

The WSGI implementations (both Werkzeug and Django) expect `iso-8859-1`-decoded string **ONLY** if it is under Python 3.

So we fixed the Unicode URL path in Python 3 but braking Python 2.

This patch should bring the Python 2 behavior back.

## GitHub Issues
A #1199 follow-up.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -563,21 +563,30 @@ class TestZappa(unittest.TestCase):
         request = create_wsgi_request(event, trailing_slash=True)
         self.assertEqual("/path:1", request['PATH_INFO'])
 
-    def test_wsgi_latin1(self):
+    def test_wsgi_unicode_path(self):
+        # '\xe6\xb8\xac\xe8\xa9\xa6' == u'\u6e2c\u8a66'.encode('utf-8')
+        # Which is "test" in Chinese
         event = {
             "body": {},
             "headers": {},
             "pathParameters": {},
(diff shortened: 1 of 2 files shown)


jneves:
Sorry that it wasn't merged before, closing as not an issue now that we've removed py2 support.

PR #1817, opened 2019-03-15
Update guide links

cdfuller:
I found two dead links in the guides section. I removed one and replaced the other with the copy of same article hosted on medium.


jneves:
Thanks!

PR #1822, opened 2019-03-19
Simplify Python version detection and remove the backslash

cclauss:
## Description
 
Simplify Python version detection and remove the backslash


jneves:
Thanks!

Issue #1823, opened 2019-03-19
Unable to update lambda function


bruceduhamel:
## Context
With the latest release of Zappa (0.48.0), I'm no longer able to update my functions via the zappa update command. I reverted to 0.47.0 and found that it worked as expected.

Initially, when using 0.48.0, the update command failed because it lacked the UpdateAlias lambda permissions. I updated the permissions and retried and it failed with the below error:

```
Traceback (most recent call last):
  File "/home/circleci/repo/env/lib/python2.7/site-packages/zappa/cli.py", line 2759, in handle
    sys.exit(cli.handle())
  File "/home/circleci/repo/env/lib/python2.7/site-packages/zappa/cli.py", line 509, in handle
    self.dispatch_command(self.command, stage)
  File "/home/circleci/repo/env/lib/python2.7/site-packages/zappa/cli.py", line 556, in dispatch_command
    self.update(self.vargs['zip'], self.vargs['no_upload'])
  File "/home/circleci/repo/env/lib/python2.7/site-packages/zappa/cli.py", line 950, in update
    self.lambda_arn = self.zappa.update_lambda_function(**kwargs)
  File "/home/circleci/repo/env/lib/python2.7/site-packages/zappa/core.py", line 1135, in update_lambda_function
    Name=ALB_LAMBDA_ALIAS,
  File "/home/circleci/repo/env/lib/python2.7/site-packages/botocore/client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/home/circleci/repo/env/lib/python2.7/site-packages/botocore/client.py", line 661, in _make_api_call
    raise error_class(parsed_response, operation_name)
ResourceNotFoundException: An error occurred (ResourceNotFoundException) when calling the UpdateAlias operation: Alias not found: arn:aws:lambda:us-west-2:xxx:function:xxx-xxx-xxx:current-alb-version
```

## Expected Behavior

The lambda function is updated when calling zappa update without errors.
## Actual Behavior

The update command fails with the above error.

## Steps to Reproduce

1. pip install zappa (installs 0.48.0)
2. zappa update test (deploy lambda for test environment)
3. See above error ResourceNotFoundException

## Your Environment

* Zappa version used: 0.48.0
* Operating System and Python version: Ubuntu 18.04, Python 2.7.15rc1
* The output of `pip freeze`:
```
argcomplete==1.9.3
backoff==1.8.0
boto3==1.9.116
botocore==1.12.116
[...]

PR #1824, opened 2019-03-19
Fix an issue with ALB aliases and already-deployed functions.

youcandanch:
### Description
In #1807, we introduced support for application load balancers as an event source for Zappa-driven functions. Unfortunately, that PR also introduced a bug where previously deployed services became unable to be updated with `zappa update`, as the code branch dealing with Lambda aliases wasn't checking for the existence of those aliases first. This PR fixes this bug two ways:

a) Check for the existence of the alias first before trying to update it. This handles the narrow case of older applications failing to deploy. 
b) Only _create_ the alias given that ALB provisioning is requested. Since we're checking to see if it exists before updating anyway, we can safely _not_ create it in any circumstance where an ALB isn't part of the configuration. 

### Issues
https://github.com/Miserlou/Zappa/issues/1823

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -280,7 +280,7 @@ class Zappa(object):
             self.manylinux_wheel_file_suffix = 'cp36m-manylinux1_x86_64.whl'
         else:
             self.manylinux_wheel_file_suffix = 'cp37m-manylinux1_x86_64.whl'
-            
+
         self.endpoint_urls = endpoint_urls
         self.xray_tracing = xray_tracing
 
@@ -1042,7 +1042,8 @@ class Zappa(object):
                                 aws_environment_variables=None,
                                 aws_kms_key_arn=None,
                                 xray_tracing=False,
-                                local_zip=None
+                                local_zip=None,
+                                use_alb=False,
                             ):
         """
         Given a bucket and key (or a local path) of a valid Lambda-zip, a function name and a handler, register that Lambda function.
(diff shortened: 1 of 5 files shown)


zappa/core.py
```diff
@@ -1126,14 +1129,27 @@ def update_lambda_function(self, bucket, function_name, s3_key=None, publish=Tru
...
+        # If the lambda has an ALB alias, let's update the alias
+        # to point to the newest version of the function.
         # Related: https://github.com/Miserlou/Zappa/pull/1730
-        self.lambda_client.update_alias(
-            FunctionName=function_name,
-            FunctionVersion=version,
-            Name=ALB_LAMBDA_ALIAS,
-        )
+        #          https://github.com/Miserlou/Zappa/issues/1823
+        try:
```
bruceduhamel: When I first saw how this was determining if an alias already existed I went to the AWS API docs to see if there was an API for testing if an alias exists e.g. HEAD /some/endpoint/alias as opposed to GET /home/endpoint/alias with the intent of avoiding the need for the try-catch block. 

It seems there is no such HEAD endpoint, and maybe everybody else is familiar with that, but it may be worth leaving a comment here explaining that.

youcandanch: Seems reasonable. Let me add that in.

youcandanch:
As far as I can tell, this is good to go. @jneves lemme know if there's anything else I should take on this.

jneves:
Doing a final test run.

Issue #1825, opened 2019-03-19
CORS issue: 0.48 release returns same value multiple times in access-control-allow-origin


vijayparne:
## Expected Behavior
access-control-allow-origin: https://www.example.com

## Actual Behavior
access-control-allow-origin: https://www.example.com, https://www.example.com, https://www.example.com

## Your Environment

* Zappa version used: 0.48
* Operating System and Python version: macOS 10.14: Mojave and python 3.6
* The output of `pip freeze`: 
* Link to your project (optional):
* Your `zappa_settings.py`:  
{
  "dev": {
    "app_function": "myapi.app.app",
    "cors": true,
    "binary_support": false,
  }
}

jneves:
Looks related to #1821

jneves:
Could you check if https://github.com/Miserlou/Zappa/pull/1828 works for you?

sookyeomKim:
[https://github.com/Miserlou/Zappa/pull/1828/files](url)
![스크린샷 2019-03-20 오후 6 02 01](https://user-images.githubusercontent.com/8381017/54673086-e1cde980-4b3c-11e9-8720-1e30d29d9cd1.png)

It worked very well.

jedestep:
LGTM, I seem to have misunderstood the intended behavior of this case. Thanks for digging in @jneves.

jneves:
@vijayparne can you confirm this is fixed in 0.48.2?

vijayparne:
@jneves it works. Thanks

Issue #1821, opened 2019-03-19
Django HTTP_HOST not valid according to RFC


marvoloe:
## Context

My `requirements.txt`:

```
Django~=1.11
django-countries-plus==1.2.1
djangorestframework==3.9.0
-e git+https://github.com/cuda-networks/django-eb-sqs.git#egg=django-eb-sqs
pymysql
``` 

My `settings.py`:

```python
ALLOWED_HOSTS = [
    'api.example.com',
    'localhost',
]
```

## Expected Behavior

I am not certain if I am mistaken , but I would hope that this error doesn't show.

## Actual Behavior

For some reason my *custom domain name* is three times in the HTTP_HOST variable:

```
DisallowedHost
Invalid HTTP_HOST header: 'api.exmaple.com, api.exmaple.com, api.exmaple.com'. The domain name provided is not valid according to RFC 1034/1035.
```

I did all the necessary setup in API Gateway, and have the domain direct to the *Target Domain Name* via `CNAME`.

## Possible Fix

I did try to change the Django `ALLOWED_HOSTS` as an ugly workaround, which didn't work:

```python
ALLOWED_HOSTS = [
    'api.example.com',
    'api.example.com, api.example.com, api.example.com',
    'localhost',
]
```

## Steps to Reproduce

n / a

## Your Environment

* Zappa version used: master (waiting for https://github.com/Miserlou/Zappa/pull/1762 to become available as version)
* Operating System and Python version: Python 3.7 on AWS Lambda
* The output of `pip freeze`:

```
argcomplete==1.9.3
awscli==1.16.125
awsebcli==3.14.13
blessed==1.15.0
boto3==1.9.115
botocore==1.12.115
cached-property==1.5.1
cement==2.8.2
certifi==2019.3.9
cfn-flip==1.1.0.post1
chardet==3.0.4
Click==7.0
colorama==0.3.9
coreapi==2.3.3
coreschema==0.0.4
Django==1.11.20
django-countries-plus==1.2.1
-e git+https://github.com/cuda-networks/django-eb-sqs.git@2c5dff0392fad2ed383e4991791239e1052437da#egg=django_eb_sqs
djangorestframework==3.9.0
docker==3.7.0
docker-compose==1.23.2
docker-pycreds==0.4.0
dockerpty==0.4.1
docopt==0.6.2
docutils==0.14
durationpy==0.5
future==0.16.0
hjson==3.0.1
idna==2.7
inflection==0.3.1
[...]

PR #1828, opened 2019-03-19
Fix for repeated values on headers

jneves:
## Description

This changes the merge_headers function to just copy the multiValueHeaders and add non-existing entries from headers instead of replicating existing values multiple times.

## GitHub Issues

#1821
#1825

### DIFF
--- a/tests/test_handler.py
+++ b/tests/test_handler.py
@@ -402,7 +402,7 @@ class TestZappa(unittest.TestCase):
         }
 
         merged = merge_headers(event)
-        self.assertEqual(merged['a'], 'c, b')
+        self.assertEqual(merged['a'], 'c')
         self.assertEqual(merged['x'], 'y')
         self.assertEqual(merged['z'], 'q')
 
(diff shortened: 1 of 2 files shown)


jneves:
@youcandanch when you can, can you check this? Main issue is that values are getting multiplied (because they come in headers and multi_headers). I changed the logic to just had on multiValueHeaders the ones that aren't already there. An alternative would be to keep them as a set instead of a list until the final `','.join()`. Was there a reason why you always merged the values in?

IlyaSukhanov:
This PR also fixes a related bug where [python3-saml](https://github.com/onelogin/python3-saml) gets confused on host string:
```
[1553099381941] [DEBUG] 2019-03-20T16:29:41.941Z 8f4fab42-7659-4739-9428-3d7de85c541e host found: [example.com, example.com]
...
[1553099382037] [DEBUG] 2019-03-20T16:29:42.37Z 8f4fab42-7659-4739-9428-3d7de85c541e SAML Error reason: The response was received at https://example.com, example.com, example.com/saml/example/assert instead of https://example.com/saml/example/assert
```

youcandanch:
The multi-header logic was from #1675, I just cherry-picked that commit into #1802. This looks totally sound by me.

marvoloe:
Thank you guys. Works for me now.

Issue #1832, opened 2019-03-21
Allow Existing ALB to be used


whahoo:
I have been following along and using the ALB events and provisioning PR #1807  that was just merged in by @PeteW and @youcandanch.

I was using those changes and integrating my Lambda's into an existing ALB provisioned via terraform

I have some proposed zappa settings and code changes to allow the use of an existing ALB, multiple Zappa projects can then be deployed onto the same ALB

When using an existing ALB, there is an assumption of the Listener existing also,
So instead of creating these resources Listener Rules need to be created
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.create_rule

These can be defiend in the zappa_settings like : 
```
"alb_vpc_config": {
    "LoadBalancerArn": "arn:aws:acm:us-east-1:[your-account-id]:loadbalancer/app/[alb-name]/[f851177b4d4eb840]",
    "alb_listener_rule_conditions": {
        "Field": "path-pattern|host-header",
        "Values": ["api/*"|"api.example.com"]
    },
    "alb_listener_rule_priority": 1
}

**Comment by whahoo** [comment]
I have updated the PR now to allow multiple listener rule conditions to be configured

The alb_listener_rule_conditions attribute is now a list, in line with the boto3 api
https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/elbv2.html#ElasticLoadBalancingv2.Client.create_rule

```
    "alb_listener_rule_conditions": [
                {
                    "Field": "path-pattern",
                    "Values": ["/api/v1/*"]
                },
                {
                    "Field": "host-header",
                    "Values":["api.domainname.org"]
                }
            ]
```

This allows for a host and path-pattern combination, and it also supports the newly announced extra routing features:
https://aws.amazon.com/blogs/aws/new-advanced-request-routing-for-aws-application-load-balancers/

PeteW:
I haven't completed an in-depth review of the code but your design and test coverage makes a lot of sense to me in terms of compatibility with partially-established environments and those with need for special rules.

cotman:
We would also be really interested in seeing this support.

[...]

PR #1833, opened 2019-03-21
Add Support for using an existing ALB and Listener

whahoo:
* Did you write a test for your new code?

* Did the Travis build pass?

* Did you improve (or at least not significantly reduce)  the amount of code test coverage?

## Description
I have been following along and using the ALB events and provisioning PR #1807  that was just merged in by @PeteW and @youcandanch.

I was using those changes and integrating my Lambda's into an existing ALB provisioned via terraform

I have some proposed zappa settings and code changes to allow the use of an existing ALB, multiple Zappa projects can then be deployed onto the same ALB

## GitHub Issues

#1832

### DIFF
--- a/README.md
+++ b/README.md
@@ -821,9 +821,17 @@ to change Zappa's behavior. Use these at your own risk!
     "dev": {
         "alb_enabled": false, // enable provisioning of application load balancing resources. If set to true, you _must_ fill out the alb_vpc_config option as well.
         "alb_vpc_config": {
-            "CertificateArn": "your_acm_certificate_arn", // ACM certificate ARN for ALB
+            // To create an ALB set CertificateArn, SubnetIds, and SecurityGroupIds
+            "CertificateArn": "your_acm_certificate_arn", // ACM certificate ARN for ALB  - not required if LoadBalancerArn is specified
             "SubnetIds": [], // list of subnets for ALB
             "SecurityGroupIds": [] // list of security groups for ALB
+            // To use an existing ALB set LoadBalancerArn, alb_listener_rule_conditions, alb_listener_rule_priority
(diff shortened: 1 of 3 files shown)


markwatson:
Is this something that's still potentially going to be merged down to master? It would be useful to some of the projects I'm working on right now, so just wondering. Thanks!

blockmar:
Any update on this? Would be very useful in bigger environment where having a 1 to 1 mapping between lambdas and ALBs makes no sense.

ssilverm:
Another bump - this would be super useful for us.

ssilverm:
@whahoo I noticed an issue testing this branch with my zappa deploys and trying to move them to an ALB.

I have two listeners on my ALB.  One for port 80, and one for port 443.  This PR is erroring on getting both listeners.

```OSError: Failure to load listeners for ALB. or more than 1 found, Response was in unexpected format. Response was:```

We just have 80 as a catch all and fwd to port 443.

ssilverm:
Also base-path was required in the zappa settings to make this work for us.  

Other than the two gotchas we experience, this is working great! @whahoo @Miserlou @jneves

ssilverm:
@jneves any updates on this?  Would love to get this running on our production ALBs

treharne:
For anyone stuck on this like I was - [here's a workaround ](https://github.com/Miserlou/Zappa/issues/1832#issuecomment-629557049) where you set up your ALB listeners to point to your lambda using an Alias **once** and then can `zappa update` without breaking the link

PeteW:
Speaking as the original developer of ALB support I've read this PR and found it both clear and comprehensive. Also, if this can get merged in I see myself using this new feature wherever possible. It is a good practice from a cost/surface area consolidation perspective to be able to reuse existing ALBs.

PR #1842, opened 2019-03-29
Support for layers - extension of #1752

jedestep:
## Description
This PR is based off #1752. My additional commits add support for `zappa update` to ensure that layers aren't removed when updating a function, as well as adding some test cases.

## GitHub Issues
#1716


jackric:
FYI your new unit test is pointless. It just patches a function, calls the patched function, and asserts the patched function was called with what you called it with. It provides no assurances of the correctness of your new code.

I suggest either removing the pointless test, or writing a test that patches low-level AWS operations with mocks , pass a config with layers in at the high-level, and assert the layers option got passed down to the AWS operations correctly.

suriya:
Worked for me as well. Thank you, this is very helpful.

jackric:
@Miserlou can we get this moving?

millarm:
Included in #2027

PR #1865, opened 2019-05-06
Updated mirrors

sourcepirate:
Updated to use pypi.org mirror instead of pypi.python.org

## Description
 Updated the codes to use pypi.org from pypi.python.org.
Since pypi.python.org is deprecated.


sourcepirate:
@mbeacom any update on this ?

mbeacom:
@sourcepirate Nothing I can really do...  LGTM, but I can't merge it. 😅

Issue #1861, opened 2019-05-03
[Enhancement] Use abi3 binary versions for packages such as cryptography that don't have manylinux for Python 3.6/3.7


nikbora:
## Context
Zappa currently uses a combination of manylinux wheels and lambda-packages to get the right binary during deploy time. In case of manylinux, some libraries like `cryptography` don't have a manylinux binary for Python 3.6 and 3.7, but do have one for abi3 and Python 3.4, which is suppossedly upwards compatible. See this thread for more info: https://github.com/pyca/cryptography/issues/4406

## Expected Behavior
When picking the right binary wheel for Python 3.x, Zappa should be smart enough to see if there is abi3 for any 3.x versions below the one needed.

## Actual Behavior
Zappa only looks at Python 3.7/3.6 manylinux binary and thinks there isn't one. In the case of cryptography (that does have an abi3 one for Python 3.4) Zappa then goes to lambda-packages and uses whatever version is there, irrespective of the one you have in your virtualenv.

## Possible Fix
1. Test that using abi3 binaries does indeed work on on AWS Lambda (using cryptography as a guinnea pig).
2. Add code into the packaging/deploy logic to resort to this if it exists.

anttisalonen:
FYI my commit above fixed this issue for me (cryptography library on Python 3.7) - but it would be good if someone else who has the same issue can try on their end.

```
Calling update for stage dev..
Downloading and installing dependencies..
...
 - cryptography==2.7: Downloading
100%|████████████████████████████████████████████████████████████████████████████████████████████████| 2.28M/2.28M [00:00<00:00, 4.96MB/s]
...
'python3.7'
Packaging project as zip.

```

magnunm:
I too had an issue when trying to use paramiko (dependent on cryptography) in pyhton 3.6. Deployment fails and CloudWatch logs show 
```
File "/var/task/bcrypt/__init__.py", line 25, in <module>
from . import _bcrypt
ImportError: cannot import name '_bcrypt'
```
After some back and forth @anttisalonen 's solution solved it! 😄

PR #1886, opened 2019-06-17
Issue #1861 add support for abi3 binaries from manylinux

anttisalonen:
## Description
 
https://github.com/Miserlou/Zappa/issues/1861 describes the lack of using abi3 wheels. This breaks use of libraries such as cryptography on Python 3.x when deployed from Mac OS. This commit adds support for using abi3 wheels.

## GitHub Issues

https://github.com/Miserlou/Zappa/issues/1861


jneves:
Thank you very much. I started with your PR and ended up with a more generic version on #1999 but it kicked off with your work, so thank you!

Issue #1888, opened 2019-06-20
"tags" values in "zappa_settings.py" are not applied to the API Gateway stage that Zappa creates


cotman:
The `tags` section of `zappa_settings.py` is documented as follows:

    "tags": { // Attach additional tags to AWS Resources
        "Key": "Value",  // Example Key and value
        "Key2": "Value2",
        },

However, when deploying for API Gateway, the tags are only applied to the Lambda function, not to the associated API Gateway stage.

## Context

This is a bit of an issue, primarily as AWS tags are often used for:

* Cost allocation
* Third-party AWS metric services (such as Datadog)

When tags aren't applied to the API Gateway stage that Zappa creates and maintains, then they can't be used to track API Gateway specific costs or used in third-party metric services to filter metrics/set monitoring alerts.

## Expected Behavior

The tags specified in the `tags` section of `zappa_settings.py` should be applied to all AWS resources that Zappa creates/manages.

## Actual Behavior

Tags are only applied to the Lambda function.

## Possible Fix

Apply the same tags that are applied to the Lambda function when creating/updating the API Gateway stage.

## Steps to Reproduce

1. Create a Zappa project that uses API Gateway
2. Set some tag key/values in the `tags` section of `zappa_settings.py` in the Zappa project
3. Deploy the API Gateway API
4. Select the API Gateway stage that Zappa has created in the AWS console and select the "Configure Tags" button - no tags are present

## Your Environment

* Zappa version used: 0.48.2
* Operating System and Python version: Linux (Ubuntu 18.04) - Python 3.6
* The output of `pip freeze`:

    argcomplete==1.9.3
    boto3==1.9.172
    botocore==1.12.172
    certifi==2019.6.16
    cfn-flip==1.2.1
    chardet==3.0.4
    Click==7.0
    docutils==0.14
    durationpy==0.5
    future==0.16.0
    hjson==3.0.1
    idna==2.8
    jmespath==0.9.3
    kappa==0.6.0
    lambda-packages==0.20.0
    pkg-resources==0.0.0
    placebo==0.9.0
    python-dateutil==2.6.1
    python-slugify==1.2.4
    PyYAML==5.1.1
    requests==2.22.0
    s3transfer==0.2.1
    six==1.12.0
    toml==0.10.0
    tqdm==4.19.1
    troposphere==2.4.7
    Unidecode==1.1.0
    urllib3==1.25.3
    Werkzeug==0.15.4
    wsgi-request-logger==0.4.6
[...]

PR #1889, opened 2019-06-20
Issue #1888 Apply tags to API Gateway stages.

cotman:
Fix for the related issue.

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1794,6 +1794,16 @@ class Zappa(object):
             ]
         )
 
+        if self.tags:
+            self.apigateway_client.tag_resource(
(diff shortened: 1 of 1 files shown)


cotman:
mmm...actually, there are two layers of tags here:

* Globally on the API Gateway as a whole
* Per-stage, within the API Gateway

This PR adds the tags to the stage within API Gateway, although it should probably add them to the API Gateway as a whole (or possibly both...maybe ;) ).

cotman:
Actually, no, I think stage tagging *is* appropriate here, so the PR stands as it is.

cotman:
PR updated so that it also tags any ALBs generated by Zappa.

Issue #1747, opened 2019-01-04
AttributeError: 'ZappaCLI' object has no attribute 'apigateway_policy'


rajdotnet:
zappa template dev --l your-lambda-arn -r your-role-arn cli command gives AttributeError: 'ZappaCLI' object has no attribute 'apigateway_policy' error in version 0.47.1
## Context

This command works fine in version 0.47.0

## Expected Behavior
the zappa template command should work and generate cloudfromation template in version 0.47.0 instead of erroring out

## Actual Behavior
Calling template for stage dev..
Oh no! An error occurred! :(

==============

Traceback (most recent call last):
  File "c:\program files (x86)\python36-32\lib\site-packages\zappa\cli.py", line 2712, in handle
    sys.exit(cli.handle())
  File "c:\program files (x86)\python36-32\lib\site-packages\zappa\cli.py", line 509, in handle
    self.dispatch_command(self.command, stage)
  File "c:\program files (x86)\python36-32\lib\site-packages\zappa\cli.py", line 553, in dispatch_command
    json=self.vargs['json']
  File "c:\program files (x86)\python36-32\lib\site-packages\zappa\cli.py", line 666, in template
    policy=self.apigateway_policy
AttributeError: 'ZappaCLI' object has no attribute 'apigateway_policy'

## Steps to Reproduce

1. run zappa init
2. run zappa template dev --l your-lambda-arn -r your-role-arn
3.

## Your Environment

* Zappa version used: 0.47.1
* Operating System and Python version: Windows 10 python version 3.6.4
* The output of `pip freeze`:
argcomplete==1.9.3
arrow==0.12.1
astroid==2.0.4
aws-lambda-builders==0.0.4
aws-sam-cli==0.9.0
aws-sam-translator==1.9.0
binaryornot==0.4.4
boto3==1.9.67
botocore==1.12.67
certifi==2018.10.15
cfn-flip==1.0.3
chardet==3.0.4
chevron==0.13.1
click==6.7
colorama==0.3.9
cookiecutter==1.6.0
dateparser==0.7.0
docker==3.6.0
docker-pycreds==0.4.0
docutils==0.14
durationpy==0.5
Flask==1.0.2
future==0.16.0
hjson==3.0.1
idna==2.7
isort==4.3.4
itsdangerous==1.1.0
Jinja2==2.10
jinja2-time==0.2.0
jmespath==0.9.3
jsonschema==2.6.0
kappa==0.6.0
lambda-packages==0.20.0
[...]

PR #1582, opened 2018-08-07
Add support for API Gateway resource policy

mjgorman:
## Description
Added the ability to include a JSON policy for the api resource policy which allows things like restricting access based on IP to the apigateway endpoints.

PR #1903, opened 2019-07-22
Resolving issue #1747 removing unexpected 'policy' keyword argument.

monkut:
## Description
Attempt to resolve:
https://github.com/Miserlou/Zappa/issues/1747

Removes the _unexpected_ `policy` kwarg in the `self.zappa.create_stack_template()` call.

## GitHub Issues
Error occurs when callling `zappa template ....`
https://github.com/Miserlou/Zappa/issues/1747

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -663,7 +663,6 @@ class ZappaCLI(object):
                                             authorizer=self.authorizer,
                                             cors_options=self.cors,
                                             description=self.apigateway_description,
-                                            policy=self.apigateway_policy,
                                             endpoint_configuration=self.endpoint_configuration
                                         )
 


mpeteuil:
👍 This worked for me with Python 3.7 on v0.48.2.

As mentioned in https://github.com/Miserlou/Zappa/issues/1747#issuecomment-451785148, the `policy` argument was brought in as part of https://github.com/Miserlou/Zappa/pull/1582, but it doesn't seem like `create_stack_template` was ever changed to use it, so I'm not sure why it was added there. Since there's no test for the original change it's a little difficult to say whether this would result in any kind of regression.

mathom:
Fixed in #1942, thanks everyone who worked on it

Issue #849, opened 2017-05-14
More newbie help needed for "Error: Failed to manage IAM roles!"


nealmcb:
## Context

When I try to use the simple-looking instructions in the README, and deploy, I get

    Error: Failed to manage IAM roles!
    You may lack the necessary AWS permissions to automatically manage a "Zappa" execution role.

Visiting the listed help page (https://github.com/Miserlou/Zappa#using-custom-aws-iam-roles-and-policies) doesn't help much.  It just claims that
> By default, the Zappa client will create and manage the necessary IAM policies and roles to execute Zappa applications. 

which leaves me very unsure about the AWS setup necessary.  I'm not in some corporation that has restrictive policies. But I also I don't have some sort of default AWS user lying around that seems to fit the bill, and I don't see any clear documentation on how to create one. Disclaimer: I started using AWS back in 2007, and some of my configuration may be archaic or confused after years of occasional experiments.

So I'm looking for some sort of easy default `user` permissions configuration that still doesn't do something rude like give full AWS admin access to zappa or this little test script.

Issue #244 Minimum AWS policies for example looks like an attempt to address this in a more advanced way, and it also leaves me scratching my head.  The nicest detailed and spelled-out advice there is from ysong-sc (thanks!): https://github.com/Miserlou/Zappa/issues/244#issuecomment-268492715, but when I do the copy-paste job suggested there, and even after I try to fix some of the customization that isn't mentioned, like changing "`[BUCKET-NAME-IN-SETTINGS-JSON]`" twice to the value that `zappa init` gave me, I still get 
>    This policy contains the following error: There are invalid ARNs in the policy. For more information about the IAM policy grammar, see AWS IAM Policies.

The policy I'm trying is [pol3.txt](https://github.com/Miserlou/Zappa/files/999059/pol3.txt)

 

## Your Environment

* Zappa version used: 0.41.2
* Operating System and Python version: Ubuntu Trusty 14.04, Python 3.4.3
* Link to your project (optional):
* Your `zappa_settings.py`: 

```
{
    "dev": {
        "app_function": "happy_birthday.__hug_wsgi__",
        "aws_region": "us-east-1",
        "profile_name": "zappa",
        "s3_bucket": "zappa-y1015l3dv"
    }
}
```

nealmcb:
My security colleagues will shake their heads and sigh, but I just threw caution to the wind and gave my zappa user `AdministratorAccess`, and got my first zappa deployment up and running (with Hug!).

[...]

PR #1910, opened 2019-07-31
Shows the message of the exception reported by AWS when it is not po…

4lberto:
Shows the message of the exception reported by boto3 when it is not po…

## Description
When creating a lambda with a long project_name value, the AWS IAM says that the name is too long, but Zappa is only telling that is a general error managing IAM roles:

Calling deploy for stage dev..
Creating **************************************************** IAM Role..
Error: Failed to manage IAM roles!
You may lack the necessary AWS permissions to automatically manage a Zappa execution role.
To fix this, see here: https://github.com/Miserlou/Zappa#custom-aws-iam-roles-and-policies-for-deployment

I spent a full hour checking my IAM configuration until I realized that it could be a problem with then length of the name thanks to this ticket: https://github.com/Miserlou/Zappa/issues/849#issuecomment-482800451

The code I added just shows the main message from the exception returned by AWS, so just 1 line or 2 are added at the output of the Zappa deploy command:

Error reported by AWS:An error occurred (ValidationError) when calling the CreateRole operation: 1 validation error detected: Value '****************************************************' at 'roleName' failed to satisfy constraint: Member must have length less than or equal to 64

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/849#issuecomment-482800451

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -705,13 +705,16 @@ class ZappaCLI(object):
             if self.manage_roles:
                 try:
                     self.zappa.create_iam_roles()
-                except botocore.client.ClientError:
+                except botocore.client.ClientError as ce:
                     raise ClickException(
                         click.style("Failed", fg="red") + " to " + click.style("manage IAM roles", bold=True) + "!\n" +
                         "You may " + click.style("lack the necessary AWS permissions", bold=True) +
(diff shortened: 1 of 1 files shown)


jneves:
Thanks!

PR #1932, opened 2019-09-24
Fix #1816: Add basic support for authorization scopes

### LINKED ISSUE #1816: AuthorizationScopes not supported
## Context
When using Cognito, API Gateway provides the authorizationScopes property on the API Gateway Method to match against scopes in the access token.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-method.html#cfn-apigateway-method-authorizationscopes

> A list of authorization scopes configured on the method. The scopes are used with a COGNITO_USER_POOLS authorizer to authorize the method invocation. The authorization works by matching the method scopes against the scopes parsed from the access token in the incoming request. The method invocation is authorized if any method scopes matches a claimed scope in the access token. Otherwise, the invocation is not authorized. When the method scope is configured, the client must provide an access token instead of an identity token for authorization purposes. 

## Expected Behavior
Using a COGNITO_USERS_POOLS, we should be able to provide a list of scopes. By example:

```
{
    "authorizer": {
        "type": "COGNITO_USER_POOLS",
        "provider_arns": [
            "arn:aws:cognito-idp:{region}:{account_id}:userpool/{user_pool_id}"
        ],
        "authorization_scopes": [
            "scope_1",
            "scope_2",
        ],
    }
}

```

## Actual Behavior
All invocations are not authorized as no matching scopes are found

## Possible Fix
Enrich the CFN template with the AutorizationScopes attributes

## Steps to Reproduce
N/A

## Your Environment
N/A

GoneWacko:
## Description
When using a Cognito User Pool as the authorizer, it is possible to define which authorization scopes should have access to each API method.

When authorization scopes are defined, the authorizer will use an access token, whereas the ID token must be used otherwise. This is only of limited use in the case of Zappa, because it does not define an API method for every endpoint of the underlying app, but at least now authorization scopes can be set globally for the entire API.

Troposphere supported this already, but Zappa did not. This PR resolves that. The new `<stage>.authorizer.scopes` property can be set to a list of scopes that will then be configured for both the `/` and the `/{proxy+}` endpoints.

I tested it and it works. I couldn't find any automated tests that would be impacted or needed extending to cover this new setting.

## GitHub Issues
Closes #1816

### DIFF
--- a/README.md
+++ b/README.md
@@ -1077,6 +1077,23 @@ You can also use AWS Cognito User Pool Authorizer by adding:
 }
 ```
 
+You can also define a list of authorization scopes, which will be applied globally across all endpoints and methods in the API Gateway:
+
+```javascript
+{
+    "authorizer": {
+        "type": "COGNITO_USER_POOLS",
+        "provider_arns": [
+            "arn:aws:cognito-idp:{region}:{account_id}:userpool/{user_pool_id}"
+        ],
+        "scopes": [
+            "com.example/view-products",
+            "com.example/view-users"
+        ]
+    }
+}
+```
+
 #### API Gateway Resource Policy
 
 You can also use API Gateway Resource Policies. Example of IP Whitelisting:
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1535,6 +1535,7 @@ class Zappa(object):
                                     api_name=None,
                                     api_key_required=False,
                                     authorization_type='NONE',
+                                    authorization_scopes=None,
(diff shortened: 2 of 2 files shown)


GoneWacko:
The PR adds 9 lines of code; I have some serious doubts about the coverage decrease being that high.

ffleandro:
Is this going to be merged?

ta2fb:
This would be a useful feature - are there plans to merge this?

PR #1937, opened 2019-10-01
Use shlex.split for parsing django manage commandline arguments

### LINKED ISSUE #1936: manage command does not handle arguments with spaces properly
## Context

Attempting to invoke a Django management command where one of the arguments contains spaces, like `zappa manage prod mycommand '\"a b c\"'` results in the argument with spaces being split.

## Expected Behavior

Arguments that are quoted should be preserved, in this case `mycommand` should be called with the argument `a b c`.

## Actual Behavior

In this case `mycommand` is called with the arguments `"a`, `b`, `c"`.

## Possible Fix

[The handler code](https://github.com/Miserlou/Zappa/blob/60fbb55fffa762a85e79e756f2a1373832d78320/zappa/handler.py#L413) is currently calling `*event['manage'].split(' ')`, but it should probably be using `shlex.split(*event['manage'])` which will do the right thing here:

```
>>> import shlex
>>> s = 'a b "c d"'
>>> s.split(' ')
['a', 'b', '"c', 'd"']
>>> shlex.split(s)
['a', 'b', 'c d']
```

## Steps to Reproduce

Run a django management command, pass an argument containing spaces, such as `zappa manage prod mycommand '\"a b c\"'`

## Your Environment
* Zappa version used: `0.48.2`

luser:
Fixes #1936

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -9,6 +9,7 @@ import inspect
 import json
 import logging
 import os
+import shlex
 import sys
 import traceback
 import tarfile
@@ -410,7 +411,7 @@ class LambdaHandler(object):
 
             # Couldn't figure out how to get the value into stdout with StringIO..
             # Read the log for now. :[]
-            management.call_command(*event['manage'].split(' '))
+            management.call_command(*shlex.split(event['manage']))
             return {}
 
         # This is an AWS-event triggered invocation.


luser:
I tested our app using a version of Zappa with this fix applied and it fixes the issue for me.

rollcat:
Using `shlex.split` is wrong. The process that calls `zappa manage` is already supplying an array of strings (the `argv` in `main(argc, argv)`). This array (the elements following the `--`) should never have been concatenated in the first place, but passed around verbatim.

PR #1942, opened 2019-10-16
Remove api_gateway policy from template creation

aabdullah-bos:
The PR #1903 attempted to fix issue #1747, however, the
travis build failed because at the time `botocore`, required
a version of `docutils` >= 0.10 and <0.15. `botocore` has
since changed the requirements to docutils>=0.10,<0.16. So

This should fix the build issue for the PR to address #1747.

## Description
Removed `policy` from template creation. This fix was attempted in PR #1903, but failed because of `botocore` requirements for `docutils` at the time.

## GitHub Issues
#1747

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -663,7 +663,6 @@ class ZappaCLI(object):
                                             authorizer=self.authorizer,
                                             cors_options=self.cors,
                                             description=self.apigateway_description,
-                                            policy=self.apigateway_policy,
                                             endpoint_configuration=self.endpoint_configuration
                                         )
 


aabdullah-bos:
@Miserlou So all this is a simple change and all of the build jobs passed except for one, [Python: 3.6 amd64 TESTCASE=tests/tests_docs.py](https://travis-ci.org/Miserlou/Zappa/jobs/598774578). Two questions:

1. What's the procedure re-running the build?
2. I made a one line change which should not have affected the code coverage yet, there appears to be a decrease in code coverage. Is this because of the test failure?

PR #1947, opened 2019-10-25
Handle pagination during zappa rollback

### LINKED ISSUE #1858: Rollback doesn't handle more than 50 lambda versions
AWS list_versions_by_function is paginated, and returns 50 results by default (sorted oldest to newest).  If the function has more than 50 versions, a rollback will not go to the most recent deployed. 

## Context
I have >200 versions, when I rollback, it goes back to version 146 every time. This is unexpected behavior for users, and especially dangerous because the user expects to have a failsafe rollback if needed, but it is a false security blanket since it won't really revert to the version expected. The cause is that the list_versions_by_function call is paginated and limited to 50, but Zappa does not handle paginating that request (in this case), althought it is handled in another place (see https://github.com/parth-verma/Zappa/blob/3d6839ee4e4330ede0afd0299ef34ef68bb81659/zappa/core.py#L1090)

## Expected Behavior
Rollback should handle version pagination correctly and only rollback to the most recent version.

## Actual Behavior
Rollback goes the most recent version on the first page of list_versions_by_function.

## Possible Fix
Handle pagination for the AWS request.

## Steps to Reproduce
1. zappa update production x 51 times
2. View versions in AWS console
3. zappa rollback production
4. Verify version is not the most recent version

## Your Environment
* Zappa version used: 0.48.2
* Operating System and Python version: python 2.7, Mac

beheh:
## Description
This PR makes the `zappa rollback` command handle the paginated response as returned by the AWS API. Previously, rollback could not handle more than one page of versions, leading to `zappa rollback` potentially rolling back to a really old version.

## GitHub Issues

- Fixes #1858.

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -1264,16 +1264,22 @@ class Zappa(object):
 
         Returns the Function ARN.
         """
-        response = self.lambda_client.list_versions_by_function(FunctionName=function_name)
+        versions_in_lambda = []
+        versions = self.lambda_client.list_versions_by_function(FunctionName=function_name)
+        for version in versions['Versions']:
+            versions_in_lambda.append(version['Version'])
+        while 'NextMarker' in versions:
+            versions = self.lambda_client.list_versions_by_function(
+                FunctionName=function_name, Marker=versions['NextMarker'])
+            for version in versions['Versions']:
+                versions_in_lambda.append(version['Version'])
 
-        # Take into account $LATEST
-        if len(response['Versions']) < versions_back + 1:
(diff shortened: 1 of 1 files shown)


beheh:
I'd like some guidance on how to write a test for this please - it looks like `zappa rollback` is only actually tested by `tests_placebo.py` so far, and I'm not sure why these tests are commented out:

https://github.com/Miserlou/Zappa/blob/d53fe4750a5cebd5fb9d92cbe2719ed9c62764c0/tests/tests.py#L960-L975

Issue #1917, opened 2019-08-08
S3 resource does not exist causing module not found


themmes:
## Context

I do not understand why I am getting the following error. It seems the lambda cannot be loaded from S3 and therefore is not able to find the module to execute the function.

```
[1565266143895] [DEBUG] 2019-08-08T12:09:03.895Z ef9eb08e-f4f9-43a5-9072-c11f15be1b80 Changing event name from docs.*.cloudsearchdomain.Search.complete-section to docs.*.cloudsearch-domain.Search.complete-section
[1565266143897] The 's3' resource does not exist.
The available resources are:
 -
: ResourceNotExistsError
Traceback (most recent call last):
  File "/var/task/handler.py", line 602, in lambda_handler
  return LambdaHandler.lambda_handler(event, context)
  File "/var/task/handler.py", line 245, in lambda_handler
  handler = cls()
  File "/var/task/handler.py", line 102, in __init__
  self.load_remote_project_archive(project_archive_path)
  File "/var/task/handler.py", line 170, in load_remote_project_archive
  s3 = boto_session.resource('s3')
  File "/var/task/boto3/session.py", line 347, in resource
  has_low_level_client)
boto3.exceptions.ResourceNotExistsError: The 's3' resource does not exist.
The available resources are:
 -
[...]

Issue #1549, opened 2018-06-26
zappa package fails with zappa from git and slim_handler


jneves:
## Context

If you're trying to run a zappa as an editable from a git repo, it works, except if you have slim_handler enabled.
## Expected Behavior
It should build a zip with zappa inside.

## Actual Behavior
It fails with a traceback like this when building the handler_venv: 
```
Traceback (most recent call last):
  File "/Users/joao/testes/zappa-1/1/ve/src/zappa/zappa/cli.py", line 2693, in handle
    sys.exit(cli.handle())
  File "/Users/joao/testes/zappa-1/1/ve/src/zappa/zappa/cli.py", line 504, in handle
    self.dispatch_command(self.command, stage)
  File "/Users/joao/testes/zappa-1/1/ve/src/zappa/zappa/cli.py", line 543, in dispatch_command
    self.package(self.vargs['output'])
  File "/Users/joao/testes/zappa-1/1/ve/src/zappa/zappa/cli.py", line 633, in package
    self.create_package(output)
  File "/Users/joao/testes/zappa-1/1/ve/src/zappa/zappa/cli.py", line 2171, in create_package
    venv=self.zappa.create_handler_venv(),
  File "/Users/joao/testes/zappa-1/1/ve/src/zappa/zappa/core.py", line 403, in create_handler_venv
    copytree(os.path.join(current_site_packages_dir, z), os.path.join(venv_site_packages_dir, z))
  File "/Users/joao/testes/zappa-1/1/ve/src/zappa/zappa/utilities.py", line 37, in copytree
    lst = os.listdir(src)
NotADirectoryError: [Errno 20] Not a directory: '/Users/joao/testes/zappa-1/1/ve/lib/python3.6/site-packages/zappa.egg-link
```

## Possible Fix
Use the function for copying editables as when slim_handler is false.

## Steps to Reproduce
With the following zappa_settings.json:
```
{
    "dev": {
        "project_name": "1",
        "runtime": "python3.6",
	"slim_handler": true
    }
}
```
Run on a virtualenv: `pip install -e git+https://github.com/miserlou/zappa#egg=zappa`
and then `zappa package`.

## Your Environment
* Zappa version used: 0.46.1 from git
* Operating System and Python version: Mac OS X 10.12.6, python 3.6.4 (from home brew)
* The output of `pip freeze`:
```
argcomplete==1.9.3
base58==1.0.0
boto3==1.7.45
botocore==1.10.45
certifi==2018.4.16
[...]

PR #1960, opened 2019-11-11
Allow full path exclusions while packaging

jneves:
## Description
Two changes (sorry, but needed the second one to test): add a new option "exclude_glob" that takes a list of glob patterns for file removals and a fix to adding editable packages with slim_handler.

## GitHub Issues
#1917 
#1549


mathom:
I've tested this a few different ways and it's working well. I do think the naming might be a little confusing vs the existing exclude feature, so maybe we need to brainstorm a bit on clearing up the documentation or naming conventions. Otherwise :+1:

jneves:
@mathom agreed, I think the excludes should probably be called exclude_files or something like that. Would need a couple of releases to make that name change though.

jackric:
Thanks for adding this, but PLEASE DOCUMENT IT IN THE README!!!!

PR #1970, opened 2019-12-06
Use safe load with yaml

jtratner:
## Description
Really small change to get rid of deprecation warning for YAML due to this - https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation. I double-checked in yaml repo and `safe_load` was added 13 years ago!

Tested this via cloning zappa client and running it locally, where you no longer see DeprecationWarning.  I _think_ this is probably okay to go without tests.

## GitHub Issues
Since so small I thought okay to open without a ticket.


jneves:
@jtratner Thanks!

PR #1978, opened 2020-01-03
Update version to 0.48.3 (removes hard limitations from dependencies)

cheedep:
https://github.com/Mattlk13/Zappa/pull/6

Removes hard dependencies.

Allows getting lambda-packages 0.21.0 thus getting newer packages


jneves:
Slved with #1998

PR #1993, opened 2020-02-07
Fixes issue that makes API gateway duplicate set-cookie headers

### LINKED ISSUE #1965: Set_Cookie option sets duplicate cookies on AWS Lambda
## Context
I have an API running Python3.7 and Zappa (in a virtualenv).
I am setting 6 cookies by using the option "set_cookie" in flask. It looks something like this:
```
resp = make_response(jsonify({'success':'true', 'message': 'Successfully authenticated!'}), 200)
resp.set_cookie("1", value="1", secure=True, samesite='Lax', domain=".example.com",max_age=3600)
resp.set_cookie("2", value="2", secure=True, samesite='Lax', domain=".example.com",max_age=3600)
resp.set_cookie("3", value="3", secure=True, samesite='Lax', domain=".example.com",max_age=3600)
resp.set_cookie("4", value="4", secure=True, samesite='Lax', domain=".example.com",max_age=3600)
resp.set_cookie("5", value="5", secure=True, samesite='Lax', domain=".example.com",max_age=3600)
resp.set_cookie("6", value="6", secure=True, samesite='Lax', domain=".example.com",max_age=3600)
return resp
```

On localhost testing Flask, this works as expected.

If I deploy the same code to AWS using Zappa, the response header will show 36 "set-cookie" headers. So the formula here is n^2. So if I add 4 cookies using the above method, it will show 16 in the request header.

The browser takes care of duplicate cookies, but the response from the API is still huge because of this issue.

Same thing happens if I use: 
`resp.headers.add("set-cookie""1"="1; Domain=.example.com; Max-Age=3600; Secure; Path=/; SameSite=Lax")`

## Expected Behavior
I believe Zappa or something at AWS is at fault here. Expected behaviour is to send 6 "set-cookie" headers and not 36.

## Actual Behavior
Sets n^2 cookies as response.

## Steps to Reproduce
Deploy a Flask route using Zappa which sets the cookies. Use the code above.

## Your Environment
* Zappa version used: 0.48.2
* Operating System and Python version: Ubuntu 18.04, Python3.7
* The output of `pip freeze`: https://pastebin.com/d4QTaTuG
* Your `zappa_settings.py`: https://pastebin.com/d1GK8sbe

thomasgustavo:
## Description
Fixes a issue that makes API Gateway duplicate set-cookie response headers when they are sent in multiple casing (like Set-cookie, set-cookie, set-Cookie).

## GitHub Issues
fixes #1965

### DIFF
--- a/zappa/middleware.py
+++ b/zappa/middleware.py
@@ -38,32 +38,17 @@ class ZappaWSGIMiddleware(object):
 
         def encode_response(status, headers, exc_info=None):
             """
-            Create an APIGW-acceptable version of our cookies.
-
-            We have to use a bizarre hack that turns multiple Set-Cookie headers into
-            their case-permutated format, ex:
-
-            Set-cookie:
-            sEt-cookie:
-            seT-cookie:
-
-            To get around an API Gateway limitation.
-
-            This is weird, but better than our previous hack of creating a Base58-encoded
-            supercookie.
+            This makes the 'set-cookie' headers name lowercase,
+            all the non-cookie headers should be sent unharmed.
+            Related: https://github.com/Miserlou/Zappa/issues/1965
             """
 
-            # All the non-cookie headers should be sent unharmed.
-            
-            # The main app can send 'set-cookie' headers in any casing
-            # Related: https://github.com/Miserlou/Zappa/issues/990
             new_headers = [header for header in headers
                            if ((type(header[0]) != str) or (header[0].lower() != 'set-cookie'))]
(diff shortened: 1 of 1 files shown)


thomasgustavo:
The previous code was a hack to deliberately send the set-cookie headers names in different cases. This was fixed by AWS, as described here: https://forums.aws.amazon.com/ann.jspa?annID=6185

While fixing the problem, it started to duplicate the headers with different casing. So this PR fixes that.

Issue #1997, opened 2020-02-12
Using pip-tools to manage dependencies


jneves:
Currently zappa has several pinned dependencies that make it likely that it works, but also makes it hard to keep using as time passes because of other packages dependency updates.

For this I believe we should be separating the hard requirements from the pinned ones.

pip-tools allows us to do that, while not having two collections of requirements that need to be updated individually.

## Context

See many complaints throughout the years.

## Expected Behavior
Installing zappa should work with an open-ended set of dependencies until a specific one doesn't work.

## Actual Behavior
Some installations only work if people ignore the dependency requirements.

## Possible Fix
With pip-tools, updating a requirement would be adding it in requirements.in or test_requirements.in as needed and then running:

`pip-compile -o test_requirements.txt requirements.in test_requirements.in && cp test_requirements.txt requirements.txt && pip-compile -o requirements.txt requirements.in`

monkut:
Just a random comment about my experience with pipenv.

I've been using [pipenv](https://github.com/pypa/pipenv) for a little over a year now for most of my projects.

It has built in support for (standard) requirements, `packages`,  and (test) requirements `dev-packages`.  It also allows you to maintain a list of requirement versions unbound by their dependency versions, but maintains the ability to create a version exact _build_ using the _sync_ command.

sample Pipfile:
```
[[source]]
url = 'https://pypi.python.org/simple'
verify_ssl = true
name = 'pypi'

[requires]
python_version = '2.7'

[packages]
requests = { extras = ['socks'] }
records = '>0.5.0'
django = { git = 'https://github.com/django/django.git', ref = '1.11.4', editable = true }
"e682b37" = {file = "https://github.com/divio/django-cms/archive/release/3.4.x.zip"}
"e1839a8" = {path = ".", editable = true}
pywinusb = { version = "*", os_name = "=='nt'", index="pypi"}

[dev-packages]
nose = '*'
unittest2 = {version = ">=1.0,<3.0", markers="python_version < '2.7.9' or (python_version >= '3.0' and python_version < '3.4')"}

```

It seems similar to what you are doing here to manage test/development requirements.

These automatically update the related `Pipfile` and `Pipfile.lock` files, which would replace the traditional `requirements.txt` files.
```
# Adding a requirement -- adds to Pipfile [packages]
pipenv install desired-pypi-package

# Adding development/test specific package -- adds to Pipfile [dev-packages]
pipenv install desired-pypi-package --dev
```

[...]

PR #1998, opened 2020-02-12
Using pip-tools to manage dependencies

jneves:
## Description
Uses pip-tools to manage the dependencies and makes setup.py only use the hard requirements from the in files instead of from the txt files.

## GitHub Issues

#1997


mathom:
This works great. I've also tested overriding some of the packages used in Zappa with my own project's versions and the install doesn't interfere with them.

Issue #1985, opened 2020-01-26
Warning! Status check on the deployed lambda failed. A GET request to '/' yielded a 500 response code.


selimppc:
I did:
1. created a IAM with full `administrativeAccess`
2. Install Zappa on venv 
3. Init Zappa 
4. Zappa deploy dev

Now getting error. I am confusing with AWS ROLE and Groups as well.

ERROR:

Deploying API Gateway..
Error: Warning! Status check on the deployed lambda failed. A GET request to '/' yielded a 500 response code.

Here is the full result: 
```
(venv) selim@sr ms % zappa deploy dev
Calling deploy for stage dev..
Downloading and installing dependencies..
 - typed-ast==1.4.1: Using locally cached manylinux wheel
 - pycrypto==2.6.1: Using precompiled lambda package 
 - psycopg2-binary==2.8.4: Using locally cached manylinux wheel
 - markupsafe==1.1.1: Using locally cached manylinux wheel
 - lazy-object-proxy==1.4.3: Using locally cached manylinux wheel
 - greenlet==0.4.15: Using locally cached manylinux wheel
 - gevent==1.4.0: Using locally cached manylinux wheel
 - cryptography==2.8: Warning! Using precompiled lambda package version 1.9 instead!
 - coverage==5.0.3: Using locally cached manylinux wheel
 - cffi==1.13.2: Using locally cached manylinux wheel
 - sqlite==python3: Using precompiled lambda package
Packaging project as gzipped tarball.
WARNING: Target directory /Users/selim.bongo/Sites/python/ms/handler_venv/lib/python3.6/site-packages/zappa already exists. Specify --upgrade to force replacement.
WARNING: Target directory /Users/selim.bongo/Sites/python/ms/handler_venv/lib/python3.6/site-packages/zappa-0.48.2.dist-info already exists. Specify --upgrade to force replacement.
Downloading and installing dependencies..
 - sqlite==python3: Using precompiled lambda package
Packaging project as zip.
Uploading dev-1580015449.tar.gz (41.0MiB)..
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 43.0M/43.0M [00:11<00:00, 3.64MB/s]
Uploading handler_dev-1580015525.zip (11.7MiB)..
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 12.2M/12.2M [00:01<00:00, 6.32MB/s]
[...]

PR #1999, opened 2020-02-13
support for ABI3 downloads

jneves:
## Description
Add manylinux1/2010/2014 download support, ABI3 download support. Based on #1886 but with support for versions other than 3.4 and with a test case. Ended up deleting the 2.7 support also as it isn't working.

## GitHub Issues
#1861 
#1985

### DIFF
--- a/zappa/core.py
+++ b/zappa/core.py
@@ -284,15 +284,18 @@ class Zappa(object):
         self.runtime = runtime
 
         if self.runtime == 'python3.6':
-            self.manylinux_wheel_file_suffix = 'cp36m-manylinux1_x86_64.whl'
+            self.manylinux_suffix_start = 'cp36m'
         elif self.runtime == 'python3.7':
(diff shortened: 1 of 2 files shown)


zappa/core.py
```diff
@@ -293,6 +292,8 @@ def __init__(self,
             # See https://github.com/pypa/manylinux for a more detailed explanation
             self.manylinux_wheel_file_suffix = 'cp38-manylinux1_x86_64.whl'
 
+        self.manylinux_wheel_abi3_file_match = re.compile('^.*cp3.-abi3-manylinux{1,2010}_x86_64.whl$')
```
mathom: It looks like there is also a 2014 platform tag - see https://www.python.org/dev/peps/pep-0599/

jneves: Yeah, and I checked and AWS Lambda supports it. Rebuilt the manylinux support so it works with all the 3 tags and added caching support for ABI3 ones. I think it's better now.

Issue #1994, opened 2020-02-08
Drop support for Python versions < 3.6


mbeacom:
Python 2.x is now end of life and will no longer receive updates.

##  Context

Python 2.x and 3.{4, 5} are either no longer supported or don't have official AWS Lambda runtimes.

Some additional references in favor of this change:

- [Python Clock - Lapsed](https://pythonclock.org/)
- [Python 3 Statement](https://python3statement.org/)
- [Pandas drops Python 2 support - July 2019](https://pandas.pydata.org/#v0-25-0-final-july-18-2019)
- [NumPy drops Python 2 support - Proposed January 2019](https://numpy.org/neps/nep-0014-dropping-python2.7-proposal.html)
- [Pytest drops Python 2 support](https://pytest.readthedocs.io/en/latest/py27-py34-deprecation.html)
- [PEP-373 - 2.7 Lifecycle](https://www.python.org/dev/peps/pep-0373/)

## Possible Fix
I propose we add a deprecation notice in the next patch for 0.49.x and increment the minor (`0.49.0` -> `0.50.0`), dropping support for Python versions `< 3.6` by setting python min requirement, removing six/compat support, and convert relevant code to Python 3.6+.

jneves:
So, this has been done more by accident than I would have liked, as 0.49 doesn't work on 2.7 and that was missed it on the preparation of the release. As nobody is thinking of going back, I'll update the README and Changelog to note that 0.48.2 was the lst one with 2.7 support. Sorry for the unexpected update.

PR #2000, opened 2020-02-13
Remove python 2.7 support

jneves:
## Description
So, as we retired python 2.7 by accident on 0.49.0 we might as well finish the job. This removes all support for 2.7.

## GitHub Issues
#1994


tests/tests.py
```diff
@@ -188,20 +175,16 @@ def test_get_manylinux_python38(self):
...
-        self.assertTrue(z.have_any_lambda_package_version('mysql-python'))
+        self.assertTrue(z.have_any_lambda_package_version('mysqlclient'))
         self.assertFalse(z.have_any_lambda_package_version('no_package'))
 
+        self.assertTrue(z.have_correct_lambda_package_version('mysqlclient', '1.3.12'))
+        self.assertFalse(z.have_correct_lambda_package_version('mysqlclient', '1.4.12'))
+
     def test_getting_installed_packages(self, *args):
-        z = Zappa(runtime='python2.7')
+        z = Zappa(runtime='python2.6')
```
mbeacom: Should this be `3.6` ?

requirements.txt
```diff
@@ -2,7 +2,6 @@ argcomplete>=1.9.3
 boto3>=1.4.7
 durationpy>=0.5
 future>=0.16.0
```
mbeacom: We should be able to remove `future` as well, no?

tests/tests.py
jneves: Oops, fixed.

requirements.txt
jneves: Yes, changed that in #1998 though.

jneves: Reverted on #1998 - both future and six will require more careful reviews for removal.

mathom: I'd definitely like to clean up the imports to get rid of `future/past` cludge but I feel like that might be a bit further down the road. If we broke 2.7 already, just getting rid of it officially in the simplest manner makes the most sense to me.

PR #2002, opened 2020-02-17
Dependencies update

jneves:
Just what it says on the tin.


mbeacom:
Why not just use poetry for this and simplify package management *and* publishing?  This would align with the 3.x theme and introduce a pyproject.yml over setup.py.

Issue #2003, opened 2020-02-17
Migrate Zappa to pyproject.toml - PEP 518


mbeacom:
## Context

Currently, Zappa is using the setup.py, setup.cfg, test_requirements.txt, and requirements.txt to handle Zappa packaging, dependencies, and publishing.

[PEP-518](https://www.python.org/dev/peps/pep-0518/)

## Possible Fix

Migrate to use [Poetry](https://github.com/python-poetry/poetry), removing setup.py in favor of the new pyproject.toml

mbeacom:
Opting for legacy package handling...

PR #2005, opened 2020-02-18
Migrate to Poetry / pyproject.toml

mbeacom:
## Description

The goal of this PR is to replace the `requirements.txt`, `test_requirements.txt` and `setup.py` with a `pyproject.toml`.

This simplifies installation, building, and publishing of the `zappa` package.  Additionally, `poetry` is used to maintain virtual environments and handle package management.

## References

- [PEP-518](https://www.python.org/dev/peps/pep-0518/)
- [Poetry Github](https://github.com/python-poetry/poetry) / [Poetry Website](https://python-poetry.org/)

## GitHub Issues

#2003 
#1997


jneves:
Sorry, not going to happen. The reason is that I want a set of dependencies in production to be a subset of the test/dev ones, so the one we release is a tested set and poetry does not give us that. Thanks for your work.

mbeacom:
@jneves What do you mean?  That's the purpose of `dev` versus standard packages with Poetry.  `--no-dev` is equivalent to production only dependencies.
🤷‍♂

jneves:
@mbeacom I know that's what people tell me. I tested it and it didn't come through and I was told it was not a bug.

jneves:
@mbeacom now that I have a minute, the case I tried was prod dependency A depends on package B (no version limitation), dev dependency C depends on B<2. Installation without dev install B==2 and with dev would install B==1 (first issue was with Zappa and tqdm, but I replicated with some other packages).

PR #2006, opened 2020-02-18
Remove the use of lambda-packages

jneves:
## Description
With the drop of python 2.7, lambda-packages stopped being useful.


mcrowson:
I like it. Remove from requirements as well?

jneves:
@mcrowson oops, done.

havocbane:
This breaks backward compatibility, which wouldn't be a big deal if the version numbering followed semvar, but since it doesn't it caused us a major headache today.

mcrowson:
How did you fix it?

havocbane:
We pinned to version 0.49.0 at first. That worked short-term, but we didn't want to stay pinned so we then later spent the effort to upgrade and remove lambda-packages and upgrade to 0.51. My point above is more that breaking changes should be a major version increment.

daneah:
Do any folks involved here know how one would best go about pre-compiling `mysqlclient` (on macOS, but for use on Lambda) these days? Hadn't deployed my app in some time and wondering how I might be able to modernize instead of pinning hard on a few things.

Issue #1879, opened 2019-05-31
Configuration of cognito event's trigger fails


Yaronn44:
## Context
I wanted to use the cognito event trigger by declaring it in the `zappa_setting.py` with this form :
```        
"cognito": {
            "user_pool": "user-pool-id", 
            "triggers": [{
                "source": "PreAuthentication_Authentication", 
                "function": "my_app.pre_authentication_function"
            }]
        }
```
## Expected Behavior
The zappa update should succeed, and the trigger should be created in the cognito pool.

## Actual Behavior
At the zappa update i have this error that trigger : 
```
Traceback (most recent call last):
  File "c:\***\***\venv\lib\site-packages\zappa\cli.py", line 2712, in handle
    sys.exit(cli.handle())
  File "c:\***\***\venv\lib\site-packages\zappa\cli.py", line 509, in handle
    self.dispatch_command(self.command, stage)
  File "c:\***\***\venv\lib\site-packages\zappa\cli.py", line 556, in dispatch_command
    self.update(self.vargs['zip'], self.vargs['no_upload'])
  File "c:\***\***\venv\lib\site-packages\zappa\cli.py", line 1016, in update
    self.update_cognito_triggers()
  File "c:\***\***\venv\lib\site-packages\zappa\cli.py", line 1138, in update_cognito_triggers
    self.zappa.update_cognito(self.lambda_name, user_pool, lambda_configs, self.lambda_arn)
  File "c:\***\***\venv\lib\site-packages\zappa\core.py", line 1803, in update_cognito
    result = self.cognito_client.update_user_pool(UserPoolId=user_pool, **description_kwargs)
  File "c:\***\***\venv\lib\site-packages\botocore\client.py", line 357, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "c:\***\***\venv\lib\site-packages\botocore\client.py", line 661, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.InvalidParameterException: An error occurred (InvalidParameterException) when calling the UpdateUserPool operation: Please use TemporaryPasswordValidityDays instead of UnusedAccountValidityDays
```
An associate of mine did the configuration 2 or 3 weeks ago and everything worked. 

## Possible Fix
In boto3 documentation it is said [here](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cognito-idp.html#CognitoIdentityProvider.Client.update_user_pool) that : 
```
Note
[...]

PR #2027, opened 2020-02-19
Lambda Layers & Cognito support

millarm:
## Description
This merges the following PRs with the 0.50.0 release

https://github.com/Miserlou/Zappa/pull/1842

To add support for Lambda layers

And fixes
https://github.com/Miserlou/Zappa/issues/1879


millarm:
And... the only coverage change here is the fact that "there is a new version of Zappa" is no longer covered!  :-)

PR #2028, opened 2020-02-20
Fix parsing of X-Forwarded-For header

jait:
When the header did not contain a comma, it was not parsed and
REMOTE_ADDR was always set to 127.0.0.1. This bug occurred when the project was deployed without a CloudFront, e.g. with an ALB only.


jneves:
Thanks!

Issue #1871, opened 2019-05-11
bug: Error condition check for binary support


lonsdale8734:
https://github.com/Miserlou/Zappa/blob/3ccf7490a8d8b8fa74a61ee39bf44234f3567739/zappa/handler.py#L542-L551

The condition `not response.mimetype.startswith("text/")  or response.mimetype != "application/json"` will never happen.

Issue #1605, opened 2018-09-11
All responses are base64 encoded


joeldentici:
It seems that Zappa always responds with a base64 encoded string for the response body and "isBase64Encoded" set to true.

The relevant file is the "handler.py" file provided/generated by Zappa.

On line 513, we have this if statement:
```python
if not response.mimetype.startswith("text/") \
    or response.mimetype != "application/json":
```

The condition's negation is:
```python
response.mimetype.startswith("text/") and response.mimetype == "application/json"
```

Clearly this is a contradiction, so the condition itself is a tautology. The true branch is taken and Zappa will always base64 encode responses.

It seems that the intention of the condition is to NOT base64 encode responses that are text (but to take a conservative approach to determine what is "text").

It seems that what we want is to NOT base64 encode when this condition is true:
```python
response.mimetype.startswith('text/') or response.mimetype == 'application/json'
```

If this is the case, the condition starting on line 513 should instead be:
```python
not response.mimetype.startswith('text/') and response.mimetype != 'application/json'
```

I noticed this while using Zappa 0.46.2

erindan:
Worth noting that this encoding bug seemingly makes it impossible to do API Gateway level compression of responses. Perhaps this needs a priority bump?

douglasjreynolds:
This seems like a pretty big bug.  I was unable to get anything but base64 responses until I disabled binary_support.

vitodsk:
+1
In my case this problem was preventing to have a normal deployment on AWS using Api Gateway Lambda proxy integration. There was a 40% overhead in the computation of the encoding/decoding.

ronenh:
I ran into the same issue and worked around it by adding `"binary_support": false` to my `zappa_settings.json`.
I suppose this may not work if any of your API endpoints do in fact return binary content but that wasn't the case for me.

monkut:
Just want to note that the "Binary Media Types" RestAPI Settings for the ApiGateway if defined as "*/*" may cause APIGateway to encode incoming request.body to b64.

monkut:
I put a PR together to work on some of the issues around this.  It may only work for my use case though...

https://github.com/Miserlou/Zappa/pull/2170

PR #2029, opened 2020-02-20
Fix binary_support handling

jait:
## Description

The logic of handling response data when binary_support is True, was faulty. It base64-encoded every response, regardless of the content-type, which caused the user/browser to receive base64-encoded bodies if deployed with an API Gateway. ALB apparently supports base64-encoded responses and manages to decode them.

## GitHub Issues

Relevant tickets:
- https://github.com/Miserlou/Zappa/issues/1605
- https://github.com/Miserlou/Zappa/issues/1871

### DIFF
--- a/zappa/handler.py
+++ b/zappa/handler.py
@@ -540,13 +540,11 @@ class LambdaHandler(object):
                         zappa_returndict.setdefault('statusDescription', response.status)
 
                     if response.data:
-                        if settings.BINARY_SUPPORT:
-                            if not response.mimetype.startswith("text/") \
-                                or response.mimetype != "application/json":
-                                    zappa_returndict['body'] = base64.b64encode(response.data).decode('utf-8')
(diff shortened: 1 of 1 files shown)


jneves:
Thanks!

martinv13:
There was indeed a bug in the logic with the condition being always true as discussed in #1605 .

However this PR broke my setup where response is compressed at application level (with `flask_compress`). The response is forced through `response.get_data(as_text=True)`, which fails with gzipped payload.

I think a lighter change (just changing the "or" for a "and" in the following condition: `not response.mimetype.startswith("text/")  or response.mimetype != "application/json"`) would have fixed the problem without breaking the use case where the response is gzipped text.

I can send a new PR if you agree.
Thanks in advance!

mzaglia:
I'm having the same problem as @martinv13 after compressing my response.

PR #2037, opened 2020-02-21
add update possibility to script and new requirements

jneves:
New set of dependencies update.

Also changed the requirements.sh script so that running `./requirements.sh -U` will update the requirements.

### DIFF
--- a/requirements.sh
+++ b/requirements.sh
@@ -2,6 +2,11 @@
 
 set -e
 
-pip-compile $* -o test_requirements.txt requirements.in test_requirements.in
+ARGS=""
+if [ "$1" == "--upgrade" ]; then
+    ARGS="-U"
+fi
+
+pip-compile ${ARGS} -o test_requirements.txt requirements.in test_requirements.in
 cp test_requirements.txt requirements.txt
 pip-compile -o requirements.txt requirements.in
--- a/requirements.sh
+++ b/requirements.sh
@@ -2,6 +2,6 @@
 
 set -e
 
-pip-compile -o test_requirements.txt requirements.in test_requirements.in
+pip-compile $* -o test_requirements.txt requirements.in test_requirements.in
 cp test_requirements.txt requirements.txt
 pip-compile -o requirements.txt requirements.in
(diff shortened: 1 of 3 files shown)


jneves:
@mbeacom my requirements are in this gist https://gist.github.com/jneves/01bfc467016fa49789a669e707a31aaa - I did try poetry, pipenv and others. I've looked for several months for alternatives. This was the only solution I found that didn't require me to code it. I'd love to be proved wrong, seriously.

mbeacom:
I suppose my argument is why do you need three uniquely separate sets of requirements. Some thought provoking questions...

- Why wouldn't development and test package requirements be the same?
- Do we maintain three separate Docker images? If not, why would we maintain such granular package requirements, but not require similar contrast in unique operating landscapes?
- What added value does such granular package management provide? How does this compare to the exponentially complex management overhead and toil associated with maintaining these dependencies and subsequent updates (as it applies to open source sustainability, longevity, and the impact on the OSS community of this project as a whole)?
- What does the ideal evolution of this look like? What genuine problems does that solution solve? Is the problem a problem that needs to be addressed as a regular part of this module's maintenance now and in the future?

jneves:
Hi,

1. Because there are packages like black that don't make sense to me to load on tests and initial installation of packages can delay tests (and that's my main limitation in time usage at the moment on this project, limiting how much I can put in each release).
2. We don't have any docker images. I do expect that we will have 6 supported environments in the medium term with local development tools, testing and production environments for both the zappa package and a zappa-handler package (a smaller one that will be in the zip).
3. This question is against a single set of requirements, right? Less security surface for remote exploitation of running lambdas. The overhead is actually lower than the previous situation as you only need to put the hard requirements in the .in files and you only need to test a dependency update where all the package updates are compatible. I can now do a weekly test and update of the couple of dependencies than changed, versus having to manage 8 PRs from doppins where most of them don't respect the dependencies from the other packages. So it's better and more sustainable because it takes less work.
4. Yes, increasing the default security of deployed lambdas is in the future of this project and it goes together with changes as defining and then automate a "production mode".

On the arguments, you're absolutely right, will remove that option to the 2 use cases I actually use: adding a dependency and upgrading.

Thanks for pushing with this @mbeacom I feel like getting these arguments and discussions out in the open is best for the project.

jneves:
@mbeacom Fixed the script.

Issue #2052, opened 2020-02-28
register_targets fails if application load balancer state is not 'active'


dsterne:
## Context
When running a new deploy, the `register_targets` step fails if it's reached before the Application Load Balancer is in active state. 

## Expected Behavior
Wait for Application Load Balancer to become active before continuing.

## Actual Behavior
`deploy_lambda_alb` currently gets a waiter using `get_waiter`, but never actually calls `waiter.wait`

## Possible Fix
At line 1341 in `core.py`, add the following:

```Python
        waiter = self.elbv2_client.get_waiter('load_balancer_available')
        print('Waiting for load balancer [{}] to become active..'.format(load_balancer_arn))
        waiter.wait(LoadBalancerArns=[load_balancer_arn], WaiterConfig={"Delay": 3})
```

## Your Environment
* Zappa version used: 0.48.2, 0.49.0, 0.50.0
* Operating System and Python version: Ubuntu 16.04.6, Python 3.7.5

jneves:
Fixed in 0.51.0.

PR #2053, opened 2020-02-28
Wait for the Load Balancer to become active before proceeding

dsterne:
If the load balancer is not 'active', the subsequent commands will fail

## Description
Calls `waiter.wait` to wait for the ALB to become active before proceeding

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/2052


jneves:
@dsterne can you update the tests too, please?

PR #2049, opened 2020-02-28
Fix syntax warnings due to comparison of literals using is.

tirkarthi:
## Description

Fix syntax warnings due to comparison of literals using is.

## GitHub Issues

Fixes #2048

PR #2054, opened 2020-02-29
Use ==/!= to compare str, bytes, and int literals

cclauss:
Add more key flake8 tests to the Travis CI run to detect and avoid Python 3.8 SyntaxWarnings:
% __python3.8__
```
>>> 3 is 3
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
>>> 'LambdaConfig' is 'LambdaConfig'
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
```
https://flake8.pycqa.org/en/latest/user/error-codes.html

On the flake8 test selection, this PR does _not_ focus on "_style violations_" (the majority of flake8 error codes that [__psf/black__](https://github.com/psf/black) can autocorrect).  Instead these tests focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).  Python >= 3.8 will raise SyntaxWarnings on these instances.
* F7 tests logic errors and syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ env:
   - TESTCASE=tests/tests.py
 before_script:
   # stop the build if there are Python syntax errors or undefined names
-  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  - flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
   # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
   - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
 script:
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -1382,7 +1382,7 @@ class ZappaCLI:
             # UUIDs
             for token in final_string.replace('\t', ' ').split(' '):
                 try:
-                    if token.count('-') is 4 and token.replace('-', '').isalnum():
+                    if token.count('-') == 4 and token.replace('-', '').isalnum():
                         final_string = final_string.replace(
                             token,
                             click.style(token, fg='magenta')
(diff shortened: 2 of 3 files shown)


jneves:
@cclauss I can't find a reason related to this PR for the test failure. Can you just rebase to current master just to make sure the tests are passing before I merge? Thank you very much!

cclauss:
This kind of hard-to-find bug is why the SyntaxWarning was put into Python 3.8.

The line that fails Travis tests is the line below the modified line in zappa/core.py.  before this PR, that code was not getting executed because:
```
>>> lc = "Lambda"
>>> lc += "Config"
>>> lc == "LambdaConfig"
True
>>> lc is "LambdaConfig"
False
```

jneves:
@cclauss thanks for the debugging!

cclauss:
I have left the __# noqa__ linter directive with a __TODO__ so that the Travis tests pass and we can land this PR.  Afterwards, it would be great if someone with more knowledge of the codebase can remove the linter directive and fix the code which is expecting a dict but getting a list instead.

Similar to #2049.

jneves:
@cclauss no need, I'm fixing it now, it was just a bug on the tests responses file.

jneves:
@cclauss just waiting for the tests on #2055 to run once then will merge that one and this one

Issue #2050, opened 2020-02-28
Zappa Deployment to AWS Lamda not working with MSSQL


ORC-1:
Django apps using MSSQL as Database deployed using Zappa, fails with error: ImproperlyConfigured("Error loading pyodbc module: %s" % e)s_removeddulele named 'pyodbc'
even if installed and works properly on local environment

## Your Environment

* Zappa version used: zappa 0.50.0
* Operating System and Python version: Linux| python 3.7

nikbora:
I get the same for cryptography package when going from 0.49.0 -> 0.50.0. Looking at the logs it never gets included in the deploy - perhaps related to recent abi3/manylinux changes?

zappa=0.50.0
cryptography=2.8

millarm:
I'm deploying those versions with mysql to lambda fine -> one change that is needed is to ensure the lambda has the appropriate mysql libraries.

I'm doing this at present -> running on CodeBuild on AWS and on an Amazon Linux build before I deploy I do

        cp /usr/lib64/mysql/libmysqlclient.* .


Which ensures that the lambda deployment has the same versions of the mysqlclient libraries that I've run the tests on the CodeBuild instance.

nikbora:
Yeah that works, however shouldn't be (and wasn't) necessary as binaries should get downloaded from PyPi.

I suspect the reason this got through is that these unit tests only check that the packaging works (which it does) but not that the binary was included: 

https://github.com/Miserlou/Zappa/blob/379c0bf822eef91faccaa57fcf0111be21430cb1/tests/tests.py#L180

nikbora:
Adding some Lambda logs to have an example (as perhaps the lib IS included but the binding version is wrong):

```
  ...
  File "/var/task/cryptography/hazmat/backends/__init__.py", line 15, in default_backend
    from cryptography.hazmat.backends.openssl.backend import backend
  File "/var/task/cryptography/hazmat/backends/openssl/__init__.py", line 7, in <module>
    from cryptography.hazmat.backends.openssl.backend import backend
  File "/var/task/cryptography/hazmat/backends/openssl/backend.py", line 16, in <module>
    from cryptography import utils, x509
  File "/var/task/cryptography/x509/__init__.py", line 8, in <module>
    from cryptography.x509.base import (
  File "/var/task/cryptography/x509/base.py", line 18, in <module>
    from cryptography.x509.extensions import Extension, ExtensionType
[...]

PR #2056, opened 2020-03-05
Fix for abi3 filename matching

nikbora:
Partly fixes https://github.com/Miserlou/Zappa/issues/2050 (e.g. cryptography abi3 manylinux not being downloaded from PyPi, so locally compiled version is used in the package instead).


jneves:
Thanks! Sorry for that.

Issue #1359, opened 2018-01-23
Zappa is ignoring some files and modules from installed packages


eldos-dl:
## Context
I've noticed that Zappa is ignoring certain files during packaging. I believe this issue is caused as we are trying to exclude packages already builtin to the python lambda environment namely
```
 exclude = [
    "boto3",
    "dateutil",
    "botocore",
    "s3transfer",
    "six.py",
    "jmespath",
    "concurrent"
]
```
If any of the other python packages have any of the files or subfolders with the above names those are getting excluded as well. I noticed this as I was getting an `ImportError` while trying to do this `from django.utils import six`
 
## Expected Behavior
The above mentioned patterns should only be ignored if they are in the root level virtualenv's `site-packages` but not as sub-module of some other package.
## Actual Behavior
These patterns are being excluded at all levels. 

## Possible Fix
Copying everything and removing these files in a later step or updating the patterns.

## Steps to Reproduce
1. Install `django==1.11.6`, `zappa==0.45.1` in virtualenv
2. Create some zappa configuration without exclude argument
3. Ensure that there are no compiled files. (If there are compiled files `from django.utils import six` will not show an error as `six.pyc` would be available.)
3. Run `zappa package env`
4. Inspect the package for file `django/utils/six.py`

## Environment
* Zappa version used: 0.45.1
* Operating System and Python version: Ubuntu 16.04, Python 2.7.12

suriya:
Zappa uses Python's [`copytree`](https://docs.python.org/2/library/shutil.html#shutil.copytree) with the `ignore` parameter constructed using [`shutil.ignore_patterns`](https://docs.python.org/2/library/shutil.html#shutil.ignore_patterns). This mechanism is not very user-friendly.

Ideally, if could support `gitignore` style patters or regular expressions that would be very convenient. @Miserlou Your thoughts?

scoates:
See also #1343

scoates:
Adding this for reference: https://pypi.python.org/pypi/pathspec

scoates:
Related: #1083

suriya:
I would like to submit a PR to address this issue. My commit is here: https://github.com/suriya/Zappa/commit/61d2972dd9a303440b1f155733bb4ee8ef220a59

[...]

PR #2070, opened 2020-03-28
Allow file exclusion using pathspec

suriya:
## Description

This commit introduces `.gitignore` style patterns to customize what files can be excluded from a project's package.

## GitHub Issues
https://github.com/Miserlou/Zappa/issues/1359

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,6 +19,7 @@ idna==2.9                 # via requests
 importlib-metadata==1.5.0  # via argcomplete
 jmespath==0.9.4
 kappa==0.6.0
+pathspec>=0.7.0
 pip-tools==4.5.0
 placebo==0.9.0            # via kappa
 python-dateutil==2.6.1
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -19,6 +19,7 @@ idna==2.9                 # via requests
 importlib-metadata==1.5.0  # via argcomplete
 jmespath==0.9.4
 kappa==0.6.0
+pathspec>=0.7.0
```
TMiguelT: Looks like you need to add this dependency to `requirements.in` rather than this file, which is required for the auto generation of test_requirements.txt. That should fix the failing tests.

suriya: Could you please check the updates and help me get this pull request into master?

wrouesnel:
I'd like to bump this so we can get some attention to it: eliminating test code and the like is quite difficult with the current exclusions system (i.e. removing my "tests" folder also knocked out Jinja2 test code which is actually used).

suriya:
I can confirm that I have used this code reliably in production for a few years (at least since https://github.com/Miserlou/Zappa/pull/1859 ) If something else needs to be done to help it get merged, please guide me and I'd be happy to do it.

I would also like to point out I exclusively use docker images with Zappa and therefore do not rely on this feature anymore. I recommend others use docker images as well as the build is more reliable than the .zip package that zappa constructs.

souravjamwal77:
Hi @suriya and @wrouesnel 
Zappa has been moved to https://github.com/zappa/Zappa repository. If you want you can reopen this issue 
https://github.com/zappa/Zappa/issues/829

PR #2086, opened 2020-04-23
Increase timeout when waiting for load-balancer resources.

vshih:
## Description
Increase timeout when waiting for load-balancer resources.


lopesjeronimo:
I'm facing this same issue when using load balancer.

My workaround, instead of changing Delay, was to increase the MaxAttempts param.

PR #2093, opened 2020-04-30
Add support for REQUEST type Lambda Authorizers

### LINKED ISSUE #1159: Support for Custom Authorizer of the REQUEST type
## Context
It is now possible when creating a custom authorizer for API Gateways to set a REQUEST type event payload which will pass all headers instead of just the authorization token header.

http://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html#api-gateway-custom-authorizer-request-lambda-function-create

Is there any plan to integrate this into Zappa?

tommie-lie:
## Description
Support for REQUEST authorizers was almost there, as the `type` authorizer property was mostly passed through to the CloudFormation templates. This PR extends the handler middleware to also treat events of type REQUEST as authorization events. It also extends the template creation to treat TOKEN and REQUEST type authorizers equal (defaulting to TOKEN just as before).
The most significant change is in the logic how the authoriz*ation* type for API routes is derived from the authoriz*er* type: Instead of assuming CUSTOM if the type is not specified, the authorizer type correctly defaults to TOKEN. The authorization type is then mapped to CUSTOM for both, TOKEN and REQUEST authorizers.

Added unit tests and tested on live environments on AWS.

## GitHub Issues
Fixes #1159

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -331,6 +331,13 @@ class TestZappa(unittest.TestCase):
         self.assertEqual(z.credentials_arn, parsable_template["Resources"]["Authorizer"]["Properties"]["AuthorizerCredentials"])
         self.assertEqual("xxx", parsable_template["Resources"]["Authorizer"]["Properties"]["IdentityValidationExpression"])
 
+        # explicit TOKEN type authorizer
+        explicit_authorizer = authorizer.copy()
+        explicit_authorizer.update({"type": "TOKEN"})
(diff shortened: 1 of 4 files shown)


rodriguesgm:
Any news on this? thanks

deviousgeek:
I'd love this too, using zappa from this PR at the moment.

pcolmer:
@Miserlou Would it be possible to have this functionality added, please?

Issue #2123, opened 2020-06-18
Update fails while updating `endpoint_url` with `base_path` when `use_apigateway` is False


jwilges:
Update fails while updating `endpoint_url` with `base_path` when `use_apigateway` is `False`.

I believe this issue is similar to issue #1563 but I am trying to scope this new issue to be a small and easily-auditable patch so we can get it in master sooner than later. If this patch also happens to resolve all of the concerns in #1563, great; but if not, that ticket can remain open for further review.

## Context
Build environment:
- Python 3.8.2 virtual environment
- Zappa 0.51.0

## Expected Behavior
Update should complete without raising a `NoneType` exception and should show the endpoint's URL based on `domain_name` and `base_path` settings.

## Actual Behavior
```
(pip 19.2.3 (/opt/python-build/lib/python3.8/site-packages), Requirement.parse('pip>=20.0'), {'pip-tools'})
Calling update for stage prd..
100%|██████████| 2.97M/2.97M [00:00<00:00, 8.37MB/s]
100%|██████████| 27.4M/27.4M [00:18<00:00, 1.50MB/s]Downloading and installing dependencies..
 - psycopg2-binary==2.8.5: Downloading
Packaging project as zip.
Uploading api-prd-1592447425.zip (26.2MiB)..
Updating Lambda function code..
Updating Lambda function configuration..
Scheduling..
Unscheduled api-prd-zappa-keep-warm-handler.keep_warm_callback.
Scheduled api-prd-zappa-keep-warm-handler.keep_warm_callback with expression rate(4 minutes)!
Oh no! An error occurred! :(
==============
==============
Need help? Found a bug? Let us know! :D
File bug reports on GitHub here: https://github.com/Miserlou/Zappa
And join our Slack channel here: https://slack.zappa.io
Love!,
Traceback (most recent call last):
  File "/opt/python-build/lib/python3.8/site-packages/zappa/cli.py", line 2778, in handle
    sys.exit(cli.handle())
  File "/opt/python-build/lib/python3.8/site-packages/zappa/cli.py", line 512, in handle
    self.dispatch_command(self.command, stage)
  File "/opt/python-build/lib/python3.8/site-packages/zappa/cli.py", line 559, in dispatch_command
    self.update(self.vargs['zip'], self.vargs['no_upload'])
  File "/opt/python-build/lib/python3.8/site-packages/zappa/cli.py", line 1043, in update
    endpoint_url += '/' + self.base_path
[...]

PR #2124, opened 2020-06-19
2123 Fix `endpoint_url` when `use_apigateway` is false

jwilges:
## Description
Update no longer raises an exception while updating `endpoint_url` with `base_path` when `use_apigateway` is `False`.

Further, when `use_apigateway` is `False`, update now also:
* amends `deployed_string` with `endpoint_url` (admittedly, Zappa's formatting for this could be improved)
* attempts to touch the endpoint based on Zappa's settings

## GitHub Issues
See issue #2123 for more information and for further discussion.

### DIFF
--- a/zappa/cli.py
+++ b/zappa/cli.py
@@ -1022,11 +1022,14 @@ class ZappaCLI:
             # but we're also updating a few of the APIGW settings.
             endpoint_url = self.deploy_api_gateway(api_id)
 
-            if self.stage_config.get('domain', None):
-                endpoint_url = self.stage_config.get('domain')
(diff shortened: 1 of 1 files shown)


jwilges:
Hi @jneves, since you appear to be the sole life-blood for pull requests as of late, would you be able to take a look at this and let me know what you think?

I have deployed and tested this with our Lambdas by pinning:

`"zappa@git+https://github.com/jwilges/Zappa@0.51.1.dev2"`

I know this is probably annoying to check into, so I tried to provide a reasonable tag to pin and test against to make your life easier.

Issue #1774, opened 2019-02-06
Fuction zappa.async.run does not work outside lambda environment


erback:
## Context
When executing a function with zappa.async.run in a non lambda environment an error is raised. This seems to happen due to the fact that zappa tries to run the function in the lambda environment even if it's not in the lambda environment.  When using the @task decorator this works as expected since there is a direct invokation of the function if AWS_LAMBDA_FUNCTION_NAME env variable is not present.

## Expected Behavior
The function should be invoked directly, without errors, in a non lambda environment.

## Actual Behavior

The function fails with error
```
Invalid type for parameter FunctionName, value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```

## Possible Fix
Change this code:  https://github.com/Miserlou/Zappa/blob/4cf80b2a0212a5a8111f0f66348b46ef39507ed1/zappa/async.py#L350

to
(copied from @task decorator)
```
if (service in ASYNC_CLASSES) and (lambda_function_name):
  send_result = ASYNC_CLASSES[service](lambda_function_name=lambda_function_name,
                                                     aws_region=aws_region,
                                                     capture_response=capture_response).send(task_path, args, kwargs)
   return send_result
else: #  For local exexution
  return func(*args, **kwargs)
```

## Your Environment

* Zappa version used: 0.47.1
* Operating System and Python version: 3.6

PR #2129, opened 2020-06-29
Allows run() function to be used in local

Yaronn44:
## Description
With this PR the asynchronous `run(func, args, kwargs)` is now working as a synchronous function in local without having to mock a lambda environnement.

## GitHub Issues
Changes were made from this issue's possible fix : https://github.com/Miserlou/Zappa/issues/1774


Yaronn44:
Ok didn't expected coverage to decrease that much with such a minor change... 
I'll take a look on this

PR #2137, opened 2020-07-17
simple function Update on asynchronous.py

SteveParadox:


SteveParadox:
Some minor return functions has been placed.

PR #2148, opened 2020-08-08
Update test.sh

carlosrh18:
Added set -e flag in order to exit the bash script when encountering an error


carlosrh18:
Everything working just fine

PR #2248, opened 2026-08-14
fix: replace bare except with except Exception

TrueFurina:
### Summary

Two bare `except:` clauses catch everything, including
`KeyboardInterrupt`/`SystemExit`:

- `zappa/cli.py` — error-rate calculation fallback
- `zappa/utilities.py` — `lchmod` unavailable fallback (intended to skip)

Replace with `except Exception:` — same intended behavior, but unexpected
exceptions (e.g. `NameError`) now propagate instead of being hidden.

### Verification

- `python -m py_compile zappa/cli.py zappa/utilities.py` → OK
- `grep -rn "except:" zappa/` → 0 remaining bare handlers

### Files changed

- `zappa/cli.py` (1 line)
- `zappa/utilities.py` (1 line)
- `zappa/core.py` (3 bare excepts + 1 regex fix: `re.sub(..., re.UNICODE)` was
  passing the flag as the positional `count` argument and used a non-raw
  string — now `flags=re.UNICODE` with a raw string, eliminating the
  `SyntaxWarning: invalid escape sequence '\w'`)


TrueFurina:
Closing due to inactivity - happy to revive if there is still interest. Thanks!

